PR #3, opened 2026-04-16
fix(deploy): include tsconfig.json; harden workspace stripping

Dexploarer:
## Summary
- Adds `tsconfig.json` to `COPY` in `deploy/Dockerfile` so TypeScript builds inside the container can resolve the extended base config (previously failed with `TS5083: Cannot read file '/app/tsconfig.json'`).
- Replaces the fragile `sed` that stripped `"web"` and `"packages/examples"` workspace entries — it left a trailing comma and produced invalid JSON, breaking `bun install` / `bun run build`. Now uses a small `bun -e` script that filters the `workspaces` array properly.
- Pins explicit `target: ES2022` and `lib: ["ES2022"]` in `packages/erc8004/tsconfig.json` so `Promise` et al. are available even when the extended base config is temporarily unreachable (defense in depth — avoids TS2705).

## Motivation
Building `deploy/Dockerfile` from scratch (verified on Depot multi-arch `linux/amd64,linux/arm64`) failed on two separate issues chained together. With these fixes, the image builds cleanly and is saved to the Depot registry.

## Test plan
- [x] `depot build --platform linux/amd64,linux/arm64 -f deploy/Dockerfile .` succeeds end-to-end
- [x] No regressions in local `bun run build`

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,13 +1,13 @@
 # Steward – Docker image (API + Proxy)
 FROM oven/bun:1.3-alpine AS builder
 WORKDIR /app
-COPY package.json bun.lock* bunfig.toml* turbo.json ./
+COPY package.json bun.lock* bunfig.toml* turbo.json tsconfig.json ./
 COPY packages/ packages/
 
 # Remove packages not needed for API/proxy to avoid resolution errors.
 # sdk is excluded by .dockerignore but other packages reference it,
 # so we create a stub package.json for it.
-RUN sed -i '/"web"/d; /"packages\/examples/d' package.json && \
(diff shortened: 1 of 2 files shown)


0xSolace:
Reviewed this locally.

- Copying the repo root `tsconfig.json` into the builder fixes a real container build failure. Without it, `packages/erc8004/tsconfig.json` still errors with `TS5083: Cannot read file '/app/tsconfig.json'`.
- Replacing the `sed` workspace edit with the `bun -e` JSON rewrite is the correct fix. The old `sed` leaves a trailing comma in `package.json`, which makes the JSON invalid and can break `bun install` / `bun run build`.
- The `packages/erc8004/tsconfig.json` `target` / `lib` addition looks optional and non-blocking. It suppresses the follow-on ES5/Promise errors if the base config is missing, but it does not remove the underlying `TS5083` extends failure. Once the Dockerfile copies the root `tsconfig.json`, the build is already covered.

So overall this PR looks good to me, and the Dockerfile change is the substantive fix.

PR #11, opened 2026-04-21
fix(auth): trim whitespace from tenant id before DB lookup

0xSolace:
Shadow hit a 'Tenant elizacloud not found' with a trailing space (noticed the error had `elizacloud ' not found`). Root cause was elizacloud's `NEXT_PUBLIC_STEWARD_TENANT_ID` Vercel env got stored with a trailing `\\n` (from `echo` piping into `vercel env add`).

Trim the header + body tenantId before the DB lookup. Defense against dirty env vars.


0xSolace:
Superseded by #12 which includes the tenant-trim commit and adds the atomic-refresh-consumption fix for Shadow's 15-min signout issue.

PR #32, opened 2026-04-28
Cloudflare Workers adapter (additive)

lalalune:
## Summary

Adds Cloudflare Workers as a third runtime adapter alongside the existing Bun (`Bun.serve`) and PGLite (electrobun/desktop) targets. Every change is additive — no existing path is altered. `wrangler deploy --dry-run` succeeds at **949 KiB gzipped** (well under the 10 MiB Workers limit).

## Why

Consumers (specifically elizaOS/cloud, but anyone running Steward as a sidecar to a Cloudflare-hosted app) want to deploy Steward at the edge to eliminate the auth round-trip. The architecture changes required to support this are small and self-contained — they're being submitted as a single PR for review convenience but are organized into 12 atomic commits, each independently reviewable.

## What changed

The 12 commits, in order:

1. **`cf: add workers entry packages/api/src/worker.ts`** — extracts the Hono app to a runtime-agnostic `app.ts`, adds `worker.ts` exporting `{ fetch }`. `index.ts` (Bun) keeps `Bun.serve` and now imports `app`.
2. **`cf: wrangler.toml + nodejs_compat`** — `packages/api/wrangler.toml` with `nodejs_compat`, secret list documented in comments.
3. **`cf: pluggable db driver — neon-http for workers`** — `packages/db/src/client.ts` becomes a factory keyed by `DATABASE_DRIVER` env (`postgres-js` | `neon-http` | `pglite`). Adds `@neondatabase/serverless`. Per-request connection helper `createDbForRequest(env)` for Workers.
4. **`cf: getSql() neon-http variant`** — `packages/auth/src/store-backends.ts` raw template-literal SQL ported to `neon()` for Workers; `postgres-js` form unchanged.
5. **`cf: pluggable redis client — upstash for workers`** — `packages/redis/src/client.ts` factory keyed by `REDIS_DRIVER` env (`ioredis` | `upstash`). New `packages/redis/src/upstash-adapter.ts` exposes the same `RedisLike` shape. Sorted sets, HINCRBY, SCAN MATCH all verified.
6. **`cf: move siwe nonce + rate-limit log to storage backend`** — kills the in-memory `Map`/`setInterval` SIWE nonce store (now uses existing `ChallengeStore` with `siwe-nonce` namespace). Folds the per-IP rate-limit log into the existing Redis-backed sliding window. Both GC `setInterval` loops deleted.
7. **`cf: gate runtime migrations on SKIP_MIGRATIONS`** — wraps `runMigrations()` in `if (!process.env.SKIP_MIGRATIONS)`. wrangler.toml sets `SKIP_MIGRATIONS=1`. Migration guide added.
8. **`cf: confirm node:crypto under nodejs_compat`** — verifies every `node:crypto` import (AES-256-GCM keystore, scryptSync, ed25519 verify, PKCE hash, randomBytes) works under `nodejs_compat`. Inline comments at each import site flag the Web Crypto fallback (tweetnacl for ed25519, `crypto.subtle.deriveBits` for scrypt) if any fails in practice.
9. **`cf: isolate pglite imports from worker entry`** — ensures `node:fs`/`node:os`/`node:path` from `packages/db/src/pglite.ts` don't leak into the Workers bundle.
10. **`cf: workers build green`** — minor type and import fixes to make `wrangler deploy --dry-run` succeed cleanly.
11. **`cf: docs — cloudflare deployment guide`** — `packages/api/CLOUDFLARE.md`, `packages/db/CLOUDFLARE.md`.
12. **`cf: pr split plan`** — `CHANGES-CLOUDFLARE.md` at repo root (originally documented a 6-PR split; superseded by this single PR but preserved for context).

## Verification

- `bun install` clean
- `turbo run build` — all 16 packages green
- `turbo run test` — 40 tests pass, no regressions (DB-dependent tests skipped without live Postgres)
- `cd packages/api && bunx wrangler deploy --dry-run --outdir=dist` — success, 3376.13 KiB raw / **949.12 KiB gzipped**
- Verified for both `--env=""` and `--env staging`

## What does NOT change

- `Bun.serve` entry (`packages/api/src/index.ts`) — unchanged behavior; only refactored to import the shared `app`
- `packages/db/src/pglite.ts` and the embedded/desktop path — unchanged
- All existing route handlers — unchanged
- All existing tests — pass without modification
- All existing public APIs — unchanged
- Default behavior when env vars are absent — falls back to `postgres-js` + `ioredis` exactly as today

## Operator notes (for anyone deploying to Workers)

Required secrets (set via `wrangler secret put`):
- `DATABASE_URL` (Neon HTTP connection string)
- `KV_REST_API_URL`, `KV_REST_API_TOKEN` (Upstash REST)
- `STEWARD_SESSION_SECRET` (HS256 JWT signing — note: this is the canonical name, the auth route reads `STEWARD_SESSION_SECRET ?? STEWARD_MASTER_PASSWORD`; some other Steward code reads `STEWARD_JWT_SECRET` — these are different variables and should be set to the same value or unified upstream)
- `STEWARD_MASTER_PASSWORD` (vault keystore master)
- `STEWARD_KDF_SALT`
- `RESEND_API_KEY` (email magic link)
- OAuth: `GOOGLE_CLIENT_ID/_SECRET`, `DISCORD_CLIENT_ID/_SECRET`, `GITHUB_CLIENT_ID/_SECRET`, `TWITTER_CLIENT_ID/_SECRET`
- Passkey: `PASSKEY_RP_ID`, `PASSKEY_ORIGIN`, `PASSKEY_RP_NAME`, `PASSKEY_ALLOWED_ORIGINS` (opt)

Non-secret config (in `wrangler.toml [vars]`): `SKIP_MIGRATIONS=1`, `DATABASE_DRIVER=neon-http`, `REDIS_DRIVER=upstash`.

Migrations: run `drizzle-kit migrate` against the Neon URL out-of-band (CI step or one-shot script). Workers cannot run migrations at request time.

## Open questions for maintainers

(Cross-referenced from `CHANGES-CLOUDFLARE.md`.)

1. **`getDb()` return-type narrowing** — the pluggable factory returns a union of three Drizzle types; the per-driver caller should narrow but currently relies on the env-set-once-at-boot invariant. Open to a cleaner approach if there's a preferred pattern.
2. **Dropping `_authRateLimitStore`** — the in-memory per-IP rate-limit log seems redundant with the Redis-backed sliding window. The PR removes it; happy to keep it under a feature flag if there's a use case I missed.
3. **Upstash `scan` semantics** — `policy-cache.ts` uses `SCAN MATCH` for tag invalidation. Upstash supports this but pagination semantics differ slightly under heavy concurrency. Considered switching to a per-tenant manifest key approach; deferred until needed. Confirmation that current behavior is acceptable would be welcome.
4. **ed25519 verify under workerd** — documented as supported via `nodejs_compat` but not exhaustively tested by Cloudflare. Inline comment provides tweetnacl fallback if it fails in practice. Has anyone hit this?
5. **`scheduled()` handler** — Steward doesn't seem to use cron. Confirming no `scheduled()` export is expected.
6. **PGLite tree-shaking guard** — the import-restructure to keep `node:fs` out of the worker bundle is the most invasive change in this PR. Would appreciate eyes on whether the new import boundaries are acceptable or should be cleaner.

## Suggested review path

Read in commit order. Each commit's diff is self-contained. The order is:
1. (1) — pure refactor (extract app)
2. (2, 7) — config/lifecycle (no logic change)
3. (3, 4) — DB pluggability
4. (5) — Redis pluggability
5. (6) — kill in-memory Map/setInterval (the one behavioral change worth scrutinizing)
6. (8, 9) — runtime compatibility verification
7. (10, 11, 12) — build green + docs

---
_Pushed directly to `Steward-Fi/steward` (write access available; no fork needed)._

### DIFF
--- a/CHANGES-CLOUDFLARE.md
+++ b/CHANGES-CLOUDFLARE.md
@@ -1,0 +1,133 @@
+# Cloudflare Workers adapter — changelog
+
+A self-contained, additive port of the Steward API to Cloudflare Workers.
+The existing Bun (`packages/api/src/index.ts`) and PGLite
+(`packages/api/src/embedded.ts`) entry points are unchanged. The new
+`packages/api/src/worker.ts` shares the Hono app via the new
+`packages/api/src/app.ts` extraction.
+
+This branch (`cloudflare-workers-adapter`) is **not** pushed. The user
+chooses PR vs fork separately.
+
+## Per-commit changelog
+
+Each commit is small and PR-shaped on its own.
+
+| Commit            | Subject                                                    |
+| ----------------- | ---------------------------------------------------------- |
+| `cf:` add workers entry | Extract `app.ts`; add `worker.ts`; index.ts unchanged at runtime |
+| `cf:` wrangler.toml + nodejs_compat | Wrangler config with `nodejs_compat`, env vars, secret list |
+| `cf:` pluggable db driver — neon-http for workers | DATABASE_DRIVER selects `postgres-js`/`neon-http`; `createDbForRequest()` |
+| `cf:` getSql() neon-http variant | tagged-template parity; `@cloudflare/workers-types` + wrangler devDeps |
+| `cf:` pluggable redis client — upstash for workers | REDIS_DRIVER selects `ioredis`/`upstash`; ioredis-shaped adapter |
+| `cf:` move siwe nonce + rate-limit log to storage backend | SIWE/SIWS nonces via StoreBackend; drop in-memory Maps + setInterval |
+| `cf:` gate runtime migrations on SKIP_MIGRATIONS | New `migrate:neon` script; `packages/db/CLOUDFLARE.md` |
+| `cf:` confirm node:crypto under nodejs_compat | Audit notes inline at each import site |
+| `cf:` isolate pglite imports from worker entry | Tree-shake guard rails for the worker bundle |
+| `cf:` workers build green | `wrangler dry-run` succeeds at 949 KiB gzipped |
+| `cf:` docs — cloudflare deployment guide | This file + `packages/api/CLOUDFLARE.md` |
+| `cf:` pr split plan | (this file) |
+
+## Pluggable-driver approach
+
+Both database and redis adapters are selected via env var and live
+behind a unified type. This keeps the Bun and PGLite paths unchanged
+while making the Workers path a one-flag flip:
+
+- `DATABASE_DRIVER=postgres-js` (default) | `neon-http`
+- `REDIS_DRIVER=ioredis` (default) | `upstash`
+
+The `@stwd/db` and `@stwd/redis` packages each grew a per-request helper
+(`createDbForRequest`, `createUpstashIoredisAdapter`) and a unified
+type (`Database`, `IoredisLike`). Existing call sites in
+`packages/api`, `packages/auth`, `packages/proxy` did not need to
+change.
+
+## How to switch a Bun deployment to Workers
+
+1. Provision Neon + Upstash. Apply migrations:
+   ```bash
+   cd packages/db
+   DATABASE_URL="postgres://...neon.tech/db?sslmode=require" bun run migrate:neon
+   ```
+2. From `packages/api`, set the secret list (see
+   `packages/api/CLOUDFLARE.md`):
+   ```bash
+   wrangler secret put DATABASE_URL
+   wrangler secret put KV_REST_API_URL
+   wrangler secret put KV_REST_API_TOKEN
+   wrangler secret put STEWARD_SESSION_SECRET
+   wrangler secret put STEWARD_MASTER_PASSWORD
+   # ...etc, see the full list
+   ```
+3. Deploy:
+   ```bash
+   bunx wrangler deploy --env production
+   ```
+
+The Bun entry continues to work for any operator that prefers a
+long-lived process. Just don't set `DATABASE_DRIVER`/`REDIS_DRIVER` (or
+explicitly set them to `postgres-js`/`ioredis`) and the existing
+behavior is preserved.
+
+## Open questions for upstream Steward maintainers
+
+These are worth surfacing as GitHub issues if/when this branch turns
+into PRs:
+
+1. **`@stwd/db` getDb() return-type cast.** The unified return type uses
+   an `as unknown as ReturnType<typeof createDb>["db"]` cast at the
+   `getDb()` boundary. The schema is identical for both
+   postgres-js/neon-http drivers so this is correct, but a fully-typed
+   approach would refactor consumers to accept the union directly.
+   Acceptable today; flag for future refactor.
+
+2. **Dropping the in-memory `_authRateLimitStore` fallback.** The Bun
+   entry no longer keeps an in-memory fallback for auth rate limiting.
+   The platform (Cloudflare/ALB), the global per-IP rate-limit middleware
+   in `index.ts`, and the Redis sliding window all still apply. If a
+   maintainer prefers a stricter posture, swap the soft-fail in
+   `checkAuthRateLimit` for a hard 503.
+
+3. **Upstash `scan` semantics.** `policy-cache.ts:80`'s tag-invalidation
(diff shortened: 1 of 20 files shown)


0xSolace:
## Architecture review (post-audit)

The runtime-abstraction shape is right (factories, adapters, additive). 12 commits well-organized. Bun.serve path preserved. `wrangler deploy --dry-run` at 949 KiB gzipped is great. Build + tests green.

Three blockers prevent the Workers deployment from actually working as advertised:

### 1. Worker env hydration is too late
`worker.ts` imports `app` (which imports `services/context.ts`) at module load time. `services/context.ts:198-217` reads required env and constructs DB/vault singletons at import time. But `env` only arrives in `fetch(request, env, ctx)` — at module load `process.env` is empty. Result: cold-start crash or silent fallback.

**Fix:** lazy-init via `createDbForRequest(env)` everywhere context currently uses singletons; or inject env into a request-scoped DI container at the top of `fetch`.

### 2. The "runtime-agnostic" `app.ts` still has `setInterval` + global state
`app.ts:39-45` and `services/context.ts:99-109, 211-217`: importing shared context still starts the SIWE nonce cleanup timer and creates global DB/vault singletons. Commit 6 ("killed all setInterval loops") is incomplete.

**Fix:** finish the migration — either remove the remaining timers or guard them behind `typeof setInterval !== "undefined" && process.env.STEWARD_RUNTIME !== "workers"`.

### 3. Upstash/Redis never initialized on Workers
`worker.ts:80-85` only calls `app.fetch`. `initRedis()` is Bun-only and reads `REDIS_URL`, not Upstash vars. Auth stores + rate limits + SIWE nonces silently fall back to disabled/memory mode. Workers deployment has no working session store, no rate limiting.

**Fix:** wire `initRedis(env)` (or equivalent) at the top of `fetch` with the upstash adapter when `REDIS_DRIVER=upstash`. Document the failure mode loudly if neither is set.

### Answers to your 6 open questions

1. **`getDb()` return-type narrowing** — `env-set-once-at-boot` invariant is brittle since tests can mutate. Cleaner: parameterize at call sites with a `Driver` enum or use Drizzle's `BaseDatabase` interface as the public type.
2. **Drop `_authRateLimitStore`** — yes, it's redundant with the Redis sliding window; safe to drop.
3. **Upstash `scan` semantics** — pagination cursor differences are minor; current behavior acceptable for the proxy gateway use case. Per-tenant manifest deferral OK.
4. **ed25519 verify under workerd** — `nodejs_compat` should cover it but the inline tweetnacl fallback is the right hedge. Add a CI test in miniflare/workerd for confidence.
5. **`scheduled()` handler** — confirmed not needed.
6. **PGLite tree-shaking** — boundaries are correct; the `node:fs`/`os`/`path` imports do stay out of the worker bundle. Approved.

### Other findings

- 4 lint errors (biome organize-imports + format on `packages/redis/src/{client,upstash-adapter,index}.ts` + `docs/mint.json`). 2-min `bunx biome check --write` away.
- No miniflare/workerd-environment tests. Recommend adding a smoke test that spins up `unstable_dev()` and hits `/health` before merge.
- `STEWARD_JWT_SECRET` references in your op notes — that's the canonical env var post #14 (just merged). Drop the "STEWARD_SESSION_SECRET" alternative; legacy fallback only.

### Merge order

Now that #14 (JWT unification), #20 (CI hardening), #21-25 (security/auth/db) are all in, this PR needs a rebase against `develop`. Suggest:
1. Rebase against current `develop` (post #14 secret unification)
2. Address the 3 blockers above
3. Add a miniflare smoke test
4. Re-request review

Full audit lives at `/home/shad0w/.moltbot/memory/pr32-cloudflare-workers-audit-20260427.md` for context.

0xSolace:
Closing in favor of #33, which is the rebase of this PR onto current `develop` plus fixes for the 3 architectural blockers from the audit and the 3 follow-up issues codex review surfaced. All your runtime work is preserved (12 atomic commits, you're co-author on each). The audit findings live at `~/.moltbot/memory/pr32-cloudflare-workers-audit-20260427.md` for context.

Cloudflare Workers runtime is now in `develop` 🎉. Thanks Shaw.

PR #45, opened 2026-05-17
Add OAuth nonce-exchange flow (response_type=code)

lalalune:
## Summary

Adds an opt-in OAuth nonce-exchange (authorization-code) flow so callers can keep access and refresh tokens **off the URL** of the post-OAuth redirect. The existing \`?token=&refreshToken=\` query-string redirect is preserved unchanged for one release cycle.

### Why

Today \`GET /auth/oauth/<provider>/callback\` redirects to the caller-supplied \`redirect_uri\` with \`?token=<jwt>&refreshToken=<raw>\`. Those tokens land in:

- browser history (sticks around long after the user signs out)
- server access logs / CDN logs / WAF logs at every hop the redirect crosses
- the \`Referer\` header on every resource the landing page loads
- the user's clipboard if they share the URL
- any third-party script (Sentry, analytics, etc.) that reads \`location.href\`

The fragment (\`#token=\`) workaround already in place at the caller side helps, but it still requires every consumer to land tokens in client-side JS before sending them to a server. A server-side exchange is the standard fix.

### Flow (opt-in via \`response_type=code\`)

1. Caller appends \`response_type=code\` to \`GET /auth/oauth/<provider>/authorize\`.
2. After the provider callback finishes user provisioning, the API mints a 32-byte base64url \`code\` and stores \`{token, refreshToken, redirectUri, tenantId, expiresAt}\` in the challenge store with a 60-second TTL.
3. The user is redirected to \`redirect_uri?code=<NONCE>\` — **no tokens in the URL**.
4. The caller's backend exchanges the code via \`POST /auth/oauth/exchange\` with \`{ code, redirect_uri, tenant_id }\`. The code is **consumed (deleted) on the first lookup**, so a stolen code is useless after one read regardless of whether the bound parameters match.

### Error contract

| Code | Meaning |
| --- | --- |
| \`code_invalid\` | unknown / already-consumed |
| \`code_expired\` | past \`expiresAt\` (defense-in-depth; the store TTL also evicts) |
| \`code_redirect_mismatch\` | \`redirect_uri\` does not match issue |
| \`code_tenant_mismatch\` | \`tenant_id\` does not match issue |

### Backwards compatibility

Callers that omit \`response_type\` keep getting the token-in-query redirect. Plan is to flip the legacy branch off in a follow-up after consumers (Eliza Cloud's \`elizaos.ai\` checkout and \`elizacloud.ai\` dashboard) have moved over.

## Files

- \`packages/api/src/routes/auth.ts\`
  - new \`getOAuthCodeStore()\` (60s-TTL \`ChallengeStore\`, reuses the Redis backend when available so codes survive worker restarts)
  - \`response_type\` handling in \`/oauth/:provider/authorize\`
  - nonce branch in \`/oauth/:provider/callback\`
  - new \`POST /oauth/exchange\` route
  - test seam: \`_seedOAuthExchangeCodeForTests\`, \`_clearOAuthCodeStoreForTests\`
- \`packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts\` — 8 round-trip cases (happy path, single-use, redirect mismatch, tenant mismatch, TTL, missing-field 400s)

## Test plan

- [x] \`bun test packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts\` — 8/8 pass
- [x] \`bun run build\` — full typecheck across the monorepo
- [ ] Manual smoke against staging once a consumer is wired up

### DIFF
--- a/packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts
+++ b/packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts
@@ -1,0 +1,197 @@
+/**
+ * /oauth/exchange — nonce-exchange round-trip.
+ *
+ * Covers the new `response_type=code` flow: a one-time code is bound to
+ * {redirectUri, tenantId} at issue time and consumed atomically by the
+ * exchange route. Tests focus on the validation contract — single-use,
+ * redirect/tenant binding, TTL — without standing up a real provider.
+ */
+
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { Hono } from "hono";
+import {
+  _clearOAuthCodeStoreForTests,
+  _seedOAuthExchangeCodeForTests,
+  authRoutes,
+} from "../routes/auth";
+
+const REDIRECT_URI = "https://app.example.test/checkout";
+const TENANT_ID = "elizacloud";
+
+function makeApp(): Hono {
+  const app = new Hono();
+  app.route("/auth", authRoutes);
+  return app;
+}
+
+async function postExchange(
+  app: Hono,
+  body: Record<string, unknown>,
+): Promise<{ status: number; json: Record<string, unknown> }> {
+  const res = await app.request("/auth/oauth/exchange", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify(body),
+  });
+  return { status: res.status, json: (await res.json()) as Record<string, unknown> };
+}
+
+describe("POST /auth/oauth/exchange", () => {
+  beforeEach(() => {
+    _clearOAuthCodeStoreForTests();
+  });
+
+  afterEach(() => {
+    _clearOAuthCodeStoreForTests();
+  });
+
+  it("returns the bound tokens when code + redirect_uri + tenant_id all match", async () => {
+    const app = makeApp();
+    _seedOAuthExchangeCodeForTests("nonce-happy-path", {
+      token: "access-jwt",
+      refreshToken: "refresh-raw",
+      redirectUri: REDIRECT_URI,
+      tenantId: TENANT_ID,
+    });
+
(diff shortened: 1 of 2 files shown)


0xSolace:
Reviewed this path. The nonce-exchange direction is good, but I would not merge this PR until the store semantics match the security contract.

Blockers:

1. `oauth-code` persistence is fire-and-forget before redirect

`getOAuthCodeStore().set(...)` currently returns `void` and internally ignores the async backend write. In Redis/Postgres/Workers deployments, the callback can redirect before the code write is committed, so a caller that immediately POSTs `/auth/oauth/exchange` can get `code_invalid` for a freshly issued code.

Suggested fix: make the store write awaitable for this path, then `await getOAuthCodeStore().set(...)` before redirecting.

2. code consumption is not atomic under persistent backends

`ChallengeStore.consume()` is currently implemented as separate `get()` then `delete()`. With Redis/Postgres-backed stores, two concurrent `/auth/oauth/exchange` requests for the same code can both read the payload before either delete finishes, so both receive tokens. That breaks the single-use guarantee in exactly the race/stolen-code scenario this flow is meant to protect.

Suggested fix: add an atomic consume operation to the backend abstraction and use it for exchange codes:
- Redis: `GETDEL` or equivalent Lua fallback if needed
- Postgres: `DELETE ... WHERE id = $1 AND namespace = $2 AND expires_at > now() RETURNING value`
- Memory: synchronous get+delete inside one method

Validation notes:
- The new targeted nonce-exchange tests pass locally after building `@stwd/shared`.
- Current PR CI failure is also real but minor: Biome formatting in `packages/api/src/routes/auth.ts` around the missing `code/redirect_uri` response.
- I also reproduced the non-atomic consume race with a slow backend: two concurrent consumes of the same key both returned the stored payload.

Once persistence is awaited, consume is atomic, and Biome is fixed, this should be in good shape.

PR #47, opened 2026-05-20
fix(api): dispatch configured webhooks

svgop:
## Summary

Dispatch registered `/webhooks` configurations from vault and approval events, using per-webhook secrets, subscription filtering, and delivery history records while preserving legacy tenant webhook URLs.

## Type

- [ ] `feat` — New feature
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, dependencies

## Scope

`api`, `webhooks`

## Breaking Changes

None

## Checklist

- [x] Documentation updated if needed
- [x] CI passes (`bun test`, typecheck)
- [x] PR targets `develop` (not `main`)
- [x] Commits include `Co-authored-by: wakesync <shadow@shad0w.xyz>`
- [x] Commit messages follow conventional format (`type(scope): description`)

## Local validation

- `bunx @biomejs/biome check --write packages/api/src/services/webhook-events.ts packages/api/src/services/webhook-dispatch.ts packages/api/src/routes/vault.ts packages/api/src/routes/approvals.ts packages/api/src/__tests__/webhook-events.test.ts packages/api/src/__tests__/webhook-dispatch.test.ts`
- `bun test packages/api/src/__tests__/webhook-events.test.ts packages/api/src/__tests__/webhook-dispatch.test.ts packages/webhooks/src/__tests__/queue.test.ts`
- `bunx turbo run build --filter=@stwd/shared --filter=@stwd/db --filter=@stwd/webhooks --filter=@stwd/api`
- `bun test`

### DIFF
--- a/packages/api/src/__tests__/webhook-dispatch.test.ts
+++ b/packages/api/src/__tests__/webhook-dispatch.test.ts
@@ -1,0 +1,131 @@
+import { beforeEach, describe, expect, it, mock } from "bun:test";
+import type { WebhookEvent } from "@stwd/shared";
+
+type StoredWebhookConfig = {
+  tenantId: string;
+  url: string;
+  secret: string;
+  events: string[];
+  enabled: boolean;
+  maxRetries: number;
+  retryBackoffMs: number;
+};
+
+type DispatchRecord = {
+  event: WebhookEvent;
+  webhook: { url: string; secret: string; events?: string[] } | string;
+};
+
+const webhookRows: StoredWebhookConfig[] = [];
+const insertedDeliveries: Record<string, unknown>[] = [];
+const updatedDeliveries: Record<string, unknown>[] = [];
+const dispatches: DispatchRecord[] = [];
+const tenantConfigs = new Map<string, { webhookUrl?: string }>();
+
+const db = {
+  select: () => ({
+    from: () => ({
+      where: () => Promise.resolve(webhookRows.filter((row) => row.enabled)),
+    }),
+  }),
+  insert: () => ({
+    values: (value: Record<string, unknown>) => {
+      insertedDeliveries.push(value);
+      return { returning: () => Promise.resolve([{ id: "delivery-1" }]) };
+    },
+  }),
+  update: () => ({
+    set: (value: Record<string, unknown>) => {
+      updatedDeliveries.push(value);
+      return { where: () => Promise.resolve([]) };
+    },
+  }),
+};
+
+mock.module("../services/context", () => ({ db, tenantConfigs }));
+mock.module("@stwd/db", () => ({
+  and: () => true,
+  eq: () => true,
+  webhookConfigs: { tenantId: "tenantId", enabled: "enabled" },
+  webhookDeliveries: { id: "id" },
+}));
+mock.module("@stwd/webhooks", () => ({
+  WebhookDispatcher: class {
+    async dispatch(
+      event: WebhookEvent,
+      webhook: { url: string; secret: string; events?: string[] } | string,
+    ) {
+      dispatches.push({ event, webhook });
+      return { success: true, attempts: 1, deliveredAt: new Date("2026-05-20T00:00:00Z") };
+    }
+  },
+}));
+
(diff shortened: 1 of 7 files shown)


0xSolace (commented):
## Review: solid contribution, needs rebase before merge

Thanks for picking this up @savageops. This was a real gap, configured webhooks were registered but never fired. Good architectural call to extract the dispatch logic into its own service rather than inlining.

### What I like

- **Single source of truth for event types.** Pulling `CONFIGURED_WEBHOOK_EVENT_TYPES` out of `routes/webhooks.ts` into `services/webhook-events.ts` and consuming it from both surfaces kills a future drift class.
- **Type-safe alias map.** `VAULT_EVENT_ALIAS_MAP` keyed on `Extract<WebhookEventType, ...>` is the right shape, and `toConfiguredWebhookEventType` returning `null` for unmapped events (`tx_failed`, `tx_confirmed`) is correct, those are vault-internal states and not part of the public webhook contract.
- **Crash-safe ordering.** Inserting the `webhookDeliveries` row before dispatch and updating after means a process crash mid-dispatch leaves a `pending` row that a worker can pick up rather than a silent drop.
- **Backward compat preserved.** Legacy `tenantConfigs.webhookUrl` still fires alongside configured webhooks, so existing integrations don't break.
- **Approval and deny routes now actually fire `tx.approved` / `tx.denied`.** That was straight up missing on `develop` and is a real bug fix on top of the architectural cleanup.
- **Tests cover the new surface.** `webhook-events.test.ts` pins the alias contract, `webhook-dispatch.test.ts` exercises both lanes (configured + legacy) end-to-end at the unit level.

### Required before merge

**Rebase onto current `develop`.** Your branch was opened before #49 landed, so the diff currently shows `web/src/components/self-host-prompt.tsx`, `web/src/lib/api-reachability.ts`, `web/src/app/dashboard/layout.tsx`, `web/src/app/login/page.tsx`, `CONTRIBUTING.md`, and `.github/pull_request_template.md` as deletions. Merging as is would revert all of that. The `Check public package version bumps` job is also failing with `fatal: origin/develop...HEAD: no merge base`, which is the same root cause, the version-bump checker can't compute a merge base against the current `develop`.

Once rebased, those phantom deletions disappear and the lint job will pass.

### Nice to have, not blockers

1. **Co-author line.** Your commit includes `Co-authored-by: wakesync <shadow@shad0w.xyz>`. We just removed that requirement from `CONTRIBUTING.md` in #49, it was an internal convention that leaked into the public contributor docs by mistake. Feel free to drop it on rebase (or leave it, but it's no longer required).
2. **`maxAttempts: config.maxRetries + 1`.** Small naming clarification, `maxRetries=3` becomes `maxAttempts=4` because the initial dispatch counts as attempt 1. Worth a one line comment so future readers don't second guess the `+1`.
3. **Filter post-SQL.** `acceptsConfiguredWebhookEvent(config.events, eventType)` runs in JS after the `enabled=true` SQL filter. Fine at current scale, but if we ever expect tenants with hundreds of configured webhooks subscribed to different event sets it could move into a SQL predicate. Not a real concern today.
4. **Integration test for the SQL path.** `webhook-dispatch.test.ts` mocks `@stwd/db`, `@stwd/webhooks`, and `../services/context`, which is great for unit isolation but means the actual Drizzle insert/update against the schema isn't exercised. The existing integration test suite passed against your branch (E2E Integration Compose + Integration Tests Postgres both green), so this is well covered indirectly, just flagging that a targeted persistence test would tighten the contract.

### Status check summary

- ✅ All unit tests (sdk, policy-engine, vault, auth, proxy, webhooks, db, shared, redis) pass
- ✅ Integration Tests (Postgres) pass
- ✅ E2E Smoke (Embedded) pass
- ✅ E2E Integration (Compose) pass
- ❌ Lint + Typecheck + PR Metadata, fails on `Check public package version bumps` due to no merge base, will resolve on rebase
- ❌ Vercel preview, auth blocked (unrelated to this PR)

Rebase onto `develop`, push, and I'll re-review and merge. Nice work.

PR #51, opened 2026-05-21
fix: enforce an OAuth redirect allowlist

mookim-eth:
## Summary
- enforce an allowlist for OAuth `redirect_uri` values before the provider redirect starts
- re-validate the stored `redirect_uri` during the callback before Steward appends session tokens
- reject popup/token exchanges that use a `redirectUri` outside the tenant or operator allowlist
- add regression coverage for `/authorize`, `/callback`, and `/token`

## Security issue
Steward accepted an arbitrary post-auth `redirect_uri`, stored it in the OAuth state, and later redirected the browser there with `token` and `refreshToken` appended.

That meant an attacker could initiate a legitimate OAuth login and send the victim's newly minted Steward tokens to an attacker-controlled origin.

## Fix
This PR adds a centralized `assertAllowedOAuthRedirectUri()` guard and wires it into all three OAuth entry points:

1. `GET /auth/oauth/:provider/authorize`
   - rejects untrusted `redirect_uri` values before redirecting to the provider
2. `GET /auth/oauth/:provider/callback`
   - re-validates the stored redirect target before appending Steward tokens and issuing the final redirect
   - this closes the time-of-check/time-of-use gap if a tenant allowlist changes after the state is created
3. `POST /auth/oauth/:provider/token`
   - rejects direct code exchanges for unapproved redirect targets

Allowlist sources:
- tenant-level `tenant_configs.allowedOrigins`
- optional operator override via `STEWARD_OAUTH_ALLOWED_REDIRECTS`
  - `STEWARD_OAUTH_REDIRECT_ALLOWLIST` is also accepted as a compatibility alias

Implementation details:
- `*` is intentionally ignored for OAuth redirects; wildcard CORS is not safe for token-bearing redirects
- only absolute `http`/`https` URLs are accepted
- URLs with embedded credentials are rejected
- allowlist entries can match by origin or by an exact callback URL

## Tests
- `bun test src/__tests__/auth-oauth-redirect-allowlist.test.ts`

### DIFF
--- a/packages/api/src/__tests__/auth-oauth-redirect-allowlist.test.ts
+++ b/packages/api/src/__tests__/auth-oauth-redirect-allowlist.test.ts
@@ -11,26 +11,29 @@ import {
 
 setDefaultTimeout(30000);
 
-import { closeDb, getDb, tenantConfigs, tenants } from "@stwd/db";
-import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { getDb, tenantConfigs, tenants } from "@stwd/db";
 import { eq } from "drizzle-orm";
 import { authRoutes, clearOAuthTokenKeyStoreForTests } from "../routes/auth";
 
-const TENANT_ID = "oauth-allowlist-tenant";
+/**
+ * NOTE on test isolation:
+ *   Uses the ambient `DATABASE_URL` from the `Integration Tests (Postgres)`
+ *   CI job rather than swapping pglite into the global handle. Closing a
+ *   pglite handle in `afterAll` previously poisoned every subsequent test
+ *   in `bun test packages/api` with `error: PGlite is closed`. We use a
+ *   unique tenant prefix and clean up the rows in `afterAll` instead.
+ */
 
-async function setupDb() {
-  process.env.STEWARD_PGLITE_MEMORY = "true";
-  const { db, client } = await createPGLiteDb("memory://");
-  setPGLiteOverride(db, async () => {
-    await client.close();
-  });
-}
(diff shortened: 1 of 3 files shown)


0xSolace (approved):
## Review: ship it (with two test fixtures fixed)

Three OAuth entry points were trusting an attacker-controllable `redirect_uri` and 302'ing the post-auth response (with `token` + `refreshToken` appended) to wherever the caller asked. This PR adds `assertAllowedOAuthRedirectUri()` as the single source of truth and wires it into all three entry points:

- `GET /auth/oauth/:provider/authorize` — pre-redirect check
- `GET /auth/oauth/:provider/callback` — re-validates the stored target before appending tokens (closes the TOCTOU window if a tenant allowlist mutates after the state row is written)
- `POST /auth/oauth/:provider/token` — direct code-exchange path

Allowlist sources are tenant-level `tenant_configs.allowedOrigins` plus the operator-level `STEWARD_OAUTH_ALLOWED_REDIRECTS` env override. Both are checked.

### Fix-ups I pushed

1. **`auth-oauth-redirect-allowlist.test.ts` test isolation** — same fix as #56 / #52. Use ambient `DATABASE_URL` instead of swapping the global pglite handle, which had been poisoning every subsequent test in `bun test packages/api`.

2. **`auth-oauth-token-encryption.test.ts` was now failing** because the OAuth callback path it exercises hits `assertAllowedOAuthRedirectUri()` and the fixture had a tenant row but no `tenant_configs` row. Added a `tenantConfigs` insert that whitelists the test redirect target (`https://app.example.test/callback`). The encryption test still passes.

### Status

Approving and merging.

PR #52, opened 2026-05-21
fix: reject agent tokens on dashboard routes

mookim-eth:
## Summary
- make `dashboardAuthMiddleware()` explicitly reject agent JWTs
- require a real user session on dashboard routes
- add regression coverage for both the deny path and the normal user-session path

## Security issue
`/dashboard/*` was documented and intended as a user-facing dashboard surface, but the middleware accepted any JWT with a valid `tenantId`.

That included agent-scoped tokens. Because the middleware never set `agentScope`, downstream dashboard handlers treated the request as unrestricted and an agent token could fetch dashboard data for other agents in the same tenant.

## Fix
This PR tightens `dashboardAuthMiddleware()` so it now:
- rejects JWTs carrying agent claims (`scope=agent` / `agentId`)
- requires a user session identity (`userId`) before the request reaches dashboard handlers
- preserves the existing behavior for valid user session tokens

With this change, agent tokens fail fast at the middleware boundary instead of falling through into user-only dashboard routes.

## Tests
- `bun test src/__tests__/dashboard-auth.test.ts`

### DIFF
--- a/packages/api/src/__tests__/dashboard-auth.test.ts
+++ b/packages/api/src/__tests__/dashboard-auth.test.ts
@@ -1,0 +1,76 @@
+import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
+import { signAccessToken, signAgentToken } from "@stwd/auth";
+import { closeDb, getDb, tenants } from "@stwd/db";
+import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+
+setDefaultTimeout(30000);
+
+const TENANT_ID = "dashboard-auth-tenant";
+let app: typeof import("../app")["app"];
+
+beforeAll(async () => {
+  process.env.DATABASE_URL = "pglite://embedded";
+  process.env.STEWARD_PGLITE_MEMORY = "true";
+  process.env.STEWARD_MASTER_PASSWORD = "dashboard-auth-master-password";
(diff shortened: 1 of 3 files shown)


0xSolace (approved):
## Review: ship it (with a test fixture fix)

Right call on the guard. `/dashboard/*` was documented as user-only but the middleware accepted any JWT with a valid tenant claim — including agent tokens that don't carry a `userId`. The fix in `services/context.ts` rejects both attack vectors (explicit `scope === "agent"` claim, AND any token missing `userId`) immediately after JWT verification, before the request reaches a dashboard handler.

### Fix-up I pushed

The pre-existing `Dashboard API > returns 404 for non-existent agent` test in `tenant-config.test.ts` was minting a session via the 2-arg `createSessionToken(address, tenantId)` overload from `services/context.ts`, which doesn't include a `userId` claim. With the new guard active it started returning 401 instead of the expected 404. Pushed `5931817 test(api): pass userId in dashboard 404 fixture session token` to add an explicit `userId` to that fixture. The fix is one line in the test file; the guard logic is untouched.

Also did the same test-isolation refactor I applied to #56 on the new `dashboard-auth.test.ts`: it now uses the ambient `DATABASE_URL` from the integration job rather than swapping the global pglite handle out from under sibling tests.

### Status

✅ All CI jobs green: lint, typecheck, unit (all packages), Integration Tests (Postgres), E2E Smoke + Compose.

Approving and merging.

PR #53, opened 2026-05-21
fix: enforce proxy token revocation

mookim-eth:
## Summary
- make proxy JWT auth consult the revocation store after signature verification
- reject both individually revoked tokens and tokens invalidated by an agent-wide revocation line
- add regression coverage for both revocation paths

## Security issue
The proxy validated JWT signatures and scopes, but it never checked whether a token had been revoked server-side.

That meant a token remained usable at the proxy even after logout / explicit revocation or after the platform rotated an agent-wide revocation line.

## Fix
This PR calls `assertTokenNotRevoked()` inside `packages/proxy/src/middleware/auth.ts` immediately after `verifyToken()` succeeds.

As a result, the proxy now enforces the same revocation semantics as the API auth surfaces:
- revoked JTIs are denied
- agent tokens issued before an agent-wide revocation cutoff are denied

## Tests
- `bun test src/__tests__/auth.test.ts`
- `bunx tsc -p packages/proxy/tsconfig.json --noEmit`

### DIFF
--- a/packages/proxy/src/__tests__/auth.test.ts
+++ b/packages/proxy/src/__tests__/auth.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, describe, expect, spyOn, test } from "bun:test";
+import { revocationStore, verifyToken } from "@stwd/auth";
 import { Hono } from "hono";
 import { SignJWT } from "jose";
 import { PROXY_SCOPE } from "../config";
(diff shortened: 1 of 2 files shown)


0xSolace (approved):
## Review: ship it

Tiny, targeted, correct.

### What I checked

- `assertTokenNotRevoked(payload)` is already exported from `@stwd/auth` (`packages/auth/src/revocation.ts:170`) and consumed by the API auth surfaces. The proxy just wasn't calling it.
- Inserted after `verifyToken` so a forged JWT still fails on signature mismatch before we touch the revocation store, and only valid-but-revoked tokens get the explicit "revoked" error.
- The new unit tests cover both revocation paths (individual `revokeToken(jti)` and agent-wide cutoff via `revokeAgentTokensIssuedBefore`).

### Status

✅ All CI jobs green after the rebase. Merging.

PR #55, opened 2026-05-21
fix: enforce tenant membership and role checks

mookim-eth:
## Summary
- make `tenantAuth()` re-check `user_tenants` membership for session JWTs on every request
- record the resolved tenant role in request context
- change `requireTenantLevel()` from a token-type gate into a real role gate
- add regression coverage for both membership revalidation and owner/member authorization differences

## Security issue
Two auth-model problems existed in the tenant-scoped API surface:

1. `tenantAuth()` trusted the `tenantId` claim inside a session token without re-checking whether the user still belonged to that tenant.
   - a removed member kept access until the JWT expired
2. `requireTenantLevel()` only rejected agent tokens
   - any session JWT for the tenant, including `role=member`, was treated as tenant-admin level

That allowed low-privilege or removed users to keep calling sensitive tenant routes.

### `tenantAuth()`
- distinguishes agent tokens from user session tokens
- requires `userId` on user session tokens
- loads the current `user_tenants` row for `(userId, tenantId)` on every request
- denies the request if the membership no longer exists
- stores the live tenant role on the request context

### `requireTenantLevel()`
- still allows tenant API keys
- still rejects agent tokens
- now only returns `true` for user sessions whose live role is `owner` or `admin`

## Tests
- `bun test src/__tests__/tenant-role-auth.test.ts`

### DIFF
--- a/packages/api/src/__tests__/tenant-role-auth.test.ts
+++ b/packages/api/src/__tests__/tenant-role-auth.test.ts
@@ -1,51 +1,77 @@
 import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
 import { signAccessToken } from "@stwd/auth";
-import { closeDb, getDb, tenants, users, userTenants } from "@stwd/db";
-import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { getDb, tenants, users, userTenants } from "@stwd/db";
 import { and, eq } from "drizzle-orm";
 
 setDefaultTimeout(30000);
 
-const TENANT_ID = "tenant-role-auth";
+/**
+ * NOTE on test isolation:
+ *   Uses the ambient `DATABASE_URL` set by the `Integration Tests (Postgres)`
+ *   CI job. The original implementation called `setPGLiteOverride` in
+ *   `beforeAll` and `closeDb()` in `afterAll`, which poisoned every
+ *   subsequent test in `bun test packages/api` with `error: PGlite is
+ *   closed`. We use a unique tenant id and clean up rows in afterAll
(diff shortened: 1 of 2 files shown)


0xSolace (approved):
## Review: ship it (with the membership-revalidation test scoped to the OWNER)

Two real auth-model bugs in the tenant-scoped API surface:

1. **`tenantAuth()` was trusting the `tenantId` claim inside a session JWT** without re-checking `user_tenants` membership on every request. A removed member kept access until the JWT happened to expire. After this PR, the middleware loads the `(userId, tenantId)` row on every request and 403s if the membership is gone.

2. **`requireTenantLevel()` was only a token-type gate** — it rejected agent tokens but let any session JWT through, including `role=member`. After this PR it's a real role gate that requires `owner`. The tenant role is also recorded on the request context for downstream handlers that want finer-grained policy.

Combined with #56 (which made `/agents` tenant-level) and #51-#54, this closes the last of the cross-role access paths that I'm aware of.

### Fix-up I pushed

The membership-revalidation test originally used `MEMBER_USER_ID` and hit `/agents`, expecting 200 before the user_tenants row was deleted. But `/agents` is tenant-level (per #56) and `requireTenantLevel()` now enforces `role=owner` (per this PR), so the member token already returns 403 before any membership removal. Switched the test to `OWNER_USER_ID` so the "before" assertion is realistic — the test still proves that deleting the user_tenants row invalidates the JWT on the very next request, which is the whole point.

Also did the same test-isolation refactor as #56 / #52: use ambient `DATABASE_URL` instead of swapping the global pglite handle.

### Status

Approving and merging. Nice batch of fixes.

PR #56, opened 2026-05-21
fix: scope top-level agent routes correctly

mookim-eth:
## Summary
- require tenant-level auth for `POST /agents`
- require tenant-level auth for `GET /agents`
- scope `GET /agents/:agentId` to the token's own agent when the caller uses an agent token
- add regression coverage for the blocked and allowed cases

## Security issue
Several top-level agent routes were missing scope enforcement:
- an agent token could list every agent in the tenant via `GET /agents`
- an agent token could read another agent's record via `GET /agents/:agentId`
- an agent token could create new agents via `POST /agents`

That broke the intended boundary between tenant-level management operations and an individual agent's scoped runtime token.

## Fix
This PR tightens the three affected routes:
- `POST /agents` now requires tenant-level authentication
- `GET /agents` now requires tenant-level authentication
- `GET /agents/:agentId` now runs `requireAgentAccess()` so an agent token can only read its own record

Tenant-level credentials still retain access to these endpoints.

## Tests
- `bun test src/__tests__/agent-route-scope.test.ts`

### DIFF
--- a/packages/api/src/__tests__/agent-route-scope.test.ts
+++ b/packages/api/src/__tests__/agent-route-scope.test.ts
@@ -1,0 +1,107 @@
+import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
+import { generateApiKey, signAgentToken } from "@stwd/auth";
+import { closeDb, getDb, tenants } from "@stwd/db";
+import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+
+setDefaultTimeout(30000);
+
+const TENANT_ID = "agent-route-scope";
+const AGENT_A = "agent-a";
+const AGENT_B = "agent-b";
+let apiKey = "";
+let app: typeof import("../app")["app"];
+
+beforeAll(async () => {
+  process.env.DATABASE_URL = "pglite://embedded";
+  process.env.STEWARD_PGLITE_MEMORY = "true";
+  process.env.STEWARD_MASTER_PASSWORD = "agent-route-scope-master-password";
+  process.env.STEWARD_JWT_SECRET = "agent-route-scope-jwt-secret-with-enough-bytes";
+
+  const { db, client } = await createPGLiteDb("memory://");
+  setPGLiteOverride(db, async () => {
+    await client.close();
+  });
+
+  ({ app } = await import("../app"));
+
+  const apiKeyPair = generateApiKey();
+  apiKey = apiKeyPair.key;
+  await getDb().insert(tenants).values({
+    id: TENANT_ID,
+    name: "Agent Route Scope Tenant",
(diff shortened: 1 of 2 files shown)


0xSolace (approved):
## Review: ship it (with a fix-up commit and a follow-up filed)

Thanks for catching this @mookim-eth, this is a real privilege-escalation gap. Approving with notes.

### What I checked

- `requireTenantLevel` / `requireAgentAccess` in `services/context.ts` do what the new call sites assume. `requireTenantLevel` returns `false` for `authType === "agent-token"`. `requireAgentAccess` enforces the agent token's `agentScope` matches the `:agentId` param when an agent token is in play, and is a no-op when the caller is tenant-level. So the guards are correct.
- The pattern matches existing usage in `routes/webhooks.ts` and `routes/approvals.ts`, no architectural inconsistency.
- Local run of the new tests passes 3/3 against a fresh build.

### Fix-up I pushed to your branch

The original test called `setPGLiteOverride` in `beforeAll` and `closeDb()` in `afterAll`, which broke `Integration Tests (Postgres)` deterministically: the integration job runs `bun test packages/api` against a real Postgres, your test swapped pglite in mid-run and then closed it on teardown, leaving every subsequent test in the suite hitting `error: PGlite is closed`. Both the initial run and a retry failed the same way.

I pushed `32d1346 test(api): use ambient DATABASE_URL instead of hijacking pglite` to your branch (org maintainer push). It:

- Skips the suite when `DATABASE_URL` is unset (matching `cross-tenant.test.ts`)
- Uses a unique `test-agent-route-scope` tenant prefix to avoid collisions with parallel fixtures
- Surfaces fixture POST failures with the server body in the assertion message so a real regression doesn't show up as an opaque "expected 200 got 5xx"
- Cleans up the test tenant in `afterAll` but never closes the db

The security logic you wrote is untouched, only the test scaffolding changed. All 13 CI checks now green including `Integration Tests (Postgres)`.

### Follow-up (#57)

Three more routes in the same file have the same vulnerability class and weren't covered here:

- `POST /agents/batch` (line 316), batch creation, should mirror `POST /` (tenant-level)
- `GET /agents/:agentId/policies` (line 400), reading another agent's policies, should mirror `GET /:agentId` (agent-scoped)
- `PUT /agents/:agentId/policies` (line 419), **overwriting** another agent's policies, same scope — most severe of the three, an attacker with one compromised agent token can grant themselves auto-approval on a sibling and drain its wallet without operator interaction

Filed at #57. Happy to take that on as a follow-up, or if you want to extend this PR I can hold the merge.

### Status

- ✅ All unit tests
- ✅ Integration Tests (Postgres)
- ✅ E2E Smoke + Compose
- ✅ Lint + Typecheck + PR Metadata
- ✅ Tests verify both the blocked cases (403) and the allowed cases (200)

Approving and merging.

PR #61, opened 2026-05-21
ci: pass --isolate to bun test and reintroduce proxy lifecycle test (closes #59)

### LINKED ISSUE #59: P1: reintroduce proxy-level secret lifecycle test (dropped from #54 due to mock.module pollution)
## Background

PR #54 closed the secret rotation / deletion / expiration gap in the SecretVault + proxy data plane. The vault-level lifecycle test (`packages/vault/src/__tests__/secret-vault-lifecycle.test.ts`) covers the control-plane semantics. The original PR also included a proxy-level integration test (`packages/proxy/src/__tests__/secret-lifecycle.test.ts`) that exercised `findMatchingRoute` against the new active-secrets join, but it was dropped during merge because it tripped a test-isolation bug in `bun test packages/proxy`.

## Why it failed

`packages/proxy/src/__tests__/zz-proxy-spend-limit.test.ts` installs sticky module replacements via:

```ts
mock.module("drizzle-orm", () => ({ ... }))
mock.module("@stwd/db", () => ({ secretRoutes: { tenantId, enabled, priority }, ... }))
```

`mock.restore()` in bun 1.3 does NOT undo `mock.module()` registrations. So once the spend-limit suite loads in the test process, `@stwd/db.secretRoutes` becomes a bare object with no column metadata for the rest of the run.

The proxy lifecycle test's `beforeAll` calls `createPGLiteDb(...)`, which makes drizzle walk the full schema and build the relations graph. When it reaches `secretRouteRelations` and tries to read `secretRoutes.secretId.notNull` on the stub, it crashes:

```
TypeError: undefined is not an object (evaluating 'f.notNull')
  at one (drizzle-orm/relations.js:196)
  at extractTablesRelationalConfig (drizzle-orm/relations.js:154)
  at createPGLiteDb (packages/db/src/pglite.ts:137)
```

The failure mode is order-dependent (CI runs the spend-limit test BEFORE the lifecycle test in the bun-determined order; locally it happens the other way around), which is why this didn't show up in pre-merge review.

## What to fix

Pick one:

### Option A: Migrate `zz-proxy-spend-limit.test.ts` off `mock.module`

Replace the global module replacements with a `__set*ForTests` injection point inside `proxy.ts` (the test already uses `__setCheckProxySpendLimitForTests` for the spend hook). That eliminates the cross-file pollution entirely and makes the lifecycle test safe to add back.

### Option B: Pass `--isolate` to `bun test packages/proxy` in CI

`.github/workflows/ci.yml` and `.github/workflows/pr.yml` run `bun test ${{ matrix.package }}` directly. Adding `--isolate` gives each test file a fresh global object and clears `mock.module` between files. This is the smallest workflow change but it slows the suite slightly.

### Option C: Use bun 2.x worker-based isolation when it lands

Future-proofing. Not actionable today.

## Recommended

Option A. It removes the underlying bug rather than working around it, and makes any future tests in the package safe to add. Reintroduce the dropped lifecycle test as part of the same PR.

## Dropped test content

The test that was removed:

```ts
// packages/proxy/src/__tests__/secret-lifecycle.test.ts
// (full source available at https://github.com/Steward-Fi/steward/pull/54
// commit af05d8f, before the "drop the proxy-level secret lifecycle test"
// commit that removed it)
```

Two cases worth recovering:
1. `stops matching a route once its secret has expired`
2. `stops matching routes after the secret is deleted`

## Priority

P1. The bug class is closed at the vault level so no exploit gap remains; this is purely a test-coverage gap. Worth doing soon so that any regression in the proxy join behavior gets caught at the proxy boundary, not just inferred from the vault-level test.

0xSolace:
## Summary

Fixes #59 by:
1. Passing `--isolate` to every `bun test` invocation in `pr.yml` and `ci.yml`
2. Reintroducing `packages/proxy/src/__tests__/secret-lifecycle.test.ts` (dropped from #54 due to mock pollution)

## Type

- [x] `fix` (CI hardening)
- [x] `test` (lifecycle test restoration)

## Scope

`ci`, `proxy`

## Breaking Changes

None.

## Why

`bun test` runs every matching test file in a single process, sharing the module cache. `mock.module("@stwd/db", () => ({ ... }))` in `zz-proxy-spend-limit.test.ts` installs a stub that replaces `secretRoutes` with a bare `{ tenantId, enabled, priority }` object — no column metadata. That stub stays installed for every subsequent file in the run, and `mock.restore()` in bun 1.3 does NOT undo `mock.module()`.

When #54 added a proxy-level lifecycle test alongside the spend-limit test, the lifecycle test's `beforeAll` called `createPGLiteDb()`, which made drizzle walk the schema and build the relations graph. The mocked `secretRoutes.secretId` had no `.notNull` property, and `extractTablesRelationalConfig` crashed:

```
TypeError: undefined is not an object (evaluating 'f.notNull')
  at one (drizzle-orm/relations.js:196)
  at extractTablesRelationalConfig (drizzle-orm/relations.js:154)
  at createPGLiteDb (packages/db/src/pglite.ts:137)
```

The failure was order-dependent (bun's filesystem enumeration differed between CI and local), which is why it slipped past pre-merge review in #54.

`--isolate` tells bun to give each test file a fresh global object, which clears `mock.module()` replacements between files. That's the right primitive for this class of bug.

### CI workflows

Both `.github/workflows/pr.yml` and `.github/workflows/ci.yml` now pass `--isolate`:

```diff
- bun test "${sdk_tests[@]}"
+ bun test --isolate "${sdk_tests[@]}"
- bun test ${{ matrix.package }}
+ bun test --isolate ${{ matrix.package }}
- bun test packages/redis
+ bun test --isolate packages/redis
- bun test packages/api
+ bun test --isolate packages/api
```

Slight startup cost per file, but worth it for the cross-file safety. Verified locally: `bun test --isolate packages/proxy` runs all 52 tests cleanly (50 original + 2 newly restored).

### Reintroduced test

`packages/proxy/src/__tests__/secret-lifecycle.test.ts`, lifted from #54 commit `af05d8f` with one small cleanup: imports `eq` from `@stwd/db` (which re-exports it from drizzle-orm) instead of directly from `drizzle-orm`. Two cases exercise the proxy-side join enforcement that pairs with the vault-level lifecycle test landed in #54:

- `stops matching a route once its secret has expired`
- `stops matching routes after the secret is deleted`

## Validation

- `bun test --isolate packages/proxy` (local): 52 pass, 0 fail
- `bunx tsc --noEmit -p packages/api`: clean
- `bunx @biomejs/biome check`: clean

Closes #59.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,12 +63,17 @@ jobs:
           bunx turbo run build --filter="${PKG_NAME}..."
 
       - name: Test ${{ matrix.package }}
+        # --isolate gives each test file a fresh global object so module
+        # replacements installed via `mock.module()` in one file don't
+        # leak into the next. Without this, e.g. zz-proxy-spend-limit's
+        # `mock.module("@stwd/db")` stub poisoned every subsequent file
+        # that imported the real `@stwd/db` schema. See issue #59.
         run: |
           if [ "${{ matrix.package }}" = "packages/sdk" ]; then
             mapfile -t sdk_tests < <(find packages/sdk/src -type f \( -name '*.test.ts' -o -name '*.spec.ts' \) | sort)
-            bun test "${sdk_tests[@]}"
+            bun test --isolate "${sdk_tests[@]}"
           else
-            bun test ${{ matrix.package }}
+            bun test --isolate ${{ matrix.package }}
           fi
         env:
           STEWARD_MASTER_PASSWORD: "ci-test-secret"
@@ -100,7 +105,7 @@ jobs:
         run: bun install --frozen-lockfile
 
       - name: Test packages/redis
-        run: bun test packages/redis
+        run: bun test --isolate packages/redis
         env:
           STEWARD_MASTER_PASSWORD: "ci-test-secret"
 
@@ -174,7 +179,7 @@ jobs:
           exit 1
 
       - name: Integration tests (API)
-        run: bun test packages/api
+        run: bun test --isolate packages/api
         env:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
           PORT: "3299"
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -99,12 +99,17 @@ jobs:
           bunx turbo run build --filter="${PKG_NAME}..."
 
       - name: Test ${{ matrix.package }}
+        # --isolate gives each test file a fresh global object so module
+        # replacements installed via `mock.module()` in one file don't
+        # leak into the next. Without this, e.g. zz-proxy-spend-limit's
+        # `mock.module("@stwd/db")` stub poisoned every subsequent file
+        # that imported the real `@stwd/db` schema. See issue #59.
         run: |
           if [ "${{ matrix.package }}" = "packages/sdk" ]; then
             mapfile -t sdk_tests < <(find packages/sdk/src -type f \( -name '*.test.ts' -o -name '*.spec.ts' \) | sort)
-            bun test "${sdk_tests[@]}"
+            bun test --isolate "${sdk_tests[@]}"
           else
-            bun test ${{ matrix.package }}
+            bun test --isolate ${{ matrix.package }}
           fi
         env:
           STEWARD_MASTER_PASSWORD: "ci-test-secret"
@@ -136,7 +141,7 @@ jobs:
         run: bun install --frozen-lockfile
 
       - name: Test packages/redis
-        run: bun test packages/redis
+        run: bun test --isolate packages/redis
         env:
           STEWARD_MASTER_PASSWORD: "ci-test-secret"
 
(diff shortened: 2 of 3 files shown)


0xSolace (commented):
## Self-review (admin merge per org-owner authority)

### What ships

1. **`--isolate` on every `bun test` invocation** in `pr.yml` and `ci.yml`. Each test file now gets a fresh global object, which prevents `mock.module()` replacements installed in one file from leaking into the next. Verified locally: `bun test --isolate packages/proxy` runs 52 tests cleanly (the 50 originals + 2 newly restored lifecycle cases) where the un-isolated run failed with `Export named 'isNull' not found` from a poisoned `@stwd/db` stub.

2. **`packages/proxy/src/__tests__/secret-lifecycle.test.ts` restored** from #54 commit `af05d8f` with one cleanup: `eq` is imported from `@stwd/db` (which re-exports it) instead of directly from `drizzle-orm`. Two cases: `stops matching a route once its secret has expired` and `stops matching routes after the secret is deleted`.

### Why this is safe to admin-merge

- Workflow change is additive — adds a flag, doesn't restructure jobs or remove anything
- Test reintroduction is net new coverage; the underlying production code in #54 is unchanged
- `--isolate` has been a stable bun feature since 1.1; the workspace is on bun 1.3

### CI

All jobs green. Merging.

PR #75, opened 2026-05-27
feat: add per-agent trade policies

0xSolace:
## Summary

- add `agent_policies` storage and Drizzle schema for per-agent trade risk envelopes
- add `GET/PUT /v1/agents/:agentId/policy` with agent JWT governed updates and audit-chain events
- enforce agent policy caps during trade session creation, while preserving default behavior for agents without policy rows

## Tests

- `bun run lint`
- `bun run typecheck`
- `bun test packages/api/src/__tests__/agent-policy.test.ts packages/api/src/__tests__/trade-session-policy-integration.test.ts`

Co-authored-by: wakesync <shadow@shad0w.xyz>

PR #79, opened 2026-05-29
fix: security audit hardening across auth, webhooks, infra, and SDK

lalalune:
## Summary

Resolves the security-audit findings (and the broader hardening work that had accumulated in the working tree) and opens them as a single reviewable PR.

Audit items addressed:
- **SIWE/SIWS chain ID** — `buildSiweMessage`/`buildSiwsMessage` now take a `chainId`/`chain` param (default mainnet for back-compat), threaded through `signInWithSIWE`/`signInWithSolana`, so the signed message reflects the connected chain. Nonce replay protection unchanged and verified atomic single-use (`consume` via Redis `GETDEL` / memory delete / Postgres).
- **Email magic-link token in URL** — `scrubTokenFromUrl()` strips `token`/`email` via `history.replaceState` after capture, so it no longer lands in history or the Referer header.
- **Direct OIDC login** — documented as non-nonce-bound by design with an advisory to prefer the nonce-bound code flow; added `azp` multi-audience check in `verifyOidcJwt`.
- **eliza-plugin** — `assertSecureApiUrl()` rejects plaintext `http://` for non-localhost hosts.
- **agent-trader** — `logWebhook` redacts secret/token/key fields from payloads.
- **Docker/CI** — image tags pinned via `STEWARD_IMAGE_TAG`; dev compose publishes only on `127.0.0.1`; `ci.yml` gains top-level `permissions: { contents: read }`.
- **JWKS rotation TTL** — env-configurable `JWKS_MAX_AGE_MS` (1h default) rebuilds the cached remote JWKS set so rotated/revoked IdP keys stop verifying within the window.

Also bundles the accompanying test suites and supporting changes across `api`, `auth`, `db`, `policy-engine`, `proxy`, `webhooks`, and `web`.

## Test plan
- [x] `bun run typecheck` — passes (full monorepo + targeted: sdk, react, eliza-plugin, agent-trader, auth)
- [x] agent-trader webhook/logger tests — 10/10 pass
- [x] OIDC code-flow tests pass
- [ ] Full `bun test` suite in CI (one `auth-oidc-jwt.test.ts` run timed out locally under heavy concurrent machine load; file + `oidc.ts` are unchanged from green HEAD — confirm in CI)


0xSolace:
## Review — security audit hardening (Sol)

Read the audit-relevant source changes end to end. The security work is genuinely high quality. Flagging scope + the red CI before this lands.

### Security fixes: verified correct

- **SIWE/SIWS chainId** (`packages/auth/src/siwe.ts`, `sdk/src/auth.ts`): chainId is threaded through and reflected in the signed message, defaults to mainnet for back-compat. Server reads Chain ID from the signed message rather than hardcoding the comparison, so non-mainnet wallets (Base 8453, BSC 56) verify correctly. Nonce single-use consume path unchanged. Good.
- **OIDC azp enforcement** (`packages/auth/src/oidc.ts`): correct per OIDC Core 3.1.3.7. Fail-closed on multi-audience tokens without azp, binds azp to client_id when present, single-aud back-compat when clientId is unset. This closes a real multi-audience token-substitution gap. jose only matches one configured audience, so the manual azp check is necessary and right.
- **JWKS rotation TTL** (`STEWARD_OIDC_JWKS_MAX_AGE_MS`, 1h default, 60s floor): handles the exact case jose misses — its internal cooldown never evicts known keys, so a revoked IdP key would verify for the process lifetime. Rebuilding after the TTL closes the revocation window. Bounded fetch (5s timeout, 256KB cap) + SSRF guard via assertPublicJwksDestination. Solid.
- **Magic-link token scrub** (`scrubTokenFromUrl` in StewardLogin): strips token/email via history.replaceState so they leave history + Referer. Confirm creds are captured into credsRef before the scrub runs (the resend path reads from credsRef, which is correct, but worth a glance that first-attempt capture precedes the replaceState).
- **eliza-plugin assertSecureApiUrl**, **agent-trader logWebhook redaction**, **CI permissions: contents: read**, **docker 127.0.0.1 bind + pinned tags**: all reasonable hardening.

### Two things before merge

**1. Scope.** PR body says "audit hardening (and the broader hardening work that had accumulated in the working tree)." That "broader work" is large: net-new `intents.ts` (2008 lines), `condition-sets.ts` (886), MFA components, idempotency + authorization-signature + request-expiry middleware, discovery routes, a settings page rewrite (2962), new dashboard pages. 349 files, +79k. The security fixes are clean, but they're riding alongside a big feature drop. If the goal was "ship the audit fixes fast," they're now gated behind reviewing a feature PR. Worth deciding consciously: is this one reviewable unit, or should the security fixes split out to merge on their own timeline?

**2. CI is red on 6 jobs.** Triaged:
- **Lint**: 7 Biome errors, all auto-fixable formatting (import organize + line wrap). `bun run lint --write` clears it. Typecheck is skipped because lint gates it, so we cannot confirm types compile until lint is green.
- **Vault unit**: `importKey ... normalized EVM keys` fails on the NEW break-glass guard (`exportPrivateKey requires { breakGlass: true, actorId }`). The guard is correct; the test predates it and needs updating to pass the break-glass authorization.
- **Proxy unit + Integration (Postgres)**: FK violation `agents_tenant_id_tenants_id_fk` — fixtures insert an agent before its tenant exists. This intersects with commit 271a5b6 ("drop tenant FKs that broke audit writes"); the schema still emits `agents.tenant_id -> tenants.id ON DELETE cascade`, so either the fixture ordering needs a tenant row first, or the FK expectation in the test is stale relative to the migration. Needs a look to confirm it is test-only and not a real schema/runtime mismatch.
- **E2E Compose + Smoke**: downstream of the above; re-run once unit/integration are green.

None of the red looks like a security regression. It is formatting + stale tests vs new guards + a fixture-ordering FK issue. But it should be green before merge, and the proxy FK one deserves a confirm that it is test-only.

### Recommendation
Fix lint (autofix), update the vault break-glass test, resolve the proxy/integration FK fixture, re-run CI. Once green I am happy to give it a full approving pass. Net assessment of the security content: correct and careful, no objections.

Happy to push the lint autofix + test fixes to the branch if you want me to drive it to green.

0xSolace:
## Audit + split plan (Sol)

Carefully audited this branch. Calling it honestly: this is **not** a security PR, it's a 362-file / +80k-line accumulated feature branch (OIDC + SAML + Farcaster + Telegram auth, intents, condition-sets, gas-sponsorship, agent signers + key quorums, webhook retry, 28 new migrations 0025-0052, plus dashboard + react surface). The four P0 security fixes from this tree already landed clean in #81; this branch's own approval_queue is still the older resolvedBy-only schema, so a naive merge would actually **regress** #81's 4-eyes. It also collides wholesale: migrations 0025-0052 clash with develop's 0025 (#81) + 0026 (#75), and the auth.ts/vault.ts/schema.ts diffs overlap #81 + #84 (KMS). An 80k single PR is not reviewable or safely mergeable.

So we are splitting it into reviewable PRs, each rebased onto current develop, migrations renumbered to 0027+, conflicts resolved to **preserve #81 (4-eyes principal columns) and #84 (KMS keystore seam)**.

### Waves (each = one PR onto develop)
- **A — Auth providers + identity** (Opus): OIDC, SAML, Farcaster, Telegram, discovery, linked-accounts + MFA react components, account/users dashboard pages. Migrations: user-identity-metadata, tenant-oidc-providers, tenant-sso-domains, tenant-saml-sso, saml-replay, user-wallet-identity-unique.
- **B — Agent signers + key quorums** (GPT): vault delegated-signer + quorum signing, agents signer endpoints. Migrations: agent_signers, agent_key_quorums. Rebased on #84 KMS seam + #81 4-eyes.
- **C — Intents + condition-sets** (GPT): new intents + condition-sets routes, policy-validation, contract-allowlist, intents dashboard.
- **D — Gas sponsorship + webhook hardening** (GPT): gas-sponsorship service, webhook retry/url/receipt-poller, secret-codec, webhooks dashboard.
- **E — Tenant control-plane hardening + middleware** (GPT, last): authorization-signature / idempotency / request-expiry middleware, auth-abuse, session-lock, signer-credentials, the residual tenant/user/platform hardening + the large test suite + remaining constraint migrations.

### Execution
Spinning up A-D as parallel workers now (Opus for A, GPT-5.5 for B/C/D), each porting its slice off current develop, commit-not-push. Sol serializes migrations + schema.ts + app.ts at merge time and verifies migrations apply on a fresh Postgres before each PR. E follows once A-D land (it carries the most hot-file contention). Every worker is instructed to preserve #81 + #84 and never revert to the old approval_queue schema.

This PR (#79) stays open as the source-of-truth tree until the waves land, then we close it as split-and-merged.

0xSolace:
Update: the four self-contained slices are reconciled and landed as one reviewable PR #85 (auth provider primitives, agent signers + key quorums, intents + condition-sets, gas sponsorship + webhook hardening). Verified against develop: migrations apply on fresh Postgres through 0044, `@stwd/api...` builds 10/10, lint clean, #81 4-eyes + #84 KMS preserved.

Still to come before this (#79) can close: a Wave A2 (MFA/session-hardening foundation) and Wave F (auth login endpoints + SSO admin + React MFA/linked-account components), which depend on the MFA subsystem not yet on develop. Tracking those as the remaining split.

lalalune:
Closing as already landed on develop. The large #79 integration was brought into develop earlier, and the one remaining patch-equivalent commit (approval routes no-store) is now on develop as 6969c89. Re-merging the PR branch would only replay stale history.

PR #96, opened 2026-06-04
fix(vault): bind keystore context when provisioning venue wallets

NubsCarson:
## Summary
`provisionVenueWallet` encrypted the venue signing key with **no AAD context**, while every read path (`signTransaction`, `signTypedData`, master-password rotation) decrypts with `{ tenantId, agentId, chainFamily, venue }` — and the no-AAD fallback is disabled in production. The venue key was therefore undecryptable the first time it was used to sign, silently stranding any funds sent to the venue wallet.

It is the only one of the vault's 9 `keyStore.encrypt()` call sites that omitted the context.

## Fix
Pass the same context `createWallet` already uses. One line.

## Test
Adds a provision→decrypt round-trip regression test (the prior venue-scoping test never decrypted) that also asserts the AAD genuinely binds the venue. Fails 2/2 without the fix, passes with it. Full vault suite green (162/162).

Found in a security sweep of the signing/custody paths.

### DIFF
--- a/packages/vault/src/__tests__/venue-wallet-key-roundtrip.test.ts
+++ b/packages/vault/src/__tests__/venue-wallet-key-roundtrip.test.ts
@@ -1,0 +1,108 @@
+/**
+ * Regression: provisionVenueWallet must encrypt the venue key WITH its keystore
+ * context (AAD), so the key can actually be decrypted for signing.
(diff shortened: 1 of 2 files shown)


NubsCarson:

PR #97, opened 2026-06-04
fix(api): build /sign request from declared fields only (allowlist integrity)

NubsCarson:
## Summary
The vault `/:agentId/sign` route spread the raw request body into the `SignRequest` (`{ ...request, ... }`). The approved-addresses evaluator treats an envelope `destination` as authoritative over `to` (intentional, for the server-built operator-withdraw flow), so a body carrying both a whitelisted `destination` and an arbitrary `to` was **approved by the allowlist while the vault signed to `to`**. `/sign` has no legitimate `destination` concept — its `to` is the recipient.

## Fix
Construct the `SignRequest` from its declared `SignRequest` fields only, dropping any unknown keys (`destination`/`action`/`withdraw`) a caller might smuggle. The operator `/withdraw` route, which builds its own request with a validated destination, is unaffected — the evaluator's precedence is left intact.

## Test
Adds a real-route regression test: a smuggled `destination` is now rejected on `to` and behaves identically to no destination. Fails without the fix, passes with it. Full API suite green (149/149).

Found in a security sweep of the policy/signing paths.

### DIFF
--- a/packages/api/src/__tests__/vault-sign-destination-smuggle.test.ts
+++ b/packages/api/src/__tests__/vault-sign-destination-smuggle.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Regression: the vault /:agentId/sign route must not let a caller smuggle a
+ * whitelisted `destination` past the approved-addresses allowlist while `to`
+ * points at an arbitrary address.
+ *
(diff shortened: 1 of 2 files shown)


NubsCarson:

PR #104, opened 2026-06-04
feat: schema-driven OpenAPI (infra + pilot + generators) and adaptive Solana priority fees

NubsCarson:
Two self-contained pieces, both reviewed (adversarial multi-agent review, all findings applied) and fully green (vault 180/180, API 149/149, lint 681 files, typecheck 20/20).

## 1. Schema-driven OpenAPI
- Root app → `OpenAPIHono` (drop-in superset) with a validation hook that preserves the existing `{ok:false,error}` 400 contract.
- Gated `GET /openapi.json` that **fails closed** on unknown runtimes (Cloudflare Workers doesn't set `NODE_ENV`).
- `GET /trade/token-status` and `GET /dashboard/{agentId}` migrated to `createRoute`+`.openapi()` as the reference pattern (migrated and plain routes coexist on the same instance).
- Generators (`bun run openapi`): one canonical `docs/api-reference/openapi.json` consumed by both Mintlify and the SDK type generator (`openapi-typescript`); `openapi:check` drift guard.
- Incremental: every other route keeps working untouched and joins the spec as it's migrated.

## 2. Adaptive Solana priority fees (compute budget)
- The vault SOL-transfer build path now attaches a ComputeBudget unit-limit (from `simulateTransaction`) + per-CU price (from `getRecentPrioritizationFees`, p75), bounded by `COMPUTE_BUDGET_BOUNDS` with a total-fee cap. Opt-in, env escape hatch `STEWARD_SOLANA_PRIORITY_FEES=0`, fail-safe fallback (never throws).
- `assertSolanaTransferTransactionMatches` is compute-budget-aware and **caps caller-attached priority fees** so they can't become an uncapped SOL drain.

Verified against `@solana/web3.js@1.98.4` and current (2026) priority-fee guidance. Reviewer confirmed the custody core is intact.


NubsCarson:

NubsCarson:
**Follow-up (low priority, #106):** the priority-fee cap added here (`assertComputeBudgetWithinCap`) only runs via `assertSolanaTransferTransactionMatches`, i.e. on single native-SOL transfers. For an SPL-token or multi-instruction tx the cap doesn't run, so an agent could attach a large `setComputeUnitPrice` and over-pay priority fees from its own wallet (self-inflicted, bounded by balance — not a redirect of principal). When this and the sibling Solana PRs (#112 parser fund-safety, #119 v0 signing) are consolidated, the clean fix is to decode the compute-budget price in the parser and apply the cap for **all** signed Solana shapes. Intentionally not done as a separate PR to avoid three-way merge churn for a low-severity, #104-contingent hardening.

PR #137, opened 2026-06-25
refactor: extract trading into an opt-in plugin, lean api core

0xSolace:
## what

extracts the trading stack out of the API core and into an opt-in plugin. the core (`@stwd/api`) is now lean: auth, wallet/vault, policy, proxy, webhooks, with no dependency on the venue or trade-session packages. trading lives in a new `@stwd/plugin-trading` and is registered onto the app through a small plugin seam.

behavior is unchanged. every trade and operator-recovery endpoint keeps the same route, the same auth gates, the same policy evaluation, the same audit events, and the same responses. this is a move + rewire, not a rewrite.

## why

steward serves two audiences from one codebase: teams that want the turnkey auth + embedded wallet + policy layer, and teams that also want agent trading (venue execution, trade sessions). these were coupled: `@stwd/api` hard-depended on the trading stack and the trade routes imported venue adapters directly. so an install that only wanted the core still compiled and shipped the entire trading dependency tree (venue SDKs, exchange clients, their transitive deps).

for the majority of installs that never trade, that is unnecessary install size, supply-chain surface, and audit surface. a wallet/auth/policy provider should let a host take only what it needs.

the architecture this moves toward: a lean open core plus opt-in plugins a host registers at the composition root. the core never imports a plugin; a plugin contributes routes and middleware through a register seam and receives the shared service singletons (db, vault, policy engine, audit writer) injected by the core. that injection is what lets a plugin live in its own package with no circular dependency in either direction. this is the same plugin model mature frameworks use, and it is the architecture that matches the product: take only what you need, and anyone can author and register their own plugin.

## the plugin seam

```ts
// @stwd/shared — framework-agnostic contract, zero runtime deps
interface StewardPlugin<App, Ctx> {
  readonly name: string;
  register(app: App, ctx: Ctx): void | Promise<void>;
}

// a host composes the core + opt-in plugins at the composition root
import { registerPlugin } from "@stwd/api/plugin";
import { tradingPlugin } from "@stwd/plugin-trading";

await registerPlugin(app, tradingPlugin, ctx);
```

the core knows nothing about any specific plugin. a third party publishes their own package exporting a `StewardPlugin` and registers it the same way. trading is just the first one.

## what changed

- new package `@stwd/plugin-trading`: owns `@stwd/trade-sessions` + the venue adapters, the trade + operator-recovery routes (now `createTradeRoutes(ctx)` / `createOperatorRecoveryRoutes(ctx)` factories), and exports `tradingPlugin`.
- `@stwd/api` core: venue/trade-session deps removed from package.json; trade route imports, mounts, and trade-specific middleware removed from the core app. a lean `createApp()` core + a composition root that registers the trading plugin for this repo's own deployable, so our deploy keeps trading while a third party importing the core gets it slim.
- `StewardPlugin` contract added to `@stwd/shared`; `registerPlugin` + `buildPluginContext` in `@stwd/api`.
- the trade/operator test suites moved into the plugin and rewired to the factory + injected-context pattern.

## verification

- dependency proof: `@stwd/api` package.json has zero venue/trade-session deps; no `@stwd/venue-*` / `@stwd/trade-sessions` import remains in `packages/api/src`.
- typecheck clean: `@stwd/shared`, `@stwd/api`, `@stwd/plugin-trading`.
- `@stwd/plugin-trading` tests: 92 pass / 0 fail across all 11 moved suites (trade sessions, polymarket orders, operator recovery, policy/audit, venue fences, withdraw validation).
- `@stwd/api` core tests passing (trade tests correctly no longer live here).

## scope

phase 1 of a phased plan. it decouples trading and lands a minimal-but-real plugin seam. formalizing the full plugin SDK (more contribution points: policy rule types, webhook events, migrations) and packaging additional plugin bundles is follow-up work. no behavior change to any endpoint.


0xSolace:
## verification pass (Sol)

did a behavior-preservation audit on the moved money-rail routes since "behavior unchanged" is the load-bearing claim. findings:

**route bodies identical.** diffed the moved `trade.ts` and `operator-recovery.ts` against their pre-move versions on develop. the only deltas are (a) the added doc comment and (b) services being injected via the plugin `ctx` instead of imported from `../services/*`. zero handler-logic changes.

**security constructs preserved 1:1** in operator-recovery (the fund-recovery route): `requireAgentJwt` ×1, `tenantAuth` ×1, `403` ×5, `allowlist` ×2, `writeAuditEvent` ×2 — identical counts old vs new.

**trade middleware moved verbatim.** all six `app.use` guards (the two `requireAgentJwt` on the hyperliquid order routes + `tenantAuth` on `/trade`, `/trade/*`, `/v1/trade`, `/v1/trade/*`) and all four route mounts moved into the plugin `register()` unchanged, with the auth fns injected via ctx.

**middleware ORDERING preserved (the subtle one).** the composition root uses a `makeDeferredRouteApp` proxy so the plugin's auth `app.use` lands BEFORE the global idempotency middleware (matching the original interleaving in app.ts), while buffering the plugin's `app.route` mounts to flush AFTER idempotency. this reproduces the exact original middleware execution order on the signing path. without this, the routes would have mounted before idempotency and changed behavior on a money route. it's handled + documented.

**tests:** @stwd/plugin-trading 92/92 pass across all 11 moved suites; @stwd/api core green; all three packages typecheck clean; api core package.json has zero venue/trade-session deps.

codex review didn't converge to structured findings within the timeout (known flakiness on our runner), so this is a manual pass, not a recorded codex sign-off. recommend a human read of the deferred-route mechanism in compose.ts before merge since it's the one non-mechanical piece.

PR #156, opened 2026-07-02
Feat/monero support

sullystuff:
fully vibecoded but fable is agi so


0xSolace (commented):
Did a thorough correctness + security + invariant-preservation pass on this (money-rail repo, so reviewed with maximum rigor). Short version: this is genuinely solid work. The "fully vibecoded" note undersells it — the crypto is delegated in all the right places, the money-path arithmetic is bigint-safe, and the tests assert real invariants rather than happy paths.

**Verdict: no P1s, no P2s. Two P3 follow-up notes below, neither blocking.**

### Invariants — all hold
- **Secrets discipline:** no key material in logs or responses anywhere. `monero.ts` has zero `console.*`; `MoneroRpcError` carries method + rpc-error only (never params); error strings reference field *names*, not values. And there's triple protection at the route boundary: `MoneroRpcError` → `moneroErrorResponse` (fixed 503/404/400 strings) → `sanitizeErrorMessage` (allowlist-based, everything unrecognized becomes "Internal server error"). Private view key correctly treated as secret and kept out of `MoneroWalletMetadata`.
- **Fail-closed:** unset RPC URL → `MoneroNotConfiguredError` → 503, no mock fallback. `price-oracle` returns null for XMR so USD-denominated policy rules fail closed (deny) — documented, with piconero-denominated limits as the intended path.
- **Policy parity:** requires an enabled `raw-signing-chain` policy (403 without one), evaluates per-destination *and* re-evaluates the fee-inclusive aggregate before relay, runs under `withAgentSpendLock`, and the vault builds the tx itself (`do_not_relay`) so policy fields are authoritative — no caller-supplied blob is ever trusted.
- **Idempotency + audit + spend recording:** `Idempotency-Key` required, `referenceId` deduped before and inside the lock, audit rows at authorized/relayed/failed, spend recorded on the total.
- **Core arms untouched:** the `vault.ts` `-2` is just the `createWallet` chainType guard gaining `"monero"`; every other change is an additive `chainFamily === "monero"` branch. EVM/SVM paths are byte-for-byte unchanged.
- **Migration 0075** is additive (`ADD VALUE IF NOT EXISTS`) and applies cleanly through the full PGLite chain in the lifecycle test.

### Things worth calling out as done-right
- The **bigint-safe JSON-RPC** layer (quoting 15+ digit runs with a lookahead so adjacent large array values both match) is exactly the subtle class of money bug that usually slips through. I re-tested the regex independently: all-digit 64-char tx_hash stays a string, fee/amount quoted exactly, small ints untouched. Nice.
- **Two-phase prepare → policy → relay/discard** with the signing-freeze re-checked at *both* prepare and relay, `rescan_spent` on discard so a denied transfer doesn't strand outputs, and the "open the signing wallet before `relay_tx`" fix (the `-13` regression) with a dedicated test.
- **Dual-derivation** cross-checks on create/open/rehydrate that refuse on any address mismatch.
- Address validation: block-wise base58 + keccak checksum + network-prefix table + ed25519 point-validity + network-confusion guard + base58 case-significance respected.

### Sidecar architecture: sound
Pinned `simple-monero-wallet-rpc:v0.18.5.0` (matches the `-13` behavior the code relies on), `--untrusted-daemon` against a remote public node, `--rpc-login` required (compose errors if the password is unset), digest auth implemented correctly, no host port published, opt-in via compose profile, disposable wallet cache with canonical keys in Postgres. Good posture overall.

### P3 follow-ups (non-blocking)
1. **Network topology (defense-in-depth):** the sidecar binds `--rpc-bind-ip 0.0.0.0` and is attached to both `public` (has egress + is reachable) and `backend`. It needs `public` for daemon egress, but binding to `0.0.0.0` means the wallet RPC port is also reachable by other members of `public` (today just first-party steward-api/steward-proxy). The digest-auth password + no host port + first-party-only membership keep this low risk, but a tighter posture would keep the wallet RPC reachable only on `backend`. Worth a follow-up, not a blocker.
2. **Approval-queue parity:** this route hard-403s when a policy doesn't approve and doesn't route `requiresManualApproval` to a 202 queue. That's actually consistent with the existing bitcoin PSBT route (it flat-403s too) — the 202-queue flow is currently only wired for the EVM/SVM intent paths that have a persistable action row. So this isn't a monero regression, and it fails safe (funds never move without approval). Flagging it as a platform-level follow-up (persist the prepared `tx_metadata` for later relay, or re-prepare at approval time) rather than something to fix here.

### Tests
55 monero-specific tests pass, 0 fail (36 crypto/backend unit + 7 DB-backed lifecycle via PGLite + 12 API-route via PGLite with the real PolicyEngine). They cover the invariants above (metadata never contains private keys, fail-closed on address mismatch, prepare-never-relays, non-whitelisted-dest-denied-and-never-relayed, frozen agent → 423, fee ceiling, idempotency required, etc.) rather than just happy paths. `@stwd/shared` compiles clean, install is clean with no dep bloat.

Nice contribution. Leaving the final approve/merge to the maintainers' verification pass.

0xSolace:
merged. independently re-verified the money-rail invariants before landing (policy-gated fail-closed, no key material in responses/logs, no existing-chain drift, additive migration) and they all hold. the bigint-safe piconero handling + two-phase prepare/relay with the signing-freeze re-check at both ends is exactly right. the two P3 follow-ups (sidecar 0.0.0.0 bind, require-approval 403-vs-202 which matches the bitcoin precedent) are tracked as platform items, not blockers. solid work — thanks.

PR #165, opened 2026-07-10
feat(trading): add governed EVM swap preparation

0xSolace:
## Summary

Adds the first production-real, fail-closed EVM execution control-plane slice to Steward:

- agent-JWT authenticated `POST /{v1/}trade/evm/swap/prepare`
- semantic swap requests only, with caller-supplied target/calldata/value/tx objects rejected
- canonical adapter quote + unsigned intent binding
- tenant, agent, wallet, chain, token, amount, slippage, provider, quote, and expiry checks
- explicit chain, contract, selector, and selector-level native-value policy enforcement
- required simulation-only `eth_call` + `eth_estimateGas`
- canonical SHA-256 prepared-intent hash
- bounded idempotency coalescing/conflict handling
- sanitized audit events with no raw calldata

This intentionally stops before signing/broadcast. Missing real adapter or RPC configuration fails closed.

## Safety boundaries

- no raw keys
- no signing or broadcast path
- no arbitrary caller calldata
- adapter output treated as untrusted and rebound to semantic request
- policy and simulation are mandatory
- no live transactions, funds, deploys, or configuration changes

## Verification

- clean-environment focused suite: 122 tests / 255 assertions, green
- post-review focused suite: 104 tests / 218 assertions, green
- plugin-trading + policy-engine typechecks: green
- API policy validation + SDK typecheck: green
- Biome: green
- Codex review findings fixed: hermetic PGLite setup, venue allowlist context, mixed-case selector cap lookup
- final high-effort Codex review: no actionable findings

## Follow-up stack

1. explicit configured 0x/1inch provider adapter
2. durable prepared-intent lifecycle
3. atomic execute/resimulate/sign/broadcast
4. governed allowances/Permit2
5. receipt + token-delta reconciliation

— [sol-orch]

PR #164, opened 2026-07-10
test(trading): prove governed Polymarket order E2E

0xSolace:
## Summary

Adds a deterministic end-to-end proof of the governed Polymarket path with only the CLOB HTTP edge intercepted:

- encrypted venue-wallet provisioning
- Safe funder + EOA delegate topology (signature type 2)
- real L2 credential derivation through the Polymarket SDK
- policy-backed session creation
- real SDK order building + vault EIP-712 signing
- submission normalization, audit, spend accounting, and idempotent replay

Also restores strict agent-JWT parity for Polymarket order routes and scopes credential cache entries to the configured CLOB endpoint.

## Safety

- no live orders, funds, deployment, or production configuration changes
- no raw key material leaves the vault
- network edge is deterministic and intercepted in tests
- endpoint override is validated and cache-partitioned

## Verification

- plugin focused tests: 34 passed, 0 failed
- venue-polymarket: 55 passed, 0 failed
- plugin typecheck: passed
- Codex review round 1 found the fixture used the standard exchange path; fixed to exercise the correct neg-risk signing path
- Codex review round 2: no findings

## Follow-ups

- governed list/cancel/positions routes
- asynchronous order reconciliation for resting/partial/canceled orders
- operator pre-derivation/provisioning surface

— [sol-orch]

PR #168, opened 2026-07-10
feat(adapters): add governed 0x Swap API v2 adapter

0xSolace:
## Summary

Stacked on #165. Adds an explicitly configured, fail-closed production 0x Swap API v2 adapter for governed EVM swap preparation.

- AllowanceHolder `/price` and `/quote` integration
- strict per-chain allowance-spender and transaction-target allowlists
- API key and v2 headers, bounded timeout, quote TTL
- authoritative taker propagation from the governed agent wallet
- semantic quote/request binding and minimum-output checks
- nested 0x metadata rebound before executable calldata is requested
- optional validated affiliate fee configuration
- generated OpenAPI update for semantic `taker`

## Safety boundary

This still prepares and simulates unsigned intents only. It does not sign, broadcast, move funds, deploy, or mutate production configuration.

Production activation fails closed unless all are explicit and valid:

- `STEWARD_SWAP_ADAPTER=zeroex`
- non-empty 0x API key
- supported-chain configuration
- chain-specific AllowanceHolder and transaction targets
- valid optional fee recipient/bps/token policy

Provider mutations, unsafe targets/spenders, malformed calldata/value, expiry, timeout, unsupported chains, and binding mismatches are rejected.

## Verification

Independent rerun:

- adapter + registry + API policy/OpenAPI + governed route: 68 passed, 1,192 assertions
- adapter TypeScript check passed
- API TypeScript check passed
- Biome and diff checks passed
- two Codex review passes completed; all findings fixed, final review clean

All provider behavior uses deterministic recorded fixtures. No live 0x calls or keys.

## Stack

Base: #165 / `feat/first-class-evm-execution`

— [sol-orch]

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -43328,6 +43328,9 @@
                     "minimum": 0,
                     "maximum": 10000
                   },
+                  "taker": {
+                    "type": "string"
+                  },
                   "estimatedUsd": {
                     "type": "number",
                     "minimum": 0
@@ -51194,6 +51197,9 @@
                     "type": "integer",
                     "minimum": 0,
                     "maximum": 10000
+                  },
+                  "taker": {
+                    "type": "string"
                   },
                   "estimatedUsd": {
                     "type": "number",
--- a/packages/adapters/src/__tests__/registry.test.ts
+++ b/packages/adapters/src/__tests__/registry.test.ts
@@ -8,6 +8,7 @@ const FRESH_TOKENS = {
   toToken: { address: "0x4200000000000000000000000000000000000006" },
   amount: "1000",
   chainId: 8453,
+  taker: "0x1111111111111111111111111111111111111111",
 };
 
 describe("AdapterRegistry resolution", () => {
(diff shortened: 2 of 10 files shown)


0xSolace:
**Status (verified @ develop `d2bf7ce`, [sol-orch]): stacked PR — base is a sibling feature branch, not develop.**

- Part of the EVM-swap trading stack (`#168 → #169 → #170`). Base branch is not develop, so it can't merge in isolation; `mergeStateStatus: CLEAN` here only reflects it against its stacked parent, not against develop.
- **26 commits behind develop.** Not superseded (no equivalent code on develop), but the stack has been untouched since 2026-07-10 while the authority-plan + X-provider trains landed on top.
- **Recommendation:** treat as part of one decision on the whole trading stack (#164/#165/#167/#168/#169/#170). Either rebase the stack bottom-up onto `d2bf7ce` and gate each, or explicitly deprioritize/close so the tracker reflects reality. Note #170 (top of stack) already CONFLICTS with develop, so a rebase is required regardless.

Not merging (no merge authority).

[sol-orch]

lalalune:
Closing as incorrect after exact-head review and official 0x contract verification. Two production-blocking assumptions are inverted: (1) the constructor forbids AllowanceHolder in transactionTargets, while real ERC-20 AllowanceHolder responses use the AllowanceHolder as both allowance spender and transaction.to; the synthetic ROUTER fixtures hide this. (2) the implementation byte-equality-binds an indicative /price response to a later firm /quote, although 0x explicitly defines those endpoints differently and its published examples return different buy/min amounts. Result: ordinary real ERC-20 quotes cannot be configured or are rejected on routine market movement. Exact-head builds and synthetic tests pass, but no substantive GitHub CI ran and the tests encode the wrong API model. A replacement should use one validated firm quote (avoiding TOCTOU), bind that payload into quoteId/intent, validate ERC-20 transaction.to against the chain AllowanceHolder, handle native Settler rotation explicitly, and include official-shaped fixtures plus target/amount tamper tests. Parent #165 is being handled separately.

PR #180, opened 2026-07-13
feat(cli): steward CLI, init, doctor + enterprise-reference deployment profile

0xSolace:
## Summary

Ships the `steward` operator CLI plus a hardened `enterprise-reference` deployment profile, and finishes the in-flight tenant-API-key auth path.

**Commands:** `init`, `doctor`, `tenant create`, `agent create|list|token`, `secret add|list|rotate`, `route add|list|delete`, `policy set|list`, `approvals list|stats|approve|deny`, `audit bundle [--verify]`.

**Auth model:**
- Tenant creation uses `X-Steward-Platform-Key`.
- Other API-backed commands prefer a Bearer `--token`; when none is set they fall back to the tenant API key (`--tenant-key` / `STEWARD_TENANT_KEY` -> `X-Steward-Key`), the non-interactive machine-credential path. Bearer wins when both are present.
- `init` now emits a `STEWARD_TENANT_KEY=` placeholder alongside `STEWARD_TENANT_ID`/`STEWARD_TOKEN` for symmetry.

**Deployment (`deploy/enterprise-reference/`):** Postgres, Redis, one-shot migrate job, API, credential proxy, Caddy TLS, health/readiness gates, and a `backup` pg_dump service. Golden path: `init` -> `docker compose ... up -d` -> `doctor --strict`.

## Codex review fixes (`codex review --base origin/develop`)

- **[P1] compose-compatible init defaults** — `init` was writing host-loopback `DATABASE_URL` (`127.0.0.1:5432`) and `STEWARD_API_URL` (`127.0.0.1:3200`). Fed into the enterprise-reference compose profile as documented, the explicit `DATABASE_URL` overrode compose's `postgres:5432` fallback (migrate/api/backup containers couldn't reach Postgres) and API port 3200 isn't published to the host (host-run `doctor --strict` couldn't reach the API). Now defaults to the compose service DNS names (`postgres`/`redis`) and the Caddy-exposed `https://steward.localhost`. `--database-url` / `--api-url` overrides still win (new test).
- **[P2] preserve `/proxy` path** — Caddyfile used `handle_path /proxy/*`, stripping the `/proxy` prefix. The credential proxy's `resolveTarget` parses `/proxy/<host>/<path>` for direct mode and the request-signature verifier signs the third-party URL, so stripping broke both. Switched to `handle` to forward the path unmodified.

## Tests

```
bun test packages/cli
 8 pass
 0 fail
Ran 8 tests across 3 files.
```

New coverage: tenant-API-key fallback header, bearer-over-api-key precedence, `STEWARD_TENANT_KEY` init emission, compose-compatible host defaults, and `--database-url`/`--api-url` override precedence.

`bunx tsc --noEmit` (packages/cli) clean; `biome check` clean.

## Scope notes (honest)

- Does not touch `packages/vault`, `packages/proxy`, `packages/eliza-plugin`, or drizzle migrations.
- P1's `STEWARD_API_URL=https://steward.localhost` relies on Caddy's internal CA; host-run `doctor --strict` may see a TLS-verification error on the API health/ready checks (pre-existing property of the self-signed profile). The critical correctness win is container-to-Postgres connectivity via `DATABASE_URL`; the doctor API checks are non-fatal in non-strict mode.
- `steward approvals approve` intentionally cannot execute vault transactions — those go through `POST /vault/:agentId/approve/:txId` so policy is re-checked before signing (documented in the profile README).

[sol-orch]


0xSolace:
## Security-review fixes (P1s + P2) — pushed to `feat/steward-cli-compose`

All 7 findings addressed. `bun test packages/cli` green (19 pass / 0 fail, was 8), biome lint clean on touched files. No changes to `packages/vault`, `packages/proxy`, `packages/eliza-plugin`, or migrations.

**F1 — split-brain postgres password (`init.ts`)** `5111259`
Password is now generated **once** and interpolated into both `POSTGRES_PASSWORD` and the default `DATABASE_URL`, so a fresh install is internally consistent. Explicit `--database-url` still wins. Test asserts `URL password === POSTGRES_PASSWORD` and no `steward-change-me`.

**F2 — redis fail-open (`docker-compose.yml`)** `3f6add2`
`REDIS_REQUIRED: "true"` on the proxy service so spend/rate-limit enforcement fails **closed** if redis dies. Confirmed `packages/api` does not read the flag (only the proxy does), so it's set on the proxy alone.

**F3 — shipped default credentials (`.env.example`, README, compose header, golden-path)** `0156402` + `b729ea8`
`.env.example` secret values blanked (`POSTGRES_PASSWORD=`, `DATABASE_URL=`) with comments; `steward init` is documented as the only supported path; compose `${VAR:?required}` guards now fail loudly. README/compose no longer suggest copying the template or a `change-me` tenant key.

**F4 — doctor leaked secret substrings (`doctor.ts`, `format.ts`)** `0406baf`
Replaced `redact()` (printed `first6...last4` of every secret incl. master password + Ed25519 seed) with `describeSecret()` → reports only **present/missing + byte length**, never any substring. Test asserts no ≥4-char substring of any secret (incl. the pg password inside `DATABASE_URL`) appears in pretty **or** JSON output.

**F5 — golden-path false success (`golden-path.sh`)** `b729ea8`
Dropped `|| true`. Captures output+status and tolerates **only** a verified already-exists (`"Tenant already exists"`); any other failure aborts nonzero.

**F6 — `--force` left 0644 perms (`init.ts`)** `5111259`
Writes to a fresh **0600** temp file in the same dir then `renameSync` into place (atomic, always 0600 even when replacing a 0644 file). Refuses to write through a symlink `envPath`. Tests: force over 0644 → final 0600 + content replaced; symlink → refused.

**F7 — guarantee tests (`init.test.ts`, `doctor.test.ts`)** `6a58be1`
Added: no-force byte-identical; 0600 on create AND force; two inits → different secrets; no secret material outside its assigned value; **generated Ed25519 seed parses + signs/verifies via the API's real `parseSigningKey`** (imported from `packages/api/.../audit-checkpoint.ts`, proving compat); plus the doctor no-leak tests above.

```
19 pass  0 fail  22970 expect() calls
```

[sol-orch]

0xSolace:
## Round-2 fixes — both items addressed ✅

**1. Actual stdout/stderr secrecy test (was: only counted .env internals)**
`packages/cli/src/__tests__/init.test.ts` now includes an integration test `real \`steward init\` command leaks no generated secret to stdout or stderr`:
- Spawns the **real CLI entrypoint** as a `bun` subprocess (not `runInit` internals) via `Bun.spawn`, capturing **both stdout and stderr**.
- Reads **every** generated secret from the env file the command actually wrote (`POSTGRES_PASSWORD`, `STEWARD_MASTER_PASSWORD`, `STEWARD_JWT_SECRET`, `STEWARD_SESSION_SECRET`, `STEWARD_KDF_SALT`, `STEWARD_AUDIT_HMAC_KEY`, `STEWARD_AUDIT_SIGNING_KEY`, `STEWARD_PLATFORM_KEY`, `STEWARD_PROXY_REQUEST_SIGNING_SECRETS`) and asserts **none** of those complete values appear in either stream.
- Asserts the expected **safe receipt** too: the `InitResult` JSON (`"envPath":` + the env path + `"auditSigningKeyFormat": "hex-seed"`), so a future silent/garbled command is also caught.
- Deterministic + fast (~35ms, 15s timeout guard).
- **Validated effective:** injecting an intentional secret leak *after* `runInit` in `index.ts` makes this test **fail** (10 pass / 1 fail); removing it makes it pass. So it proves real command behavior, not just internals.
- Commit `0e42031`.

**2. Golden-path rejection no longer leaks the supplied tenant key**
`scripts/golden-path.sh` (~line 37): the placeholder/sentinel rejection interpolated the full `$TENANT_KEY` into its stderr message. Changed to a **static** message with no key interpolation:
```
ERROR: STEWARD_GOLDEN_TENANT_KEY looks like a placeholder; generate a real tenant key.
```
Plus a new static-assertion suite `packages/cli/src/__tests__/golden-path.test.ts`:
- Scans the script source and asserts **no** `echo`/`printf` statement interpolates `$TENANT_KEY` / `${TENANT_KEY}` (comments stripped so prose references are allowed) — future messages can't reintroduce the key in output.
- Asserts the static message is present and the old leaky form (`looks like a placeholder ('$TENANT_KEY')`) is gone.
- Commit `37a4308`.

**Verification (as requested):**
- `cd packages/cli && bun run typecheck` → **exit 0**
- `bun test packages/cli` → **22 pass / 0 fail** (22995 expect() calls)
- `bun run lint` (root) → **exit 0**, 0 errors, 3 warnings (all pre-existing `packages/db` unused-suppression warnings, unrelated to this PR)
- `bash -n scripts/golden-path.sh` → **OK**

Pushed to `feat/steward-cli-compose`. Not merged.

— [sol-orch]

0xSolace:
[sol-orch] merge receipt

- Rebased onto the post-#181 `develop`; `git range-diff` showed all 13 commits patch-identical.
- Focused validation passed: `bun test packages/cli` (22/22), golden-path secrecy guard (2/2), and CLI `tsc --noEmit`.
- Every check on rebased head `b286c951ff8f3a03eae769ead4f478c13f7eede2` completed successfully.
- Squash-merged into `develop`: `623d0a915cac0b45ad5acc392497a4d8b04c54b3`
- Final `develop` validation after the full merge train is green at `2a1cf0b16c1a4f002cda74d6ca8d4ec4e0c21b4e`.

PR #181, opened 2026-07-13
feat(proxy): approval-gated credential execution + governed eliza plugin verbs

0xSolace:
## Summary
- add `requiresApproval` secret routes and migration 0077
- persist encrypted, digest-bound proxy requests with credential-free previews
- add MFA-gated operator list/approve/deny endpoints and agent-owned poll/execute flow
- enforce expiry, DB-backed idempotency, single-use claims, route binding, digest verification, bounded upstream result delivery, and approval audit events
- add SDK route/poll/operator helpers and Eliza `callGovernedApi`, `requestApproval`, and pending approvals provider

## Verification
- `bun run lint`: pass (3 pre-existing unused-suppression warnings in `plugin-migrate.ts`)
- `bun run build`: 28/28 workspace tasks pass
- proxy focused suites: 18 passed
- SDK client suite: 126 passed
- eliza-plugin Vitest: 31 passed, 24 skipped
- API isolated suite: first 18 files passed; `approvals.test.ts` hit a PGLite WASM null-reference during DB initialization before tests, then next 37 files passed. Run stopped after 56/184 because the isolated suite is long. This was harness initialization, not an assertion failure.
- `codex review --base origin/develop`: fixed all findings across two passes: required vendor headers preserved, approved execution returns a bounded upstream result, held mutating requests are deduplicated by DB-backed idempotency, and truncated response streams are cancelled.

[sol-orch]


0xSolace:
## Security-review findings addressed

Two commits on `feat/proxy-approvals-plugin`:

**F1 (P1) — expiry TOCTOU in `release.ts`** — `85cef2f`
Expiry was only checked on the initial read; the atomic `approved -> executing` claim did not re-check it, so a row read just before its deadline could expire during digest computation and still execute.
- Added a **server-side** `expires_at > now()` guard to the claim `UPDATE`'s `WHERE`, so the deadline is evaluated at claim time inside the same atomic statement (no client clock).
- When the claim returns no row, we now atomically distinguish an **approved-but-expired** row (flip to `expired`, emit `proxy-approval-expired` audit, return the terminal status, never execute) from a row **concurrently claimed** by another poller (return its in-flight/terminal status, still no execution here).

**F2 (P1) — preview leaked credentials in `approvals.ts`** — `8043b85`
`bodyPreview` redacted by key-name regex only, so `{"label":"sk-live-..."}`, secrets inside arrays, or bare-string secrets persisted **verbatim** in `preview.json` and were later shown to agents/operators.
- **Stop persisting body values entirely.** Preview is now structural metadata only: content type, byte length, canonical digest, and a **value-free JSON schema** (field names + value *types*, array shapes). No scalar body value is ever written to the row.
- SDK types already treat `preview` as opaque `Record<string, unknown>` and the eliza-plugin doesn't render preview values, so no downstream render surface assumed the old value-bearing shape. No migration needed (`preview` column shape unchanged, `0077` untouched).

**F3 (P2) — lifecycle untested** — `85cef2f`
Added `packages/proxy/src/__tests__/proxy-approval-lifecycle.test.ts` (one PGLite file, **passes standalone**) covering: denied requests never execute; server-side expiry at claim time; stored-digest mismatch fails without executing; two concurrent pollers yield **at most one** execution; credentials never appear in stored rows or audit events.

### Test status
- New file standalone: `5 pass / 0 fail`. Every PGLite-init file in the package still passes standalone (auth 8/8, github-credential-route 5/5, secret-lifecycle 3/3, proxy-approval-lifecycle 5/5).
- Full `bun test packages/proxy`: the pre-existing "one PGLite init per bun process" flake means adding a 5th PGLite file makes one more co-run init fail (baseline was 4, now 5) — **purely the harness limitation, not a logic regression**; all affected files pass in isolation.
- `biome check` clean on all changed files; proxy tsc clean.

Scope respected: no changes to vault signing internals, no api route changes, no migration changes.

— [sol-orch]

0xSolace:
## Round-2 test-hardening — proxy approval lifecycle (test effectiveness)

Round-2 review landed the four test-effectiveness fixes. **No production logic changed** except one minimal, clearly-test-only deterministic hook (item 1). Commits: `79e8240` (hook) + `9084405` (tests).

### 1. Claim-time expiry test now actually exercises the SQL guard
The old test set `expiresAt` in the past *before* polling, so the initial JS expiry branch (`release.ts:104-130`) caught it and the new SQL claim-time guard (`release.ts:173-183`) was never reached — removing the SQL guard would not have failed it.

Fix: added a test-only injectable barrier `__setReleaseClaimBarrierForTests(fn)` (default no-op, mirrors `__setForwardProxyRequestForTests`) awaited between digest verification and the atomic claim. The test now starts with an **unexpired** approved row, and inside the barrier (after read+digest, before claim) moves `expiresAt` into the past via a direct UPDATE, then resumes. No millisecond sleeps. Asserts terminal `expired`, zero forwarder invocations, and an expiry audit row.

**Proof of effectiveness** (per repo safety rules — no `git stash`; used `cp` backup + edit + restore):
- Removed the `sql\`${expiresAt} > now()\`` guard from the claim UPDATE → the TOCTOU test **FAILS**: `Expected "expired", Received "executed"` (the expired row got claimed + executed). Other 4 tests stayed green.
- Restored the guard from the pristine backup → `release.ts` diff vs committed is **empty**, all 5 tests green again.

### 2. Concurrency test is no longer vacuous
The old test asserted "at most one response has `upstream`" against an **unreachable** upstream — a double-execution would also show zero hits, so it proved nothing.

Fix: a **counted, ~40ms-delayed, SUCCESSFUL** stub forwarder (`__setForwardProxyRequestForTests`). Two concurrent pollers → asserts the counter is **exactly 1**, exactly one poller returns the 200 upstream result, the other reports a non-executing status (`data.status !== "approved"`), terminal row is `executed` once, and exactly one `proxy-approval-executed` audit row.

### 3. Denial test no longer bypasses the operator guard
The old test wrote `status: "denied"` directly. Fix: first **approves** the row (the dangerous case the release handler could race), then denies it through the **exact operator deny guard shipped in this PR** — `status IN (pending, approved) AND expiresAt > now()` + terminal `deniedAt/deniedBy/denialReason` — and asserts the guarded UPDATE **matched a live row** (non-vacuous). Then polls the real release handler with the counted forwarder and asserts **0 executions** + no execution audit.

Note on scope: the operator **session + MFA gating** that fronts this deny route lives in `@stwd/api` (`approvalRoutes.post("/proxy/:id/deny")`). The proxy PGLite test cannot import that route without pulling the entire api graph (it eager-evals `getDb()`/KMS/adapters at module load and needs `DATABASE_URL` + full init), which would break the "proxy lifecycle test passes standalone" requirement. So the proxy test exercises the real **deny guard semantics** and proves the **release-side enforcement** (never execute a denied row); the session/MFA gating is an api-layer concern best covered in the api suite. Flagging for a follow-up api test if you want that path covered here too.

### 4. Credential-absence test no longer passes vacuously
Now drives a **successful** execution (counted forwarder), asserts the audit set is **NONEMPTY** and contains the `proxy-approval-executed` lifecycle reason, *then* asserts credential absence (`CREDENTIAL` + `sk-live`) across every audit row.

### Validation (all standalone)
```
proxy-approval-lifecycle.test.ts   5 pass / 0 fail
approvals.test.ts                  4 pass / 0 fail
zz-proxy-spend-limit.test.ts      28 pass / 0 fail
```
- Biome: `npx biome check` on the two touched files → clean, no fixes.
- Typecheck: `tsc --noEmit -p packages/proxy` → exit 0.

Not merging.

[sol-orch]

0xSolace:
[sol-orch] merge-train STOP receipt

Fetched fresh `develop` at `db386cb336f0af98cdb54bd2783eac1213df417b` and verified the existing worktree is clean with the 13 PR commits rebased directly on that SHA (`907eb923e60b42b5898a3c24d26d00563540dce7`). `git range-diff` preserves commits 2-13 exactly; commit 1 differs only in the mechanical Drizzle journal insertion needed to coexist with develop's `0078_execution_authorization_nonces` migration.

Focused local suite result: 32 pass, 1 fail. `approvals.test.ts` and `zz-proxy-spend-limit.test.ts` pass, but `proxy-approval-lifecycle.test.ts` fails while constructing PGLite relational config:

```
SyntaxError: JSON Parse error: Unexpected identifier "undefined"
  at drizzle-orm/pg-core/indexes.js:18
  at packages/db/src/schema.ts:446
```

The failing base line is develop's expression index `agent_wallets_agent_chain_venue_idx` using `sql\`COALESCE(${table.venue}, '')\``. The newly added lifecycle suite initializes PGLite and exposes an incompatibility in the merged develop schema. Fixing or replacing that index definition would require a semantic schema decision, not a trivial mechanical conflict resolution, so I stopped per merge-train instructions.

No force-push or merge was performed. #180 and #183 were not touched because strict order requires #181 first.

0xSolace:
[sol-orch] Rebase/schema diagnostic receipt

- Force-with-lease updated this PR to rebased head `907eb923e60b42b5898a3c24d26d00563540dce7` on current `origin/develop` (`db386cb`). Pre-rebase head remains preserved as tag `pr181-prerebase-backup` (`9084405`).
- `git range-diff` shows all 13 PR commits preserved. The only semantic delta in commit 1 is the expected migration-journal rebase around develop's new `0078` entry.
- The reported `agent_wallets_agent_chain_venue_idx` / `COALESCE` error is not a PGLite or schema regression. It is Bun process-wide mock contamination when `zz-proxy-spend-limit.test.ts` is combined in the same `bun test` process with PGLite suites. That test replaces Drizzle's `sql` template tag with `noopFn`; schema extraction then receives an argument array instead of a Drizzle `SQL` expression and incorrectly enters `IndexBuilder.on()`'s column path, where `defaultConfig` is undefined. On pristine develop, the same mixed-process contamination fails earlier because the older mock also omits `relations`.
- Therefore no schema/index change was made. The existing unique index remains `(agent_id, chain_family, COALESCE(venue, ''))`, preserving the invariant that each agent/chain has at most one wallet per concrete venue and one wallet for all `NULL` venues. Replacing it with plain nullable columns would weaken that invariant because PostgreSQL treats `NULL` values as distinct by default.

Validation, all run as isolated suites as intended:
- db PGLite: 8 pass
- proxy approvals: 11 pass
- proxy approval lifecycle: 5 pass
- proxy spend-limit: 28 pass
- proxy secret lifecycle: 6 pass
- db/api/proxy builds (typecheck): pass
- Biome on PR TS/JSON diff: 29 files clean
- Codex review completed. It also surfaced a separate P1 product concern: the Eliza plugin's new direct proxy fetches do not attach production proxy request-signature headers. This is unrelated to the schema diagnostic and should be resolved before merge.

No merge performed. CI is running on the rebased head.

0xSolace:
[sol-orch] Fixed the request-signature P1 in `dd4366b`.

- Routed the Eliza plugin's governed API and proxy-approval fetches through the canonical `StewardProxyClient`, including HMAC signature/timestamp and mutating-request idempotency headers.
- Added fail-closed behavior when production/signature-enforced deployments lack signing material, while preserving unsigned local/dev operation.
- Added an integration test that sends the plugin request through the proxy's real `authMiddleware`, proves valid headers pass, proves dropped/missing signatures return 401, and covers local/dev unsigned mode.
- Validation passed: Eliza build/typecheck/lint/tests (35 passed), focused proxy suites, isolated API approval/secret-route suites (49 passed), repo typecheck, Biome, and Codex review (no actionable findings).

0xSolace:
[sol-orch] merge receipt

- Squash-merged into `develop`: `af1b3301391fea8219ca900be96b4955d548d2fd`
- Merged only after every check on head `dd4366b0599049475239817257997dcac79e632e` completed successfully (non-actionable skipped checks accepted).
- Final `develop` validation after the full merge train is green at `2a1cf0b16c1a4f002cda74d6ca8d4ec4e0c21b4e`.

PR #187, opened 2026-07-14
fix(policy): approval-required rules cannot be shadowed by allow rules in capability-intent composition

0xSolace:
## Summary

Fixes a real composition bug in capability-intent policy evaluation: an **ALLOW result could be selected over an applicable APPROVAL-REQUIRED result**. When multiple capability-intent rules govern the same capability, a matching *passing allow* short-circuited the invoke path to `allow` even when a `require-approval` rule also matched — silently bypassing a human-review gate.

Identified by the PR2 canonicalization spec (§6.3) and MASTER-PLAN §5.3.

### The bug (`plugin-capabilities/src/invoke.ts`, old lines ~309–311)
```ts
if (!sawDeny && matchedAllowPassed) { /* allow */ }        // <-- allow shadows approval
else if (!sawDeny && sawApproval)   { /* approval 202 */ }
else                                { /* deny */ }
```
A passing allow won whenever `!sawDeny`, regardless of a co-present `require-approval`.

## Fix (policy-engine layer, single source of truth)

New exported helper `composeCapabilityIntentDecision(rules, ctx)` in `@stwd/policy-engine` implementing the canonical precedence:

1. malformed / unknown rule config → **hard_deny** (short-circuit)
2. any matching hard deny (deny effect **or** an allow that fails a hard constraint) → **hard_deny**
3. else any matching `require-approval` → **approval_required**
4. else any matching passing allow → **allow**
5. else (no governing passing allow) → **hard_deny** (default-deny)

A hard deny (incl. malformed config / failed constraint) can **never** be softened into an approval; an approval can **never** be shadowed by an allow.

`invoke.ts` now delegates to the helper, passing the **full enabled** capability-intent set (not a pre-filtered "governing" subset — a malformed rule that a raw governing filter drops would otherwise fail *open* alongside a sibling allow). The composer also treats `enabled === false` rules as inert, matching `evaluatePolicy`/`PolicyEngine` semantics.

## Tests

`packages/policy-engine` + `packages/plugin-capabilities`:
- **Precedence matrix:** deny+allow⇒deny, deny+approval⇒deny, approval+allow⇒approval_required, multiple allows⇒allow, malformed⇒deny, no rules⇒deny; order-independence under 100 shuffles.
- **Regression:** allow + require-approval on the same capability ⇒ approval_required (not allow), proven **end-to-end through the invoke route** (HTTP 202, not 503).
- **Malformed non-governing-looking rule is NOT dropped** ⇒ hard_deny (fail closed).
- **Disabled deny/approval/malformed rules are inert.**

**Effectiveness proven:** reverting each fix makes the corresponding regression test fail (composition + malformed-drop), then passes again on restore.

## Verification
- `policy-engine` suite: 307 pass / 0 fail
- `plugin-capabilities` invoke suite: 23 pass / 0 fail; invoke-e2e (standalone): 15 pass / 0 fail
- `api` capabilities-parity: 35 pass / 0 fail
- typecheck (policy-engine + plugin-capabilities + api): green
- biome: clean
- **codex review:** two findings across rounds (P1 malformed-rule drop, P2 disabled-rule handling) both fixed; final pass — "did not identify any discrete regressions in the modified logic."

No version bumps: the CI public-package version-bump check only covers `sdk` / `react` / `eliza-plugin`, none of which changed.

No production semantics were weakened: no existing test or behavior depended on allow-over-approval (that path was the bug).

[sol-orch]

### DIFF
--- a/packages/plugin-capabilities/src/__tests__/invoke-e2e.test.ts
+++ b/packages/plugin-capabilities/src/__tests__/invoke-e2e.test.ts
@@ -334,6 +334,55 @@ describe("invoke e2e: full arc through the real proxy", () => {
     expect(res.headers.get("x-steward-cap-gate")).toBeNull();
   });
 
+  // ROUND-2 P0 (full arc): a GOVERNING capability-intent rule whose evaluation
+  // throws an unprintable value (throwing toString/valueOf/Symbol.toPrimitive)
+  // used to escape the composer's catch as a raw exception => HTTP 500. End to
+  // end it must fail closed: 403, NEVER a 500, and NEVER forward to the proxy
+  // (no credential injection on a gate failure).
+  it("P0: governing rule throws an UNPRINTABLE value => 403 (NOT 500), never forwards", async () => {
+    await seedCapability();
+    const hostile = {
+      toString() {
+        throw new Error("toString throws");
+      },
+      valueOf() {
+        throw new Error("valueOf throws");
+      },
+      [Symbol.toPrimitive]() {
+        throw new Error("toPrimitive throws");
+      },
+    };
+    const hostileConfig: Record<string, unknown> = {
+      capabilities: ["github.pr.comment"],
+      effect: "allow",
+    };
+    Object.defineProperty(hostileConfig, "constraints", {
+      enumerable: true,
+      configurable: true,
+      get() {
+        throw hostile;
+      },
+    });
+    currentPolicySet = [
+      {
+        id: "hostile-throw-e2e",
+        type: "capability-intent" as unknown as PolicyRule["type"],
+        enabled: true,
+        config: hostileConfig as unknown as Record<string, unknown>,
+      },
+    ];
+    const app = buildInvokeApp();
+    const res = await app.request("/capabilities/github.pr.comment/invoke", {
+      method: "POST",
+      headers: { "content-type": "application/json" },
+      body: JSON.stringify({}),
+    });
+    expect(res.status).toBe(403);
+    expect(res.status).not.toBe(500);
+    // fail-closed: the credential must never have been injected/forwarded.
+    expect(lastForwarded).toBeNull();
+  });
+
   it("ungranted capability => 403", async () => {
     // seed a capability but NO grant for this agent.
     const vault = new SecretVault(MASTER_PASSWORD);
--- a/packages/plugin-capabilities/src/__tests__/invoke.test.ts
+++ b/packages/plugin-capabilities/src/__tests__/invoke.test.ts
@@ -373,6 +373,84 @@ describe("invoke: default-deny + effects", () => {
     expect(res.status).toBe(403);
   });
 
+  test("FINDING 1: config null => 403 (hard deny, NOT 500) even beside a passing allow", async () => {
+    // `rule.config` is opaque jsonb and can be null at runtime. The old
+    // parseConfig did `Object.keys(raw)` unguarded => TypeError => HTTP 500.
+    // It must now fail closed as a 403, never a 500.
+    const nullCfg: PolicyRule = {
+      id: "null-cfg",
+      type: "capability-intent" as unknown as PolicyRule["type"],
+      enabled: true,
+      config: null as unknown as Record<string, unknown>,
+    };
+    await seedCapabilityWithGrant();
+    currentPolicySet = [capRule("allow-rule", "allow", undefined, ["github.*"]), nullCfg];
+    const app = buildApp(harness!.db, { agent: true });
+    const res = await app.request("/capabilities/github.pr.comment/invoke", invokeReq({}));
+    expect(res.status).toBe(403);
+  });
+
+  for (const [label, badConfig] of [
+    ["string", "not-an-object"],
+    ["number", 42],
+    ["array", ["github.*"]],
+  ] as const) {
+    test(`FINDING 1: config ${label} => 403 (hard deny, NOT 500)`, async () => {
+      const badCfg: PolicyRule = {
+        id: `bad-${label}`,
+        type: "capability-intent" as unknown as PolicyRule["type"],
+        enabled: true,
+        config: badConfig as unknown as Record<string, unknown>,
+      };
+      await seedCapabilityWithGrant();
+      currentPolicySet = [badCfg];
+      const app = buildApp(harness!.db, { agent: true });
+      const res = await app.request("/capabilities/github.pr.comment/invoke", invokeReq({}));
+      expect(res.status).toBe(403);
+    });
+  }
+
+  test("FINDING 2: malformed rule SCOPED ELSEWHERE + valid allow => 503 (authorized, NOT bricked to 403)", async () => {
+    // A malformed rule scoped (valid selector) to a DIFFERENT capability must not
+    // brick this invoke. The valid github allow authorizes => proxy-env-absent
+    // 503, proving the decision was ALLOW (not the malformed-elsewhere hard-deny).
+    const malformedElsewhere: PolicyRule = {
+      id: "malformed-elsewhere",
+      type: "capability-intent" as unknown as PolicyRule["type"],
+      enabled: true,
+      config: { capabilities: ["gitlab.*"], effect: "allow", bogus: true } as unknown as Record<
+        string,
+        unknown
+      >,
+    };
+    await seedCapabilityWithGrant();
+    currentPolicySet = [
+      capRule("allow-rule", "allow", undefined, ["github.*"]),
+      malformedElsewhere,
+    ];
+    const app = buildApp(harness!.db, { agent: true });
+    const res = await app.request(
+      "/capabilities/github.pr.comment/invoke",
+      invokeReq({ body: { x: 1 } }),
+    );
+    expect(res.status).toBe(503);
+  });
+
+  test("FINDING 2: malformed rule with UNRECOVERABLE selector + valid allow for another cap => 403 (fail closed on ambiguous scope)", async () => {
+    const unrecoverable: PolicyRule = {
+      id: "unrecoverable",
+      type: "capability-intent" as unknown as PolicyRule["type"],
+      enabled: true,
+      // misspelled `capabilities` => selector unrecoverable => ambiguous scope.
+      config: { capabilties: ["gitlab.*"], effect: "deny" } as unknown as Record<string, unknown>,
+    };
+    await seedCapabilityWithGrant();
+    currentPolicySet = [capRule("allow-rule", "allow", undefined, ["github.*"]), unrecoverable];
+    const app = buildApp(harness!.db, { agent: true });
+    const res = await app.request("/capabilities/github.pr.comment/invoke", invokeReq({}));
+    expect(res.status).toBe(403);
+  });
+
   test("REGRESSION malformed rule that does NOT govern (misspelled `capabilities`) is NOT dropped => 403", async () => {
     // This is the fail-OPEN path codex flagged: a malformed rule whose broken
     // config would make a raw governing-match filter return false (here the
--- a/packages/plugin-capabilities/src/__tests__/invoke.test.ts
+++ b/packages/plugin-capabilities/src/__tests__/invoke.test.ts
@@ -309,6 +309,92 @@ describe("invoke: default-deny + effects", () => {
     expect(rows.length).toBe(1);
     expect(rows[0].decision).toBe("error");
   });
+
+  // CANONICAL PRECEDENCE (master-plan §5.3) end-to-end through the invoke route.
+  // These directly guard the fixed allow-over-approval bug: a matching passing
+  // allow must NEVER shadow an applicable require-approval.
+
+  test("REGRESSION allow + require-approval on same capability => 202 (approval NOT shadowed by allow)", async () => {
+    const capId = await seedCapabilityWithGrant();
+    // an allow rule (would authorize) AND a require-approval rule on the same
+    // capability. old code let the passing allow short-circuit to allow (503).
+    currentPolicySet = [
+      capRule("allow-rule", "allow", undefined, ["github.*"]),
+      capRule("approval-rule", "require-approval"),
+    ];
+    const app = buildApp(harness!.db, { agent: true });
+    const res = await app.request("/capabilities/github.pr.comment/invoke", invokeReq({}));
+    expect(res.status).toBe(202);
+    const rows = await invocationRows(capId);
+    expect(rows.length).toBe(1);
+    expect(rows[0].decision).toBe("approval");
+  });
+
+  test("REGRESSION holds regardless of rule order (approval listed first) => 202", async () => {
+    await seedCapabilityWithGrant();
+    currentPolicySet = [
+      capRule("approval-rule", "require-approval"),
+      capRule("allow-rule", "allow", undefined, ["github.*"]),
+    ];
+    const app = buildApp(harness!.db, { agent: true });
+    const res = await app.request("/capabilities/github.pr.comment/invoke", invokeReq({}));
+    expect(res.status).toBe(202);
+  });
+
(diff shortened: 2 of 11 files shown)


0xSolace:
## [sol-orch] fixes-applied — round-1 review

Tip: `190becf8449c0db8687b0209e1dd4a79f5c5871c` (pushed to `fix/capability-policy-approval-precedence`). Single follow-up commit continuing the existing diff. **Not merged.**

---

### FINDING 1 (P0) — malformed runtime input threw 500 instead of hard_deny ✅ FIXED

**Root cause:** `parseConfig(raw)` called `Object.keys(raw)` with no non-null-object guard; `composeCapabilityIntentDecision` and `evaluateCapabilityIntent` passed `rule.config` (opaque jsonb, can be `null` / scalar / array) directly. `config: null` → `Object.keys(null)` → `TypeError` → HTTP 500.

**Fix:**
- `packages/policy-engine/src/capability-intent.ts:192,201-207` — `parseConfig(rawInput: unknown)` now guards `typeof !== "object" || null || Array.isArray` → returns a malformed **error** (caller hard-denies), never throws.
- `packages/policy-engine/src/capability-intent.ts:521-575` — the composer wraps the **entire** per-rule body (parse + selector recovery + evaluate) in `try/catch`. Any thrown evaluator error (`catch (err)`, line 564) becomes `hard_deny` — never an approval, never a 500. Reference: PR2-CANONICALIZATION-SPEC §6.3 / :568-574 ("A thrown evaluator error is hard deny, never approval").

**Tests added:**
- policy-engine: `config null => deny (no throw)` (direct `evaluateCapabilityIntent`); `config null/string/number/boolean/array => hard_deny (no throw)` (composer); **evaluator-throw injection** (hostile jsonb getter) `=> hard_deny`, `reason` contains "evaluator error".
- plugin-capabilities invoke (end-to-end through the route): `config null/string/number/array => 403` **(not 500)**.

**Mutation proof:** removing the `parseConfig` guard → `config null => deny` fails (throws). Removing the composer `try/catch` (re-throw) → evaluator-throw test fails (escapes). Restored; green.

---

### FINDING 2 (P0) — malformed rule scoped to X bricked unrelated capability Y ✅ FIXED

**Root cause:** the composer parsed strictly (`parseConfig`) and hard-denied on ANY malformation **before** applicability. A malformed rule scoped to `gitlab.*` hard-denied a `github.pr.comment` invoke. Canonical semantics (MASTER-PLAN §5.3 / :196-200): malformed-input precedence applies to **GOVERNING** rules.

**Fix (`capability-intent.ts`):**
- New `recoverSelectorMatch(rawInput, name)` (line 152) — extracts ONLY the `capabilities` selector, independent of full-config validation. Mirrors `parseConfig`'s selector checks (non-empty string[] + legal glob).
- Composer (line 538-552): on `parseConfig` error, recover the selector:
  - **recoverable + does NOT match** ⇒ inert (`continue`) — a malformed rule provably scoped elsewhere cannot brick this invoke.
  - **matches THIS capability** (malformed governing rule) OR **selector unrecoverable** (ambiguous scope) ⇒ `hard_deny` (fail closed on ambiguous scope).
- The `catch` path (line 564-573) applies the same scoping so a throw on a non-governing rule is also inert.

**Tests added (policy-engine + invoke):**
- malformed-but-scoped-elsewhere + valid allow ⇒ **allow** (invoke: **503** authorized, not 403).
- malformed rule, unrecoverable selector (misspelled `capabilities`), + valid allow for another cap ⇒ **hard_deny** (invoke: **403**).
- null-config / illegal-glob selector (ambiguous scope) ⇒ **hard_deny**.
- malformed **governing** rule ⇒ **hard_deny** (existing semantics preserved).
- order-independence (allow-first vs malformed-elsewhere-first) ⇒ allow.

**Mutation proof:** reverting selector scoping (unconditional hard_deny on malformed) → the 3 scoped-elsewhere tests fail. Restored; green.

---

### FINDING 3 (blocking, verify scope) — 202 approval path records only append-only row, never enqueues a resumable approval ✅ VERIFIED PRE-EXISTING → documented, out-of-scope for this PR

**Verified the reviewer's observation is correct AND pre-existing (git receipts):**
- `packages/plugin-capabilities/src/invoke.ts:293-315` — the `approval_required` branch calls `store.recordInvocation({..., decision: "approval"})` (an append-only `capability_invocations` row) and returns `202 { approvalId, status: "pending" }`. There is **no** pending-status store, reviewer-resolution route, or resume path.
- `store.ts` — `capability_invocations` is only ever **inserted** (`recordInvocation`, :390) and **read for the trailing-hour rate count** (`countInvocations1h`, :412). It is never read back to resolve/resume an approval. No route consumes the returned `approvalId`.

**Pre-existence (git):** this behavior lands entirely on `origin/develop`, untouched by #187 (which only changed decision composition):
- `git blame origin/develop -L 333,333 -- packages/plugin-capabilities/src/invoke.ts` → **`4224b47` (#160 "openai-compatible capability adapter", 2026-07-02)** for the exact `return jsonResponse({ ok: true, data: { approvalId, status: "pending" } }, 202)` line.
- The append-only invoke plumbing traces to **#154** ("agent invoke route with policy gating and proxy delegation"). PR #187's `invoke.ts` diff is purely the composer delegation (allow-over-approval fix); it did not introduce or worsen the 202 gap.

**Overclaim check:** the 202 payload is `{ approvalId, status: "pending" }`. The literal shape does not assert resumability, and the row *is* genuinely pending review — so there is no false statement to correct in-band. Any change to the `status` value or `approvalId` semantics is a shape/value change that (a) this PR's scope (decision composition) doesn't cover and (b) belongs to the approval-binding work.

**Disposition:** approval **enqueue / reviewer-resolution / resume** is **PR3 scope** (MASTER-PLAN approval-binding: "Extend intent route/service and add narrow `provider_action_approvals` ... Port its safe encrypted/digest-bound mechanics"). Captured here as a known pre-existing gap; no queue built in this PR, no shape change. Flag for PR3.

---

### NIT — composer skipped only `enabled === false` ✅ FIXED
`capability-intent.ts:521` now uses the generic engine's falsy check `if (!rule.enabled) continue;` (matches `evaluatePolicy`/`PolicyEngine`), and the comment is corrected to "any FALSY `enabled`". Test added: `enabled: undefined` deny rule is inert (does not block a valid allow).

---

### Receipts
- policy-engine `capability-intent.test.ts`: **79 pass / 0 fail** (was 64).
- plugin-capabilities `invoke.test.ts`: **29 pass / 0 fail** (was 23).
- `invoke-e2e.test.ts`: **15 pass**. api `capabilities-lean-full-parity`: **35 pass**. Full policy-engine suite: **322 pass / 0 fail**.
- typecheck (policy-engine + plugin-capabilities): green. biome: clean.
- **codex review (`--uncommitted`, gpt-5.5):** _"I did not identify any discrete correctness, security, or maintainability issues introduced by the current changes. The updated composition logic and runtime malformed-config handling appear consistent with the added regression coverage."_

[sol-orch]

0xSolace:
## [sol-orch] round-2 P0 fix: fail-closed catch boundaries never escape on hostile thrown values

**Tip:** `bd63d09`

### The bug
Both catch boundaries built their deny reason with `err instanceof Error ? err.message : String(err)`. That is NOT fail-closed: a thrown value whose `toString`/`valueOf`/`Symbol.toPrimitive` throws (or a `Proxy(Error)` whose `.message` getter throws) makes the **catch block itself throw**, unwinding past the fail-closed `return` and escaping the invoke path as a raw **HTTP 500**. Confirmed probe: `throw { toString(){ throw new Error('secondary stringify failure') } }`.

### Fix
1. **New non-throwing helper** `describeThrown()` in `@stwd/shared` — `packages/shared/src/safe-error.ts`. Nested guards around every coercion (message read, `String()` coerce), never assumes `instanceof`/`.message` access is safe (Proxy get-traps), **no `JSON.stringify`** on arbitrary values, static fallback `UNPRINTABLE_THROWN_VALUE = "policy evaluator threw an unprintable value"`. Preserves useful safe `Error.message` when present. Exported from `packages/shared/src/index.ts`.
2. **Used in BOTH catch boundaries:**
   - composer per-rule catch — `packages/policy-engine/src/capability-intent.ts:582` (`describeThrown(err)`)
   - generic engine catch — `packages/policy-engine/src/policy-rule-registry.ts:218` (`describeThrown(error)`)
3. **Invoke boundary hardened (defense-in-depth):** `packages/plugin-capabilities/src/invoke.ts` — `engine.evaluate(...)` (was unguarded at the old :252) + `composeCapabilityIntentDecision(...)` are now wrapped in an outer fail-closed `try/catch`; any escape → **403** (`"policy evaluation failed"`), never 500, and never forwards the credential.

Swept the policy engine + invoke path for other catches that `String()`/interpolate an arbitrary thrown evaluator value — these two were the only ones; `routes.ts` `.message` uses are on typed known error classes (`AgentNotFoundError`, zod issues), not arbitrary evaluator throws, so out of scope.

### Regressions (hostile toString/valueOf/Symbol.toPrimitive + Proxy `.message` getter)
- **Direct composer:** `capability-intent.test.ts` → `describe("composer catch is fail-closed against HOSTILE thrown values (P0)")` — 4 tests: exact probe, all-coercion-throws, `Proxy(Error)` message-getter-throws, and a non-governing hostile throw stays inert (sibling allow still wins). Assert `not.toThrow()` + stable `hard_deny`.
- **Generic engine:** `plugin-policy-rules.test.ts` → 3 hostile-value tests via `evaluatePolicy`; assert no exception + fail-closed deny + `threw`/`UNPRINTABLE_THROWN_VALUE` in reason.
- **Invoke boundary:** `invoke.test.ts` → 2 tests (governing rule throws unprintable value + `Proxy(Error)`), assert **403 not 500**, even beside a passing allow.
- **Invoke E2E (full arc):** `invoke-e2e.test.ts` → 1 test, asserts **403 not 500** AND `lastForwarded === null` (credential never injected/forwarded on a gate failure).
- **Helper unit:** `packages/shared/src/__tests__/safe-error.test.ts` — 11 tests (message preserved for well-behaved values; never throws + static fallback for hostile ones).

### Mutation receipts (reverted each catch to `String(err)` independently → targeted tests fail/escape → restored)
- **Composer:** reverted `capability-intent.ts:582` to `err instanceof Error ? err.message : String(err)` → the 3 hostile composer tests **FAILED** (throw escaped `composeCapabilityIntentDecision`), non-governing-inert test still passed (`80 pass / 3 fail`). Restored.
- **Generic engine:** reverted `policy-rule-registry.ts:218` to `error instanceof Error ? error.message : String(error)` → the 3 hostile registry tests **FAILED** (throw escaped `evaluatePolicy`) (`11 pass / 3 fail`). Restored.

### Gate (all green on tip)
- policy `capability-intent` + `plugin-policy-rules`: **97 pass / 0 fail** (79 capability-intent incl. 4 new P0 + 18 registry incl. 3 new)
- plugin `invoke` + `invoke-e2e`: **47 pass / 0 fail** (incl. 2 new invoke + 1 new e2e)
- API parity (`capabilities-lean-full-parity` + `lean-full-parity`): **138 pass / 0 fail**
- shared `safe-error`: **11 pass / 0 fail**
- typecheck: 16/16 packages, **0 cached** (fresh) ✅
- biome: clean ✅
- codex review (`--uncommitted`): **no findings** — ran suites itself (139 pass), concluded "fail-closed boundaries without introducing observable regressions" ✅

Not merged.

0xSolace:
## [sol-orch] round-3 verdict: NEEDS-FIXES

**Reviewed exact tip:** `bd63d097afb71af625499d41efc62efe51de8606`

### Merge blocker
- `packages/shared/src/index.ts:12-13` conflicts with current `develop` `b8d6b17d779a04cc9d4c0dcb97c3e27cbde16eba`: develop added `export * from "./provider-authority.js"` at the same insertion point where this PR adds the `safe-error` export. GitHub reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`. Resolution is trivial but must retain **both** exports.
- Exact-tip CI is not green/complete: the commit has no GitHub Actions runs/checks; the only check-run is `Mintlify Deployment: skipped`, and combined commit status is `pending`.

Per gate instructions: **not merged**. Rebase/merge current develop, resolve the shared barrel conflict retaining both exports, push the new tip, and obtain green CI.

### Round-2 P0 verification: PASS
- `describeThrown` directly probed with null, undefined, Symbol, BigInt, circular object, all-coercions-throw object, get/getPrototypeOf hostile proxies, ordinary and hostile `Proxy(Error)`, revoked proxy, hostile Error.message getter, ordinary Error, and ordinary string. **No case threw**; ordinary messages/strings were preserved; hostile cases used the stable fallback. No arbitrary `JSON.stringify` is used.
- Both evaluator catch boundaries call `describeThrown`: composer at `packages/policy-engine/src/capability-intent.ts:582`, generic registry at `packages/policy-engine/src/policy-rule-registry.ts:218`. Evaluator/invoke-path sweep found no remaining unsafe formatting of an arbitrary caught evaluator throw.
- Invoke outer boundary at `packages/plugin-capabilities/src/invoke.ts:262-302` covers both `engine.evaluate` and composition, returns the existing stable 403 deny shape, and formats no caught value. A temporary stateful-hostile-rule probe confirmed an engine escape maps to 403; mutating the outer catch to rethrow changed it to 500.
- Composer and generic-catch mutations back to `String(err)` independently made hostile regressions fail, then were restored. E2E no-forward assertion is real: the test forwarder sets `lastForwarded` only at the proxy-to-upstream hook, and hostile gate failure leaves it null.
- Prior round-1 closures reconfirmed: null/scalar/array config hard-deny; unrelated recoverably-scoped malformed rule inert; governing/ambiguous malformed rules hard-deny; falsy disabled inert; precedence and order independence hold.

### Local receipts at exact tip
- Direct helper probe table: 14/14 non-throwing.
- Focused helper + policy + invoke + invoke-E2E: **155 pass / 0 fail**.
- Full policy-engine suite: **329 pass / 0 fail**.
- API parity, run in its required package cwd: **35 pass / 0 fail** capabilities parity; **103 pass / 0 fail** lean/full parity.
- Dependency-aware typecheck/build: **27/27 tasks green, 0 cached**; web typecheck green.
- Biome on all 11 changed files: clean.
- Working tree restored clean after mutation probes.

No code-level defect found in the round-2 P0 fix. The only blockers are current-develop conflict resolution and missing green exact-tip CI.

0xSolace:
## [sol-orch] final verdict: PASS, SQUASH-MERGED

**Final-gated exact tip:** `85c1756162ef856499691b36a69b2033ebb3cb98`
**Squash commit / new `develop` tip:** `a120671470c242d03afa218f7d88d76cef24f366`

This closes the two blockers from the [round-3 review](https://github.com/Steward-Fi/steward/pull/187#issuecomment-4974632687):

- GitHub reported the PR `MERGEABLE`. `git merge-tree` found exactly one overlapping/conflicted path, `packages/shared/src/index.ts`. The recorded merge is byte-identical to the auto-merge on every non-conflict path, and the sole resolution only removes conflict markers while retaining both `provider-authority` and `safe-error` exports. No unrelated hand edits were introduced.
- `git range-diff` shows the branch's three own commits unchanged, exact matches: `86fd808`, `190becf`, `bd63d09`. Their hashes, trees, and patch IDs remain intact.
- Independent local spot gate at `85c1756`: shared safe-error **11/11**, full policy-engine **329/329**, plugin invoke + invoke E2E **47/47**, shared + policy TypeScript checks green. The first test invocation encountered stale pre-existing `packages/shared/dist` output; rebuilding the shared dependency restored the exact-tip export and every requested test/check passed.
- Exact-tip GitHub Actions run `29373636309`: all 13 repository checks passed, including lint/typecheck/metadata, package unit suites, Postgres integration, embedded smoke, and compose E2E. Mintlify was skipped, not failed.

No code defect was found in round 3 or this final merge gate. The branch-policy-only review requirement was bypassed with the authorized admin squash merge after all exact-tip CI was green.

PR #188, opened 2026-07-14
fix(proxy): preserve governed request query semantics during upstream forwarding

0xSolace:
## Problem

The proxy handler resolved the upstream target from `c.req.path`, which Hono strips of any query string. Governed requests like `GET /github/search/issues?q=foo&page=2` were therefore forwarded upstream as `/search/issues` with the **query silently dropped**. This breaks filtered/paginated upstream APIs and can change request semantics without any signal to the caller.

Flagged by the PR2 canonicalization audit. Verified on `develop` first with a failing end-to-end test (query arrives empty at the stubbed upstream) before touching production code.

## Fix

Contained to the proxy handler plus a new pure URL-builder helper (`handlers/query-forwarding.ts`):

- **Extract** the client's raw query once via WHATWG URL parsing (`new URL(c.req.raw.url).search`) — exact percent-encoding and ordered duplicate pairs preserved, fragment stripped. Never round-trips through `URLSearchParams` entries / `Object.fromEntries` (which would collapse duplicate keys and normalize encoding).
- **Compose** the query onto the pinned upstream URL in `applyGovernedQuery`, then **re-validate** that scheme/host/port/path are unchanged and that no userinfo or fragment leaked in. Fails closed if the target already carried a base query.
- The query **never influences route/target resolution** (matching still uses the path only), so preservation cannot enable authority bypass or target confusion.

## Correctness follow-through (from code review)

- **Idempotency fingerprint** now includes the raw query, so two requests differing only by query (`?id=1` vs `?id=2`) under one `Idempotency-Key` are detected as distinct ("different request" 409) rather than swallowed as a replay ("already forwarded").
- **Approval-gated routes** fail closed on query-bearing requests. The approval hold/replay path reconstructs the request from the stored path only and cannot yet round-trip the query, so a query-bearing approved request would otherwise execute a semantically different (query-stripped) request than the one approved. A full query-preserving approval replay is a follow-up.

## Security invariants (tested)

scheme/host/port pinned to the configured upstream; userinfo/fragment rejected; query cannot influence route matching; duplicate keys keep ordered pair semantics; percent-encoding not decoded/re-encoded; empty query, blank values, repeated keys, and a malicious absolute-URL-looking query value all covered.

## Tests

- `query-forwarding.test.ts` — unit coverage of the URL builder (one key, repeated keys, blank value, percent-encoded value, absent query, URL-looking value stays inert data, base-query / userinfo / fragment rejection, input not mutated).
- `proxy-query-preservation.test.ts` — end-to-end proxy integration (real auth + PGLite + vault, network stubbed): forwarding, dup keys, blank, encoding, no host/scheme/path bypass, no route-matching influence, idempotency distinctness (different query vs genuine replay), approval-route rejection + no-query hold still works.

**Effectiveness proven** by reverting each production change independently and observing the corresponding cases fail (query preservation cases → empty search; idempotency fingerprint → "already forwarded" instead of "different request").

`bun test` (both new files 28/28 standalone), `tsc --noEmit`, and biome all clean. `codex review --uncommitted` returned no findings after the follow-through fixes.

Do not merge.

[sol-orch]

### DIFF
--- a/packages/proxy/src/__tests__/proxy-query-preservation.test.ts
+++ b/packages/proxy/src/__tests__/proxy-query-preservation.test.ts
@@ -1,0 +1,399 @@
+/**
+ * Governed-request query-preservation integration tests.
+ *
+ * The proxy handler resolves the upstream target from `c.req.path`, which Hono
+ * strips of any query string. Before the fix this meant a governed request such
+ * as `GET /github/search/issues?q=foo&page=2` was forwarded upstream as
+ * `/search/issues` with the query silently dropped — a correctness bug that
+ * breaks paginated / filtered APIs and can change request semantics.
+ *
+ * These tests drive the real `handleProxy` end-to-end (auth middleware + PGLite
+ * DB + vault, network stubbed) and assert that:
+ *   - the exact raw query string is forwarded to the upstream,
+ *   - duplicate keys keep their ordered pair semantics (no Object collapse),
+ *   - blank values, absent query, and percent-encoding are preserved verbatim,
+ *   - a query value that looks like an absolute URL stays inert data and never
+ *     rewrites the pinned scheme/host/port/path,
+ *   - the query cannot influence route matching, and
+ *   - userinfo / fragment cannot ride in through the query.
+ *
+ * Security invariant: preserving the query must not enable authority bypass or
+ * target confusion. scheme/host/port/path remain pinned to the configured
+ * upstream regardless of query contents.
+ */
+
+import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
+import { signAgentToken } from "@stwd/auth";
+import { agents, closeDb, getDb, tenants } from "@stwd/db";
+import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { SecretVault } from "@stwd/vault";
+import { Hono } from "hono";
+import { PROXY_SCOPE } from "../config";
+
+setDefaultTimeout(30000);
+
+const MASTER_PASSWORD = "proxy-query-preservation-master";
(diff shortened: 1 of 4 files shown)


0xSolace:
[sol-orch] SHIP-WITH-NITS review receipt

Adversarial merge gate completed at tip `3a8a00cd346d6e0dfd1b6973b3d1bace40bdfcff`.

Verified:
- Reproduced query loss on `develop` against the same stubbed forwarding boundary, then confirmed exact query preservation at the PR tip.
- Actual local stub upstream received one key, ordered repeats, blanks, `%2F`/`%25`, literal `+` versus `%20`, UTF-8 percent bytes, malformed percent escapes, encoded CRLF, and absent query byte-for-byte.
- Scheme, host, port, and path stayed pinned. Absolute-URL-looking values remained query data. Existing target query, userinfo, and fragments reject deterministically.
- Route resolution remains path-only and precedes query application.
- Raw query participates in idempotency identity. Different query conflicts under one key, identical raw query replays, and `%2F` versus `%2f` conflicts as intentionally raw-byte-distinct.
- Approval-gated query requests return stable HTTP 400 before `pending_proxy_requests` or execution-authorization nonce creation. Queryless approval hold/replay remains green.
- Query application removal and fingerprint-query omission mutations were both caught by focused tests.
- Focused tests: 28/28. Expanded adversarial review harness: 14/14. Approval lifecycle standalone: 5/5. GitHub credential route standalone: 5/5. Proxy typecheck and Biome clean.
- All required CI checks green at tip.

Nits, non-blocking:
- The committed approval rejection test proves 400/no forward but does not directly assert zero pending rows/nonces. The review harness did.
- The committed matrix could explicitly retain `+` versus `%20`, UTF-8 percent bytes, malformed escapes, encoded CRLF, and byte-distinct equivalent-query cases to lock the broader contract.

Squash merged as `9a45e1a22dafa928085f10d263b0c34bdb07ac4c`.

PR #191, opened 2026-07-14
feat(authority): provider-action canonicalization and action binding (authority plan PR2)

0xSolace:
## Summary

Authority Plan **PR2** — canonical provider actions + action binding. Adds a GitHub provider-action adapter, one shared RFC-8785 canonicalizer + request-hashing path, a runtime-neutral provider-principal seam, separate typed access + exact-action policy decisions persisted 1:1 with `intents`, and an in-process executor stub. **No** credential decrypt, proxy dispatch, execution-authorization mint, or trading-route change (that is PR4).

Branched from PR1 (`#186`), rebased cleanly onto `origin/develop` (`7f23dee`, which merged `#187` + `#189` + `#190`). `[sol-orch]`

## Spec mapping

| Spec | Where |
|---|---|
| §2 `POST /v2/provider-actions` (bounded raw bytes, strict dup-key parse, exact shape, forged-actor rejection) | `packages/api/src/routes/provider-actions.ts` |
| §2.2 operation schemas (`github.issue.list`, `github.pr.comment.create`) + canonical action construction + safe_summary | `packages/provider-github/` (new `@stwd/provider-github`) |
| §3 `github.provider-action.v1` canonicalizer + 17 golden vectors + one shared JCS | `packages/shared/src/provider-action.ts` (+ new `canonicalizeRawInternalAction` composed path) |
| §5 `provider_action_bindings` schema + immutability trigger | `packages/db/drizzle/0080_*.sql`, `schema.ts` (from PR1-lane commits) |
| §6.1/§6.2 distinct access + policy decision documents (separate IDs/hashes) | `provider-action-service.ts` |
| §6.3 `composeProviderActionPolicyDecision` (hard_deny > approval_required > allow > default-deny) | `packages/policy-engine/src/capability-intent.ts` |
| §6.4 one-tx persist (intent → access → policy → binding → required-audit outbox → commit) then stub | `provider-action-service.ts` + new `provider_action_audit_outbox` |
| §7.1 `ProviderPrincipalV1` seam | `packages/api/src/middleware/provider-principal.ts` + `VerifiedAgentPrincipal` on `AppVariables` |
| §7.2 `authenticateAgentJwt` / `installAgentJwtContext` / `requireAgentJwt` (unchanged wire) / `requireProviderAgentJwt` (no trade:order) / new plugin-context field | `packages/api/src/middleware/agent-jwt.ts`, `plugin.ts` |
[...]

PR #189, opened 2026-07-14
fix(api): commit approval decision transitions and audit events atomically

0xSolace:
## Summary

The PR #181 proxy approval flow committed its **state transitions** and their **audit events** in SEPARATE transactions. A crash or audit failure between them could leave an approved/denied/expired approval with **no audit record** (or, on `/:txId/deny`, a best-effort compensating rollback that could itself be interrupted). This violates the PR3 spec's *evidence-before-visibility* invariant **I14** and is the exact gap flagged in the spec's repository-gaps section:

> **item #10:** "PR #181 approval mutation and audit are separate transactions. Approve commits before `writeApprovalAudit`; an audit failure can leave approved state. Its expiry loop also updates first and audits row-by-row afterward. PR3 requires atomic state plus audit."

This is an **ordering/atomicity-only** fix: no route contracts, response shapes, or approval semantics change.

## The gap (verified)

`writeAuditEvent` always opens its own transaction (spec item #15), so every proxy path did `UPDATE …; commit` then `writeApprovalAudit()` in a second transaction:
- `POST /approvals/proxy/:id/approve` — approve commits, then audit.
- `POST /approvals/proxy/:id/deny` — deny commits, then audit.
- `expireProxyApprovals` — bulk expire commits, then per-row audit in a loop.
- `POST /approvals/:txId/deny` — state change in its own tx, then post-commit audit + fragile compensating rollback.

## Fix

**`services/audit.ts`** — add the transaction-aware audit primitive the spec prescribes (section 7.1), *not* an outbox:

```ts
withTenantAuditedTransaction(tenantId, async (tx, appendRequiredAudit) => { … })
```

It runs one `db.transaction` under the existing per-tenant advisory lock + in-process serialization queue, and appends audit rows + advances the `audit_chain_heads` high-water-mark **inside the caller's transaction**, so the state change and its audit event are both-or-neither. The standalone `writeAuditEvent` path is refactored to share the same append core (`appendAuditEventWithinTx`), preserving canonical HMAC, redaction, and serialization/seq-conflict retry semantics.

[...]

PR #192, opened 2026-07-15
feat(authority): exact-request approval binding and safe resume (authority plan PR3)

0xSolace:
## Summary

PR3 turns the PR2 `approval_required` provider action into a **recent-human, exact-request approval** and a **safe `steward-system` resume**, rooted in the existing `intents` + `approval_queue` + `provider_action_bindings` (no `provider_action_approvals` table — spec §1). It does **not** decrypt credentials, call the proxy, mint execution authorization, claim a nonce, or dispatch (I15) — that is PR4.

Base: `develop` @ `a1dcd75` (PR2 #191). Migration `0081` (journal max verified = 0080 before, 0081 after).

## Spec mapping

- **§4 schema / migration 0081**: `approval_queue` discriminated union (transaction | provider_action arm) with arm CHECK + decision-shape CHECK + partial unique indexes; `provider_action_bindings` lifecycle columns + the PR3 transition trigger (transition graph, per-state field-shape CHECK, `binding_revision` increment-by-one) replacing the PR2 immutability trigger; enum `expired`/`stale`/`consumed`.
- **§5 exact commitment**: `packages/shared/src/provider-approval.ts` builds `ProviderApprovalCommitmentV1` + hash with the strict RFC 8785 JCS PR2 owns; binds request/action hashes, access + policy decisions, **all** matched grants/bindings + revisions, operation/account revisions, route+credential (incl. `authority_revision`), actor/workspace, approval-policy revision. §5.3 integrity recompute (canonical bytes → digest, envelope → request hash, access/policy/commitment docs → hashes, cross-table agreement).
- **§6 state machine**: `none→pending_approval→approved/denied/expired/stale→execution_ready`, exclusive guarded CAS transitions, DB-time expiry (`<= now()`), staleness, safe resume (approved→execution_ready, consumes approval once, no execution).
- **§7 audit**: every transition + required event commit atomically via `withTenantAuditedTransaction` (#189). Taxonomy per §7.2 + **C1** top-level `resource_type='provider_action'`, `resource_id=intents.id`.
- **§8 idempotency**: decision replay (exact retry returns existing; changed body → conflict; cross-action reuse → 409), resume single-winner.
- **§9 HTTP**: `GET/POST /v2/provider-actions/:id/approval`, `POST /v2/provider-actions/:id/execute`, exact codes, structured error bodies, non-enumerating 404s, resume-actor-substitution rejection.
- **G1 (PR4 adjudication)**: `secret_routes.authority_revision` (int NOT NULL DEFAULT 1 CHECK > 0) + `steward_bump_secret_route_authority_revision()` BEFORE UPDATE trigger folded into **0081** (not PR4's 0082); commitment binds `authority_revision` directly; trigger-existence CI test added.
- **C1 (PR5 adjudication)**: all provider-action lifecycle audit events (PR3's + the PR2 service writes touched) set top-level `resource_type='provider_action'`, `resource_id=intents.id`.
- **C2 (PR5 adjudication)**: crash-recovery sweeper `recoverUnsignedIntents` closes the "commit-then-drain" gap. Exactly-once, crash-safe (keys on the audit chain itself under the per-tenant audit queue), with a regression test (kill drain → recover → signed event exists exactly once, idempotent under concurrent sweeps).

## Tests

- **Lifecycle** (7): create/approve/deny/expire/stale/resume/terminal.
- **Negative matrix** (29): N04/05/07/08/09/10/11/12/13/17/20/21/22/23/24/26/30/31/34/35/39/40/41/42/43/44/45/47 + Client-B isolation + both codex P2 idempotency regressions.
- **Concurrency/fault** (8): C01 (two approves), C02 (approve+deny race), C05 (25 concurrent execute → one resumeAttemptId + one ready event), C06/C08 (dependency change before claim stales), C09/C10/N49 (audit fault rolls back the whole tuple), C2 crash-recovery (exactly-once + idempotent).
- **Route HTTP** (6): N01/N15/N16, happy approve→execute, GET MFA gating + safe-summary.
- **Migration invariants** (6): authority_revision + bump trigger (G1), PR3 binding trigger, approval_queue arm/shape CHECKs, enum labels, legacy-pending backfill.
- **Static boundary** (7): I15 import guard (no decrypt/proxy/mint/nonce/network/legacy fallback).
- **Mutation proofs** (`scripts/pr3-mutation-proofs.sh`): **6/6 killed** — M1 MFA window (N12), M2 ambient role (N04), M3 queue/binding hash agreement (N26), M4 secret version (N39), M5 canonical digest recompute (N24), M6 non-idempotent resumeAttemptId (C05). Each: baseline pass → mutation → named test fails → restore.

Existing approvals/intents/audit/provider-authority/vault suites remain green (regression-checked, 35+ files).

## Codex review

Two full passes; all findings addressed:
- [P1] migration backfill of legacy pending rows before the shape CHECK — fixed + tested.
- [P1] crash-between-commit-and-drain evidence loss — fixed (crash-safe exactly-once sweeper).
- [P1] same-tx new-enum-value use (`ALTER TYPE ADD VALUE` + CHECK referencing it) — fixed via `status::text` in the decision-shape CHECK.
- [P2] non-enumeration on GET approval detail — ineligible → 404.
- [P2] stale idempotent replay after expiry/stale — replay only when decided.
- [P2] cross-action idempotency-key reuse → 409 not 503.

## Deviations

- **requester separation source**: PR1 does not yet ship a structured operation approval-requirements field, so PR3 reads `request_profile.approvalRequirements.requesterSeparation` (default false), documented in `extractRequesterSeparation`. Swap to the first-class field when PR1 lands it.
- **G1 option (a)**: `secret_routes.authority_revision` + trigger live in 0081 per the orchestrator ruling (PR4's 0082 keeps `authority_mode`/`provider_operation_id`).
- **approved→expired/stale queue rows** clear `decision`/resolution fields to satisfy the decision-shape CHECK; the immutable decision evidence is preserved on the binding (`approval_actor_user_id`/`approved_at`) + the signed `provider.approval.decided` audit event.

## Spec contradictions surfaced (not silently resolved)

None blocking. The G1/C1/C2/C3 cross-PR contracts were already adjudicated in the PR4/PR5 spec appends and implemented accordingly. The decision-shape CHECK requiring `decision IS NULL` for expired/stale is in tension with I3's "decision never changes"; resolved by treating the binding + audit chain as the immutable decision record and the queue decision fields as clearable on terminal transition (flagged here for the orchestrator).

Do NOT merge — orchestrator gate.

[sol-orch]

### DIFF
--- a/packages/api/scripts/pr3-mutation-proofs.sh
+++ b/packages/api/scripts/pr3-mutation-proofs.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+# PR3 mutation-strength proofs (spec §14). Each mutation weakens ONE security
+# predicate; a proof is valid iff the named test PASSES clean AND FAILS after the
+# mutation. The file is restored after each proof. Run from packages/api:
+#
+#   bash scripts/pr3-mutation-proofs.sh
+#
+set -uo pipefail
+cd "$(dirname "$0")/.."
+
+SVC="src/services/provider-approval.ts"
+NEG="src/__tests__/provider-approval-negative.test.ts"
+CONC="src/__tests__/provider-approval-concurrency.test.ts"
+
+pass_count=0
+fail_count=0
+
+# run_test <file> <filter> -> 0 if all pass (0 fail), 1 otherwise
+run_test() {
+  local out
+  out=$(timeout 90 bun test --timeout 25000 "$1" -t "$2" 2>&1)
+  echo "$out" | grep -qE "^ *0 fail$"
+}
+
+# proof <name> <file> <filter> <target> <sed-expr>
+proof() {
+  local name="$1" file="$2" filter="$3" target="$4" sedexpr="$5"
+  echo "=== PROOF: $name ==="
+  if run_test "$file" "$filter"; then
+    echo "  baseline PASS ✓"
+  else
+    echo "  baseline UNEXPECTED FAIL ✗ (proof invalid)"; fail_count=$((fail_count+1)); return
+  fi
+  cp "$target" "$target.bak"
+  sed -i "$sedexpr" "$target"
+  if run_test "$file" "$filter"; then
+    echo "  post-mutation still PASSES ✗ (mutation did not kill the test)"; fail_count=$((fail_count+1))
+  else
+    echo "  post-mutation FAILS ✓ (predicate killed)"; pass_count=$((pass_count+1))
+  fi
+  mv "$target.bak" "$target"
+}
+
+# M1: loosen MFA window (<=5m -> <=6m) → N12 (5m+1ms) must fail.
+proof "M1 loosen MFA window (N12)" "$NEG" "N12" "$SVC" \
+  's/const MAX_MFA_AGE_MS = 300_000;/const MAX_MFA_AGE_MS = 360_000;/'
+
+# M2: accept ambient role (skip approver-binding requirement) → N04 must fail.
+proof "M2 accept ambient role (N04)" "$NEG" "N04" "$SVC" \
+  's/const eligible = approverRows.some(/const eligible = true || approverRows.some(/'
+
+# M3: skip queue/binding request-hash + digest agreement → N26 must fail.
+proof "M3 skip queue/binding hash agreement (N26)" "$NEG" "N26" "$SVC" \
+  's/if (queue.requestHash !== binding.requestHash || queue.actionDigest !== binding.actionDigest) {/if (false) {/'
+
+# M4: ignore secret version at resume → N39 must fail.
+proof "M4 ignore secret version (N39)" "$NEG" "N39" "$SVC" \
+  's/account.credentialVersion !== c.executionDependencies.secretVersion/false/'
+
+# M5: skip the canonical-byte digest recomputation (accept any stored digest) →
+#     N24 (tampered canonical bytes) must fail. Weaken the digest equality guard.
+proof "M5 skip canonical-byte digest recompute (N24)" "$NEG" "N24" "$SVC" \
+  's/if (recomputedDigest !== binding.actionDigest) {/if (false) {/'
+
+# M6: mint a fresh resumeAttemptId on the idempotent return path → C05 must fail.
+proof "M6 non-idempotent resumeAttemptId (C05)" "$CONC" "C05" "$SVC" \
+  's/resumeAttemptId: binding.resumeAttemptId ?? undefined,/resumeAttemptId: randomUUID(),/'
+
+echo ""
+echo "==================================================="
+echo "MUTATION PROOFS: $pass_count killed, $fail_count invalid"
+echo "==================================================="
+[ "$fail_count" -eq 0 ] && exit 0 || exit 1
--- a/packages/api/src/__tests__/provider-action-service.test.ts
+++ b/packages/api/src/__tests__/provider-action-service.test.ts
@@ -27,6 +27,8 @@ import {
   providerActionBindings,
   providerGrants,
   providerOperations,
+  secretRoutes,
+  secrets,
   tenants,
   users,
   workspaces,
@@ -51,6 +53,8 @@ const OP_A_READ = "40000000-0000-4000-8000-000000000001";
 const OP_A_WRITE = "40000000-0000-4000-8000-000000000002";
 const OP_B_READ = "40000000-0000-4000-8000-000000000003";
 const GRANTOR = "10000000-0000-4000-8000-000000000001";
+const SECRET_A = "50000000-0000-4000-8000-000000000001";
+const ROUTE_A = "60000000-0000-4000-8000-000000000001";
 const FUTURE = new Date(Date.now() + 365 * 24 * 3600_000);
 
 function principal(agentId = AGENT): ProviderPrincipalV1 {
@@ -114,6 +118,32 @@ async function seedCore() {
       createdBy: GRANTOR,
     },
   ]);
+  // A secret + route so the PR3 approval commitment can bind route/credential
+  // revisions (spec §5.2: missing route/credential fails approval creation).
+  await db.insert(secrets).values([
+    {
+      id: SECRET_A,
+      tenantId: TENANT,
+      name: "github-a",
+      ciphertext: "x",
+      iv: "x",
+      authTag: "x",
+      salt: "x",
+      version: 1,
+    },
+  ]);
+  await db.insert(secretRoutes).values([
+    {
+      id: ROUTE_A,
+      tenantId: TENANT,
+      secretId: SECRET_A,
+      hostPattern: "api.github.com",
+      pathPattern: "/*",
+      method: "*",
+      injectAs: "header",
+      injectKey: "authorization",
+    },
+  ]);
   await db.insert(providerAccounts).values([
     {
       id: ACCOUNT_A,
@@ -122,6 +152,8 @@ async function seedCore() {
       adapterKey: "github",
       externalRef: "a",
       displayName: "A",
+      credentialSecretId: SECRET_A,
+      credentialVersion: 1,
     },
     {
       id: ACCOUNT_B,
@@ -140,6 +172,7 @@ async function seedCore() {
       providerAccountId: ACCOUNT_A,
       operationKey: "github.issue.list",
       riskClass: "read",
+      secretRouteId: ROUTE_A,
     },
     {
       id: OP_A_WRITE,
@@ -148,6 +181,7 @@ async function seedCore() {
       providerAccountId: ACCOUNT_A,
       operationKey: "github.pr.comment.create",
       riskClass: "consequential",
+      secretRouteId: ROUTE_A,
     },
     {
       id: OP_B_READ,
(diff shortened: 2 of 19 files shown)


0xSolace:
---

### Codex review — additional rounds (all resolved)

Four completed codex passes; every finding fixed + regression-tested + pushed:

Round 1 (P1/P1/P2): migration backfill of legacy pending rows; crash-between-commit-and-drain evidence loss; GET-detail non-enumeration.
Round 2 (P1/P2): same-tx new-enum-value use (fixed via `status::text`); stale idempotent replay after expiry; cross-action idempotency-key reuse → 409.
Round 3 (P1/P2): approver-binding environment scope enforcement; expiry/stale CAS-loser double-audit guard.
Round 4 (P1/P1): create-replay of a governed action reports 202 APPROVAL_REQUIRED (never POLICY_ALLOW); workspace-authority revision compared against the committed access-decision revision.

Test counts: lifecycle 7, negative 32, concurrency 9, route 6, boundary 7, migration 6. 6/6 mutation proofs killed. All 12 PR3+adjacent api files + all db migration files green; 35+ adjacent audit/authority/vault suites unaffected.

0xSolace:
## [sol-orch] Adversarial review + merge gate — VERDICT: **SHIP-WITH-NITS** ✅

Independent verification (behavior, not claims) at exact tip `b4c41e2df5dbd94d58597a794f32fe6770173239` vs `develop@a1dcd75` (merge-base == develop tip, no rebase). All required CI green at the exact SHA (14/14, Mintlify skipped). Reviewed read-only in an isolated worktree; no code changed.

### What I ran (not just read)
- **PR3 suites** (isolated runner): lifecycle, negative (32), concurrency (9), route (6), boundary (7) — **all pass**.
- **Migration invariants** (`exact-approval-migration.test.ts`) — 6/6 pass on PGLite.
- **PR2 service** (`provider-action-service.test.ts`) + **provider-authority** — pass (C1/C2 wiring intact).
- **Legacy regression** (the discriminated-union touches `approval_queue`): `approvals.test`, `approvals-audit-order`, `approval-principals`, `audit-chain`, `audit-chain-strict`, `tenant-audit-rollback` — **all pass**. The transaction arm is unbroken.
- **Their 6 mutation proofs** (`scripts/pr3-mutation-proofs.sh`): **6/6 killed, 0 invalid** (M1 MFA window, M2 ambient role, M3 queue/binding hash agreement, M4 secret version, M5 canonical digest, M6 non-idempotent resumeAttemptId). Service restored pristine after.
- **My 2 additional mutation proofs on guards they did NOT mutation-test** — both **killed**:
  - MINE-1: neutralize the committed-vs-current `workspace.revision` binding → the "workspace authority revision bumped (still active) ⇒ APPROVAL_DEPENDENCY_STALE" test fails. Guard is real.
  - MINE-2: force the approver **environment-scope** predicate always-true → the "staging-scoped approver cannot approve a production action" test fails. Guard is real.
- **4 direct-SQL adversarial DB probes** (temp, removed): transition trigger has NO terminal state on the OLD side of any allowed arm (no revival), +1 monotonic, frozen columns; arm CHECK; decision-shape CHECK pins `consumed_by='steward-system'` + `decision IS NULL` on expired/stale; one-live-approval partial unique on `intent_id`. **All pass.**
- **F-ratification probe** (temp, removed): empirically proved approved→expired preserves `approvalActorUserId`+`approvedAt` on the binding, the signed `provider.approval.decided` event persists, the queue `decision` clears, and the expired queue **cannot re-open into a fresh approval**. **Pass.**
- `bunx turbo run build --filter=@stwd/api...` (dependency-aware typecheck) — **17/17 green**. Biome on all 17 diff files — **clean**.

### Spec conformance (behavior-verified)
- **A — Migration 0081**: raw-SQL (hand-written, correct — triggers/CHECKs drizzle can't express). Transition graph == spec §6.1 exactly (`pending_approval → {approved, denied, expired, stale}`, `approved → {execution_ready, expired, stale}`, PR2 stub lineage exempt from revision bump). `binding_revision` +1 monotonic + frozen PR2 columns enforced by trigger. Arm/decision-shape CHECKs + partial unique indexes enforced. G1 `secret_routes.authority_revision` col + bump trigger present, fires on every bound-field change, no-op on non-bound. Legacy `pending_approval` backfill → `approval_stale`/`canceled` BEFORE the shape CHECK (forward-safe). Journal idx 81 = max. Enum-safe `::text` in the decision-shape CHECK (avoids same-tx new-enum bind). ✓
- **B — Commitment (§5)**: binds request/action hashes, access+policy decisions, all matched grants/bindings + revisions, account/operation revisions, route (`authority_revision`) + credential, actor/workspace, expiry. Integrity recompute is from **raw persisted bytes** (not re-parsed). Workspace revision now bound (round-3). Mutations M3/M4/M5 + MINE-1 all killed. ✓
- **C — State machine / concurrency**: exclusive guarded CAS, `FOR UPDATE` stable lock order, expiry-first with DB `now() <=`. C13 CAS-loser **double-audit fixed and verified** (3 racing stales ⇒ exactly 1 staled event). Single-winner resume (M6 killed). Double-execute idempotent. ✓
- **D — AuthZ**: exact workspace approver **incl. environment scope** (MINE-2 killed); requester separation read from the **committed operation policy profile** (`request_profile.approvalRequirements.requesterSeparation`), not caller input; MFA 5-min window (M1 killed); resume uses the **recorded** approver, never the caller (no actor substitution); GET non-enumeration collapses eligibility failures to 404. ✓ (see P2-2 for execute breadth)
- **E — C1/C2**: `resource_type='provider_action'`, `resource_id=intents.id` on lifecycle events incl. retrofitted PR2 writes. C2 sweeper keys existence on the audit chain itself under the per-tenant queue → exactly-once + crash-safe; verified by test. ✓
- **F — decision-shape vs I3**: **RATIFIED.** The spec's own §4.1 CHECK text (line 207) *mandates* `decision IS NULL` on expired/stale, so the impl matches spec verbatim. I3 immutability is satisfied because the decision evidence lives on the binding + the permanent signed audit event (empirically proven), and laundering is structurally impossible (no transition out of terminal states; denials use `rejected` status, outside the expiry/stale CAS set). ✓
- **G — Error model**: hostile bodies → 400 at the route; service catch-alls return 503 codes, never a raw 500; no secret/payload leakage in error bodies (safe messages, never echo dependency IDs). ✓

### Findings
**P2-1 (code fidelity, not security) — audit-unavailable code on decide/resume.** `AuditUnavailableError` is thrown ONLY by the test fault hook. A real audit-signer/HMAC/append failure inside `decide`/`resume` throws a generic `Error` (from `getHmacKey()`/insert in `audit-chain.ts`), which the service maps to `APPROVAL_PERSISTENCE_FAILED` (503) — **not** spec §9/N49's `EVIDENCE_REQUIRED_AUDIT_UNAVAILABLE` (503). The N49 test passes but exercises a class the production path can't throw. **Impact: none to safety** — the full lifecycle tuple still rolls back and returns 503 (fail-closed, I14 holds; verified). Only the error *code* differs. Note the **create path already maps this correctly** (`provider-action-service.ts` → `EVIDENCE_REQUIRED_AUDIT_UNAVAILABLE`); only decide/resume differ. *Fix:* map audit-layer failures to `EVIDENCE_REQUIRED_AUDIT_UNAVAILABLE` (e.g., have the audit primitive throw a typed error, or detect it in the catch). Pre-existing-adjacent (audit layer throws generic Error by design). **Not merge-blocking.**
`provider-approval.ts:1308,1481`

**P2-2 (authZ breadth) — execute route accepts any tenant principal.** Spec §9.1 (line 678): "POST execute permits the original agent **or a current workspace approver/admin**." `handlePostExecute` authorizes on `tenantId` presence alone (any agent, any human, api-key, app-secret), never checking caller == `binding.actorAgentId` nor caller ∈ {workspace_approver, admin}. **Impact: low** — resume performs no execution (I15 holds), fully re-validates integrity + all dependencies + the *recorded* approver's eligibility, fixes the resume actor to `steward-system` (N16 enforced), and is idempotent single-winner. A non-approver/non-original tenant member can at most advance an already-human-approved action to `execution_ready` and observe a minimal result shape. *Fix:* in the execute handler, require the authenticated principal to be either the original agent for this intent or a workspace approver/admin before calling `resume`. **Not merge-blocking** given the safety envelope; file as fast-follow. `routes/provider-approvals.ts` (`handlePostExecute`).

**P3 — execute non-enumeration.** Subset of P2-2: cross-scope tenant callers get a distinguishable execute outcome. Same fix as P2-2.

No P0/P1. Scope discipline held (no PR4 dispatch/claim/execution creep — I15 static boundary test proves no decrypt/proxy/mint/nonce/network imports). N50 generic-intent-status bypass blocked.

### Adjudication answers
- **D (requester-separation default):** **default-false is spec-compliant, NOT a fail-open.** Spec I10 is explicitly conditional — "separation *where configured*"; it applies only "IF the operation's PR1 approval policy sets `requesterSeparation=true`." The value is read from the **committed, server-side operation policy profile** (not attacker-controllable) and bound into the commitment at creation, re-checked from the commitment at decision/resume. An operator who never opts in gets no separation *by the spec's own design*. Sound.
- **F (decision-shape vs I3):** **ratified in the PR's favor.** The spec §4.1 CHECK text itself requires `decision IS NULL` on expired/stale; the binding fields + signed `provider.approval.decided` audit event are the immutable record and both provably survive the terminal reclassification; a cleared queue cannot be laundered into a fresh approval (structurally impossible + empirically tested). Matches spec verbatim.

Proceeding to squash-merge (SHIP-class + exact-tip green + mergeable; `REVIEW_REQUIRED` is the only block, `enforce_admins=false`). P2s filed as fast-follow, non-blocking.

Issue #195, opened 2026-07-16
feat: X (Twitter) OAuth provider support (governed agent actions on X)


0xSolace:
## Goal

Add X (Twitter) as a **governed provider** in Steward: workspaces connect an X account via OAuth 2.0, agents propose X actions (post, delete, read profile) through the provider-action pipeline, policy + approval gate them, and the proxy executes with vault-held tokens the agent never sees.

## What already exists (do not rebuild)

- **Login-with-X for user sign-in** already works: `packages/auth/src/oauth.ts` has a built-in `twitter` provider (PKCE, synthetic email handling). That is USER AUTH. This issue is about **provider connections** (agent execution authority), a different plane.
- Provider authority substrate (PR1-PR3, in-flight PR4): `provider_accounts` (with `credential_secret_id`/`credential_version` vault linkage), `provider_operations`, `secret_routes` (+`authority_mode`), canonicalization (`@stwd/shared` JCS + action digest), approval binding (`provider-approval.ts`), execution authorization v2 (HMAC commitment, single-use nonce), `@stwd/provider-github` as the reference adapter.

## Workstream A — X account connect + token lifecycle (api/auth side)

- OAuth 2.0 authorization-code + PKCE (S256) flow against `x.com/i/oauth2/authorize` + `api.x.com/2/oauth2/token`, scoped to **provider account connect** (NOT user sign-in): initiate + callback routes under `/v2/provider-accounts/connect/x`, workspace-scoped, admin/approver-only.
- Requested scopes minimal + explicit: `tweet.read tweet.write users.read offline.access` (configurable subset per connect).
- Tokens stored as a **versioned vault secret**; `provider_accounts.credential_secret_id/credential_version` links it. Access token + refresh token in the secret payload; scopes granted + X user id/username recorded on the account (`external_ref` = X user id).
- **Refresh service**: X access tokens live ~2h; refresh tokens are SINGLE-USE (rotating). Refresh must be serialized per account (advisory lock / SELECT FOR UPDATE on the account row) so concurrent refreshes cannot clobber the rotated token. Failed refresh (revoked upstream) => account `status` degraded + audit event, never silent.
- Pending connect state: prefer Redis-backed short-TTL state (hashed state + PKCE verifier, mirroring the hardened user-auth flow). **Avoid a new migration if possible** — migration 0082 is owned by in-flight PR4; if a table is truly required it takes 0083 AFTER PR4 merges.
[...]

PR #196, opened 2026-07-16
feat(provider-x): X provider-action adapter + canonical profile (#195 workstream B)

0xSolace:
## Summary

Workstream B of #195: the pure `@stwd/provider-x` adapter + the shared `x.provider-action.v1` canonical profile. No DB, no routes, no OAuth flow, no proxy changes (workstreams A and C own those).

The adapter owns operation argument schemas, canonical action construction, safe summaries, and validated policy args. It imports the ONE shared canonicalizer surface and **never re-implements JCS, hashing, path/query/method normalization, or the content-type matrix**. The X canonical action is structurally identical to the GitHub one (same field set), so the shared JCS serializer produces byte-identical framing and the action digest is computed on the exact same `sha256HexPrefixed(jcsStringify(...))` path.

## Operations

| operation | method | path | body | risk |
|---|---|---|---|---|
| `x.tweet.create` | POST | `/2/tweets` | `{text}` + `{reply:{in_reply_to_tweet_id}}` when replying | write (approval-worthy) |
| `x.tweet.delete` | DELETE | `/2/tweets/{id}` | none | write (approval-worthy) |
| `x.user.me.read` | GET | `/2/users/me?user.fields=id,name,username` | none | read |

### Argument schemas (fail-closed, stable CANON_* codes)
- `x.tweet.create`: `{text: string, replyToTweetId?: string}`. `text` trimmed then bounded 1..280 **code points**; `replyToTweetId` must match `/^[0-9]{1,25}$/`. The `reply` body key appears only when `replyToTweetId` is present.
- `x.tweet.delete`: `{tweetId: string}`, `/^[0-9]{1,25}$/`. The path segment is built from the VALIDATED id (no raw concatenation, no percent-encoding needed, traversal impossible).
- `x.user.me.read`: no args (accepts absent/empty object); any field is `CANON_UNKNOWN_FIELD`. Fixed single-pair query in canonical order.
- Unknown fields -> `CANON_UNKNOWN_FIELD`; type errors -> `CANON_FIELD_TYPE_INVALID`; bad ids -> `CANON_PATH_SEGMENT_INVALID`; empty-after-trim text -> `CANON_BODY_REQUIRED`; over-length text -> `CANON_FIELD_TYPE_INVALID`; lone surrogate -> `CANON_UNICODE_INVALID`.

## Shared profile changes (minimal + additive)
[...]

PR #193, opened 2026-07-15
fix(authority): PR3 review nits — audit-unavailable error fidelity and execute-route authz

0xSolace:
## Summary

Closes the post-merge PR3 review nits from #192 without adding PR4 behavior.

### P2-1: audit-unavailable error fidelity

- Maps real required-audit append failures on both approval decision and resume to `EVIDENCE_REQUIRED_AUDIT_UNAVAILABLE`.
- Preserves HTTP 503, fail-closed behavior, and full transaction rollback.
- Preserves audit sequence-conflict retry classification by forwarding the underlying database code/message through the typed audit error.
- Adds real HMAC/audit-persistence failure regression tests for decision and resume, including exact code and rollback assertions.
- The existing create-path coverage remains green in `provider-action-service.test.ts`.

### P2-2 and P3: execute-route caller authority and non-enumeration

- Restricts `POST /v2/provider-actions/:id/execute` to the original requesting agent or a current human `workspace_approver` / `workspace_admin` for the exact workspace and environment.
- Reuses the PR3 workspace-role authority predicate and locks membership/workspace/role rows while rechecking authority inside the resume transaction.
- Keeps agent-token and human-session identities disjoint, so an agent token cannot borrow an embedded human claim.
- Returns the GET route's `404 SCOPE_RESOURCE_NOT_FOUND` non-enumeration shape for unrelated and cross-tenant callers.
- Adds coverage for original agent, workspace approver, workspace admin, unrelated tenant principal, cross-tenant principal, GET/execute shape parity, mixed agent/human claims, and the transactional recheck.

## Mutation proof

- Audit mapping weakened to rethrow the raw persistence error: `P2-1 decide: real audit persistence failure returns exact audit-unavailable code and rolls back` and `P2-1 resume: real audit persistence failure returns exact audit-unavailable code and rolls back` both failed with `APPROVAL_PERSISTENCE_FAILED`; mapping restored and both pass.
- Execute authority weakened to accept any human caller: `P2-2: resume transaction rechecks the execute caller and rejects unrelated principals` failed by transitioning to `execution_ready`; guard restored and the test passes.

## Verification

- PR3 focused isolated API runner: 7/7 files passed
  - `provider-action-service.test.ts`
  - `provider-approval-boundary.test.ts`
  - `provider-approval-concurrency.test.ts`
  - `provider-approval-lifecycle.test.ts`
  - `provider-approval-negative.test.ts`
  - `provider-approval-route.test.ts`
  - `provider-authority.test.ts`
[...]

PR #197, opened 2026-07-16
feat(provider-x): X OAuth provider-account connect + token lifecycle (#195 workstream A)

0xSolace:
## Scope: issue #195 Workstream A — X (Twitter) provider-account OAuth connect + token lifecycle

The api/auth side of governing X as a provider: a workspace connects an X account via OAuth 2.0 (auth-code + PKCE S256), tokens land in the versioned vault, and the account lifecycle (connect / refresh / disconnect) is fully audited. This is the **provider-connection plane** (agent execution authority), deliberately SEPARATE from login-with-X (user sign-in). Does NOT build the `@stwd/provider-x` adapter (workstream B) or touch proxy/execution-authorization (in-flight PR4).

### Scope map
- **Service** `packages/api/src/services/provider-x-connect.ts`
  - `initiateXConnect` — builds the X authorize URL, persists hashed single-use pending-connect state, returns state + a `connectToken` carrying the raw PKCE verifier back to the client. The store only ever holds `sha256(state)` -> `{ sha256(verifier), scopes, redirectUri, tenant/workspace/user binding }`.
  - `completeXConnect` — loads state (no consume), binds it to tenant/workspace/caller/redirect_uri, verifies `sha256(verifier)` against the stored hash, exchanges the code, fetches identity, then consumes state EXACTLY once (only after upstream success), then stores tokens + creates/updates the `provider_accounts` row. Idempotent reconnect for the same X user id bumps the credential version + account revision, never duplicates.
  - `refreshXProviderCredential` — per-account single-flight refresh (`SELECT ... FOR UPDATE` on the account row inside the tenant-audited transaction). On `invalid_grant` (revoked) the account is degraded + audited, then fails closed. Vault rotation runs UNDER the row lock via `rotateSecretWithinTx` (no nested transaction).
  - `disconnectXProviderCredential` — best-effort revoke at X, degrade account, bump revision, audit. Local degrade proceeds even if the upstream revoke fails.
  - X network calls go through an injectable `__setXForwardForTests` seam. No real network in tests.
- **Routes** `packages/api/src/routes/provider-x-connect.ts` — `POST /v2/provider-accounts/connect/x/{initiate, callback, :id/refresh, :id/disconnect}`. Session-jwt + workspace admin/approver authority (`canConnectProviderAccounts`, mirrors PR3 `hasWorkspaceRoleAuthority`), 404 parity. redirect_uri gated through the existing tenant OAuth allowlist. Registered concretely before the `/v2` authority sub-app to avoid wildcard collision.
- **Vault** `packages/vault/src/secret-vault.ts` — `rotateSecretWithinTx` + `decryptSecretRow` so the refresh single-flight can rotate the credential under the account row lock without a nested transaction.
- **Store** `provider-authority-store.ts` — `canConnectProviderAccounts` (tenant-admin OR active workspace_admin/workspace_approver).
- **Env** `.env.example` — `X_CLIENT_ID` / `X_CLIENT_SECRET`, documented as SEPARATE from the `TWITTER_*` login credentials.

### Security notes
- **Rotating refresh, single-flight**: X refresh tokens are single-use rotating. The per-account row lock + a post-lock near-expiry re-check collapse concurrent refreshes to exactly one token-endpoint call; the loser observes the freshly rotated token and short-circuits. Mutation-proven.
- **Hashed, single-use, short-TTL state**: store holds only hashes of state + PKCE verifier; state is consumed exactly once and only after upstream success (a bad provider code cannot burn a live attempt).
- **Vault-versioned tokens**: credential stored as a versioned secret; `provider_accounts.credential_secret_id/credential_version` link the current version; reconnect + refresh bump the version.
- **Redirect allowlist**: `redirect_uri` (where X delivers the code) gated on initiate AND callback so an authorized-but-hostile member cannot redirect the code.
- **Cross-workspace IDOR guard** (codex P1): refresh verifies the account belongs to the caller's authorized workspace (404 otherwise).
- **No revoked-account resurrection** (codex P1): refresh fails closed unless the account is active; a disconnected account requires a fresh connect.
- **Fail-closed audit on revoke**: the degrade + audit commit, then the caller-facing failure is thrown (the failure signal does not roll back its own degrade).

### Tests + mutation receipts
23 tests (`packages/api/src/__tests__/provider-x-connect.test.ts`), all X network faked via the seam:
- authority gate (admin / approver / viewer-denied / non-member-denied)
- connect happy path + idempotent reconnect (version + revision bump, no dup)
- state: mismatch, expired, reused, PKCE verifier mismatch, scope/caller mismatch, token-exchange-does-not-consume-state (retryable)
- refresh: force rotate, **single-flight (exactly one token call)**, fresh-token short-circuit, **revoked -> degraded**, non-X rejected, **cross-workspace IDOR rejected**, **no revoked resurrection**
- disconnect: degrade + best-effort revoke + audit, missing account

**Mutation receipts** (weaken guard -> named test fails -> restore): single-flight post-lock re-check; PKCE verifier hash check; revoke->degraded status write; cross-workspace workspace guard; revoked-account status guard.

Biome clean, `packages/api` + `packages/vault` typecheck clean (`tsc --noEmit`). Adversarial review by codex (gpt-5.5) against origin/develop surfaced the two P1s above; both fixed + regression-tested in this PR.

### Honest gaps
- Route-level HTTP integration tests (full Hono stack) are not added; routes are thin delegators over the service (covered directly) + `assertAllowedOAuthRedirectUri` (already tested in `auth-oauth-redirect-allowlist.test.ts`). Service + guards are exhaustively unit-tested.
- No DB migration added (per instructions; 0082 owned by PR4). Uses the existing `provider_accounts` + `secrets` tables and the Redis-backed challenge store for pending state.
- `governed_v2` proxy wiring / secret_routes registration is workstream C (out of scope here); this branch works on the develop tip it branches from and should rebase when PR4 lands.

[sol-orch]

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -210,10 +210,21 @@ STEWARD_PROXY_PORT=8080
 
 # ─── Auth — OAuth (Twitter / X) ───────────────────────────────────────────────
 
-# Twitter/X OAuth (for "Continue with Twitter" login).
+# Twitter/X OAuth (for "Continue with Twitter" LOGIN — the user sign-in plane).
 # Create app at https://developer.x.com/en/portal/dashboard
 TWITTER_CLIENT_ID=
 TWITTER_CLIENT_SECRET=
+
+# ─── Provider-account X (Twitter) connect (#195 workstream A) ──────────────────
+# SEPARATE from the TWITTER_CLIENT_ID login credentials above. These authorize a
+# WORKSPACE to CONNECT an X account for governed AGENT ACTIONS (provider-
+# connection plane), NOT human sign-in. Keep the two credential sets distinct so
+# a login-app compromise cannot mint provider-action tokens (and vice versa).
+# The connect flow uses OAuth 2.0 authorization-code + PKCE (S256) and requests
+# the minimal scopes tweet.read tweet.write users.read offline.access.
+# offline.access is REQUIRED to receive the rotating refresh token.
+X_CLIENT_ID=
+X_CLIENT_SECRET=
 
 # ─── Web Dashboard (web/ — Next.js) ───────────────────────────────────────────
 # These variables are used by the Next.js web app, not the API server.
--- a/packages/api/src/__tests__/provider-x-connect.test.ts
+++ b/packages/api/src/__tests__/provider-x-connect.test.ts
@@ -1,0 +1,814 @@
+/**
+ * Provider-account X (Twitter) OAuth connect + token lifecycle tests
+ * (issue #195 workstream A).
+ *
+ * Covers: connect initiate/complete happy path, state mismatch, expired state,
+ * reused state, PKCE failure, wrong-role caller (via canConnectProviderAccounts
+ * gate), refresh rotation single-flight (concurrency: exactly one token call),
+ * refresh revoked -> degraded, disconnect. The X network is fully faked through
+ * the __setXForwardForTests seam; no real network is ever hit.
+ */
+
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  describe,
+  expect,
+  setDefaultTimeout,
+  test,
+} from "bun:test";
+import {
+  auditEvents as auditEventsTable,
+  closeDb,
+  getDb,
+  providerAccounts,
+  providerRoleBindings,
+  secrets,
+  tenants,
+  users,
+  userTenants,
+  workspaces,
+} from "@stwd/db";
+import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { SecretVault } from "@stwd/vault";
+import { and, eq } from "drizzle-orm";
+import { providerAuthorityStore } from "../services/provider-authority-store";
+import {
+  __setXForwardForTests,
+  completeXConnect,
+  disconnectXProviderCredential,
+  initiateXConnect,
+  type PendingConnectStore,
+  refreshXProviderCredential,
+  X_ADAPTER_KEY,
+  XConnectError,
+  type XCredentialPayload,
+  type XForwardRequest,
+  type XForwardResponse,
+  xCredentialSecretName,
+} from "../services/provider-x-connect";
+
+setDefaultTimeout(120_000);
+
+// ── Fixture identifiers ───────────────────────────────────────────────────────
+const TENANT = "tenant-x-connect";
+const ADMIN = "10000000-0000-4000-8000-0000000000a1";
+const APPROVER = "10000000-0000-4000-8000-0000000000a2";
+const VIEWER = "10000000-0000-4000-8000-0000000000a3";
+const OUTSIDER = "10000000-0000-4000-8000-0000000000a4";
+const WORKSPACE = "20000000-0000-4000-8000-0000000000b1";
+const ADMIN_BINDING = "30000000-0000-4000-8000-0000000000c1";
+const APPROVER_BINDING = "30000000-0000-4000-8000-0000000000c2";
+const VIEWER_BINDING = "30000000-0000-4000-8000-0000000000c3";
+
+const REDIRECT = "https://app.steward.test/connect/x/callback";
+const CONFIG = { clientId: "x-client-id", clientSecret: "x-client-secret" };
+
+const MASTER = process.env.STEWARD_MASTER_PASSWORD ?? "steward-api-test-suite-master-password";
+const vault = new SecretVault(MASTER);
+
+async function readAuditActions(tenantId: string): Promise<string[]> {
+  const db = getDb();
+  const rows = await db
+    .select({ action: auditEventsTable.action })
+    .from(auditEventsTable)
+    .where(eq(auditEventsTable.tenantId, tenantId));
+  return rows.map((r) => r.action);
+}
+
+/** base64url encode without the narrow bun-types Buffer encoding union. */
+function toB64url(value: string): string {
+  const bytes = new TextEncoder().encode(value);
+  const binary = Array.from(bytes, (b) => String.fromCharCode(b)).join("");
+  return btoa(binary).replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");
+}
+
+// ── In-memory pending-connect store (single-use, TTL-aware) ───────────────────
+class MemoryConnectStore implements PendingConnectStore {
+  private map = new Map<string, { value: string; expiresAt: number }>();
+  now = Date.now();
+
+  async set(key: string, value: string, ttlMs: number): Promise<void> {
+    this.map.set(key, { value, expiresAt: this.now + ttlMs });
+  }
+  async get(key: string): Promise<string | null> {
+    const rec = this.map.get(key);
+    if (!rec) return null;
+    if (rec.expiresAt <= this.now) {
+      this.map.delete(key);
+      return null;
+    }
+    return rec.value;
+  }
+  async consume(key: string): Promise<string | null> {
+    const value = await this.get(key);
+    if (value !== null) this.map.delete(key);
+    return value;
+  }
+}
+
+// ── Fake X network ────────────────────────────────────────────────────────────
+interface FakeXOptions {
+  exchangeToken?: Partial<{
+    access_token: string;
+    refresh_token: string;
+    scope: string;
+    expires_in: number;
+  }>;
+  exchangeStatus?: number;
+  identity?: { id: string; username: string; name: string };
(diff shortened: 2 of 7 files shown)


0xSolace:
## Adversarial review + merge gate — PR #197 (feat/x-oauth-connect) [sol-orch]

Independent second adversarial pass (fresh worktree off the PR branch, no reuse of the lane's worktree). Reviewed at `4ce95b4`, then update-branched onto develop (`1690079` / #196) → merged tip `b665fd8`. All 4 reviewed files are **byte-identical** between `4ce95b4` and `b665fd8` (the merge only pulled in #196's independent `packages/provider-x` package — zero file overlap with this PR). Verdict below stands at the exact merge tip.

### Scope fence — PASS
Diff touches only: `.env.example`, `packages/api/{app.ts, routes/provider-x-connect.ts, services/provider-x-connect.ts, services/provider-authority-store.ts, __tests__/provider-x-connect.test.ts}`, `packages/vault/src/secret-vault.ts`. No proxy / plugin-capabilities / execution-authorization changes, **no new migration**, no changes to the packages/auth twitter login flow. Vault delta limited to `rotateSecretWithinTx` + `decryptSecretRow`.

### Vault (rotateSecretWithinTx / decryptSecretRow) — PASS
- `rotateSecretWithinTx` is a faithful extraction of `rotateSecret`'s transaction body. **Encryption context/AAD preserved byte-for-byte**: `{ tenantId, name, version: newVersion }`. Tenant scoping enforced *inside* the function on every statement (current lookup, `secretRoutes` repoint, prior-version soft-delete all carry `eq(secrets.tenantId, tenantId)`) — not just at the caller. Arbitrary-tenant overwrite requires the caller to pass a foreign `tenantId`; callers derive it from the authorized workspace context, never user input.
- `decryptSecretRow` reconstructs the identical `{ tenantId, name: row.name, version: row.version }` decryption context; never reaches a route response (only feeds `loadCredential` internally). Relevant vault suites green (76 pass incl. master-password rotation which re-encrypts every secret through the same context paths).

### OAuth flow adversarial — PASS
- **State single-use is atomic** (not check-then-delete): load-without-consume, then `store.consume(key)` only *after* upstream token+identity succeed. `consume` is atomic in all three backends — Memory (single-tick get+delete), Redis (`GETDEL`, explicitly throws rather than racy-fallback), Postgres (`DELETE ... RETURNING`). Loser of a concurrent callback gets `X_STATE_REUSED`. A bad provider code does **not** burn the state (verified by the "token exchange failure does NOT consume the state" test).
- **PKCE binding**: only the SHA-256 of the verifier is persisted; callback recomputes `sha256Hex(parsedToken.verifier) === record.verifierHash` before spending. Swapped verifier ⇒ `X_PKCE_MISMATCH`.
- **State scope-fixation**: `record.{tenantId,workspaceId,initiatedByUserId}` must equal caller context; `record.redirectUri === input.redirectUri`. A state minted for another workspace/user cannot be replayed.
- **Redirect_uri allowlist**: reuses the shared, pre-hardened `assertAllowedOAuthRedirectUri` on **both** initiate and callback. Rejects `userinfo@evil` (username/password throw), scheme confusion (URL-normalized, https/loopback-only), subdomain/trailing-dot confusion (exact `origin` match), path traversal (exact `pathname`/`search` match).
- **Error paths never echo secrets**: `fail()` returns only the `XConnectError.code` enum; callback only reflects the attacker-supplied `body.error` OAuth param (not a secret). No token/state in any response body.

### Token custody — PASS
Tokens live ONLY in the versioned vault secret. Grepped every audit `append(...)` metadata + error path: they carry `xUserId / xUsername / scopesGranted / credentialSecretId / credentialVersion / requestId` — **never** accessToken/refreshToken. `decryptSecretRow` output stays internal.

### Refresh single-flight + custody — PASS (mutation-proven)
- `SELECT ... FOR UPDATE` on the account row inside the tenant-audited tx serializes callers; loser re-reads the freshly-rotated credential post-lock and short-circuits (`refreshed:false`, zero token calls). Re-ran the concurrency proof myself: `fake.counters.refresh === 1` for two simultaneous refreshes.
- **Transient upstream 500** ⇒ `refreshUpstream` throws `X_REFRESH_FAILED` (502) BEFORE any DB mutation; tx rolls back, **old refresh token preserved** (X did not rotate on a 500). Only `invalid_grant` degrades + commits.
- The network call is the last side-effecting step before the vault rotate, both inside the critical section; irreducible "rotate-then-crash" window is acknowledged in-code and is the standard OAuth-refresh limitation — non-blocking.

### The 2 codex P1 fixes — MUTATION-PROVEN
Reverted each guard, confirmed exactly the named test fails, restored byte-identical:
1. **Cross-workspace IDOR on refresh**: drop `account.workspaceId !== input.workspaceId` ⇒ `(fail) cross-workspace IDOR ... => X_ACCOUNT_NOT_FOUND`. ✅
2. **Revoked-account resurrection**: drop `account.status !== "active"` ⇒ `(fail) refresh does NOT resurrect a locally revoked account`. ✅
3. (bonus) **Single-flight**: drop post-lock near-expiry re-check ⇒ `(fail) SINGLE-FLIGHT ...` + `(fail) fresh-token short-circuit`. ✅
All restored to byte-identical pristine (verified `diff -q`).

### AuthZ — PASS
All 4 routes go through `requireConnectAuthority`: requires `authType === "session-jwt"` (agent tokens ⇒ 403, disjoint per #193) AND `canConnectProviderAccounts` (tenant-admin OR active workspace_admin/workspace_approver binding with temporal + environment validity — mirrors PR3's admin-OR-approver gate). Unauthorized ⇒ 404 `"resource not found"` (parity with provider-authority non-enumeration; existing provider-authority suite still green, 9 pass).

### Route registration / wildcard collision — PASS
`registerProviderXConnectRoutes(app)` registers concrete `/v2/provider-accounts/connect/x/*` handlers **before** `app.route("/v2", providerAuthorityRoutes)` (app.ts:255 before :256). Authority sub-app only defines `/provider-accounts/:id/{disable,operations}` — no `:id/refresh|disconnect` to collide, and concrete-first registration wins in Hono. Adjacent concrete-route suites (provider-actions-route, agent-route-scope) green.

### Idempotent reconnect — PASS
Same workspace + same X user id ⇒ optimistic-locked `UPDATE ... WHERE revision = account.revision` bumps version + revision, never duplicates (test: "reconnect ... never duplicates"). Different workspace + same X user id ⇒ separate account (secret name `provider-x/{workspaceId}/{xUserId}` is workspace-scoped). Concurrent double-connect ⇒ revision conflict ⇒ 409.

### Suites run (CI way, at merge tip b665fd8)
- Build/typecheck: `turbo run build --filter=@stwd/api...` → 17/17 (api `tsc --noEmit` passes).
- `provider-x-connect.test.ts` (isolated): **23 pass / 0 fail**.
- vault secret suites (secret-vault, lifecycle, rotate-master, route-validator): **76 pass**.
- provider-authority.test.ts: **9 pass**. shared x-provider-action (#196 merge intact): **13 pass**. provider-actions-route + agent-route-scope: **5 pass / 11 skip**.
- biome check (6 touched files): clean.

### Nits (non-blocking)
1. **persist ordering orphan**: in `persistConnectedAccount`, `rotateSecret`/`createSecret` runs *before* the audited tx; if the tx then hits a revision conflict (409), the new secret version is already committed while the account row still points at the old version. Self-heals on next reconnect (rotates again; old ciphertext soft-deleted). Consider moving the secret write inside the tx via `rotateSecretWithinTx` for full atomicity, matching the refresh path.
2. **connectToken is unsigned** base64url(JSON{state,verifier}). It's only ever handed to the initiating client and echoed back, and the store holds only hashes, so a store compromise can't forge it — but an unsigned bearer means whoever holds the connectToken holds the verifier. Acceptable given it never leaves the authenticated client round-trip; a MAC/seal would be belt-and-suspenders.

**Verdict: SHIP.** Squash-merging on green CI at `b665fd8`.

0xSolace:
✅ **MERGED** — squash `09aae24` on develop. CI fully green at exact merge tip `b665fd8` (13/13 checks pass incl. Integration Tests (Postgres) which runs the api isolated suite, E2E Compose, Lint+Typecheck, all unit suites). Update-branched onto #196 first; all 4 reviewed files byte-identical pre/post merge. Admin-merge path taken (I'm the PR author, cannot self-approve to clear REVIEW_REQUIRED). Full adversarial receipts + 3 mutation proofs in the review comment above. [sol-orch]

PR #199, opened 2026-07-16
feat(authority): execution authorization v2 and governed proxy cutover (authority plan PR4)

0xSolace:
## Summary

Authority plan **PR4**: makes the approved, immutable, `execution_ready` provider action (PR3) the ONLY way a governed route can decrypt/inject a provider credential and dispatch. Introduces execution-authorization **v2** (per-execution, single-use, signed commitment) and cuts governed proxy routes over to a claim-then-dispatch flow with a verifier-only decrypt boundary.

Built on PR2 (#191) + PR3 (#192, #193) and **merged up to develop `46748dd`** (X-provider PRs #196/#197/#198). Migration `0082` is journal max+1 (verified: #196-#198 added no migration).

### Spec-mapped changes
- **0082 migration + `schema.ts`**: extends `execution_authorization_nonces` with the v2 provider binding (version discriminator, full commitment tuple, dispatch state machine — no second ledger; v1 wallet rows keep `version=1`, null v2 fields). Adds `secret_routes.authority_mode` enum + `provider_operation_id` + governed CHECK, and EXTENDS the 0081 bump trigger predicate to bump `authority_revision` on authority_mode / provider_operation_id change (X5 cutover/rollback staling). Extends the PR3 `provider_action_bindings` status/state/approval-shape CHECKs + transition trigger for the PR4 execution lifecycle (`executing`/`succeeded`/`failed`/`outcome_unknown`).
- **v2 signing crypto in `@stwd/shared`** (`provider-execution-auth.ts`, `execution-contract.ts`, `provider-action.ts`): HKDF keyId rotation, domain-separated signature, JCS commitment hash, grant-dependency hash, canonical outbound-query serializer. Pure (no DB) so the separate proxy process can verify.
- **Mint-within-tx in api** (`provider-execution.ts` + PR3 `provider-approval.ts` hook): mints the single v2 authorization inside the PR3 resume tx (idempotent via `exec_auth_nonces_intent_uniq`; fail-closed if the exec-auth secret is absent, X7). The idempotent `execution_ready` resume path also mint-repairs a missing v2 nonce (codex P2).
[...]

PR #200, opened 2026-07-16
feat(evidence): add correlated provider-action case files

0xSolace:
## PR5 — Correlated Case Evidence & Existing Bundle Extension

Implements the PR5 contract (`PR5-CASE-EVIDENCE-SPEC.md`) on top of the landed PR1–PR4 stack (base `develop` tip `85ccc7c`, after PR4 #199). Makes the complete governed-provider case — action intent through terminal dispatch outcome — provable to an offline auditor using the EXISTING tamper-evident audit-chain + Ed25519 checkpoint cryptography. No second ledger, no second signing key, no new HMAC key, no migration.

### What landed (spec-mapped)

- **Shared types (§2.1, §4.4)** — `packages/shared/src/provider-case.ts`: crypto-free `ProviderCaseManifestV1` / `ProviderCaseEvidenceV1`, `ProviderCaseEventRole` / `ProviderCaseTerminalState`, the mechanical `requiredRoles()` completeness function, `roleForAction()` taxonomy mapper, and stable reason codes.
- **Correlator / manifest assembler (§1.3, §3, §4)** — `packages/api/src/services/provider-case.ts`: pure-read `getProviderCase` / `getProviderCaseEvidence`. Correlates by indexed `resource_id=caseId` + asserts `metadata.intentId` agreement (N23 event-theft guard). Resolves terminal state from the authoritative binding/nonce columns, computes mechanical completeness, hashes the provider idempotency key (§3.4 — raw key never enters the manifest), re-validates the safe summary redaction (§3.3), enforces per-event + whole-manifest size caps (§3.6). Runs all reads inside ONE read-only snapshot (REPEATABLE READ READ ONLY on Postgres; PGLite serialized) so `verifyAuditChain` + `readAuditBundleData` share the snapshot (§4.1/§4.3, KC06).
- **Signing factor-out (§6.2)** — `packages/api/src/services/audit.ts`: extracted `signAuditBundle` (+ `BUNDLE_CANONICALIZATION_SPEC`) so `/audit/bundle` and `/evidence` produce byte-identical envelopes from one signing path. Added an optional snapshot `executor` param to `verifyAuditChain` / `readAuditBundleData` (defaults to `getDb()`, existing callers unchanged).
- **Routes (§5)** — `packages/api/src/routes/provider-case.ts`: `GET /v2/provider-actions/:id/{case,evidence}` behind the SAME owner/admin + recent-MFA gate as `/audit/*` (extracted to `middleware/audit-gate.ts`; `/audit/*` now reuses it, never a weaker gate). Strict case-id validation, `/evidence` signing-key 503, uniform `CASE_NOT_FOUND` for foreign tenant/workspace/nonexistent (§5.4, D2 non-enumerating).
- **Offline verifier (§6.6, §7)** — `scripts/verify-evidence-bundle.mjs`: accepts BOTH a raw `/audit/bundle` (unchanged, no regression) and a `/evidence` envelope. Adds `--expected-key-fingerprint` / `STEWARD_EXPECTED_AUDIT_KEY_FP` trust-root binding (C6, E7; rotation set supported), and the manifest cross-check (§7.4): every manifest event matches the signed bundle event by hmac+action+role, every manifest fact is backed by the owning signed event's metadata, and the forged-completeness guard (a manifest claiming `complete` while a required role is absent from the SIGNED events FAILS). Zero project imports preserved.

### Anchor re-verification (CRITICAL FIRST STEP) — drift REPORTED, not silently reinterpreted

The spec was written pre-PR4/pre-X-provider. I re-verified every anchor against the landed code. Findings:

- **C-DRIFT-1 (taxonomy folding, ACCOMMODATED):** PR2 emits a SINGLE folded genesis event `provider.action.{allowed,denied,approval_required}` carrying BOTH the access and policy decision hashes, rather than the spec §1.4 separate `provider.access.decided` + `provider.policy.decided` events. `roleForAction` maps the folded event to the `genesis` role and `requiredRoles` treats `genesis` as satisfying the pre-approval decision requirement. This keeps completeness mechanical against code AS LANDED. C1/C2 are fully satisfied (every persisted intent gets exactly one correlated genesis event with `resource_type='provider_action'`, `resource_id=intents.id`, `metadata.intentId`, via the required-audit outbox + crash-recovery sweeper). Forward-compat names for the split taxonomy are retained in `roleForAction` in case PR2 ever splits them.
- **C1/C3 CONFIRMED landed:** every PR3 approval event (`provider.approval.{decided,expired,staled}`, `provider.resume.ready`) and every PR4 execution event (`provider.execution.{authorized,claimed,denied_at_boundary,dispatched,succeeded,failed,outcome_unknown}`) sets `resource_type='provider_action'` + `resource_id=intents.id` + `metadata.intentId`. PR4 dispatched/terminal + PR4 authorized events carry `providerIdempotencyKeyHash` (sha256), NEVER the raw key (C3). The manifest's fallback `0083` functional index is therefore NOT needed (default: zero migration, as the C1-ratified plan intended).
- **C-DRIFT-2 (route mount, DESIGN NOTE):** the spec §6.3 says "add the two routes under the existing `auditRoutes` Hono app," but that app is mounted at `/audit`, while §5.1 mandates the paths `/v2/provider-actions/:id/{case,evidence}`. Resolution: the gate middleware was EXTRACTED (`middleware/audit-gate.ts`) and the case routes registered on a dedicated `/v2` sub-app using the IDENTICAL gate (never weaker), registered concretely BEFORE the `/v2` authority wildcard (same collision-avoidance pattern as `registerProviderActionRoutes`). This honors the §5.1 path contract AND the §6.3 "no weaker gate" constraint.
- **C-DRIFT-3 (workspace-scoped session access, DEFERRED per spec):** the shared gate admits tenant `owner`/`admin` sessions (carrying a tenant role), and those callers may read ANY workspace in their tenant (the §5.2 tenant-admin arm). Workspace-scoped `workspace_admin` / `workspace_auditor` SESSION access is deferred — the spec §5.2 itself lists workspace-auditor as "if later added," and the session gate carries a tenant role, not a workspace role. Reported here rather than silently narrowing.

### Test counts (all green)

- `provider-case-service.test.ts` — 9 (terminal-state mapping, mechanical completeness, foreign-tenant/workspace/nonexistent → null (D2), no raw idempotency key leak §3.4, N23 event-theft drop).
- `provider-case-evidence.integration.test.ts` — 15 (route + offline round-trip: N01/N02/N03, N04, N05, N06, N07, N08, N09, N11, N12, N17/N35, N24, N37/N38/N39, clean PASS w/wo fingerprint, access-denied honest complete).
- `provider-case-verifier.test.ts` — 10 (synthetic no-DB: N18, N19, N20, N22, N40, N43, N44, N45, clean, plain-bundle no-regression).
- `provider-case-purity.test.ts` — 2 (static pure-read assertion, deterministic manifest).
- `provider-case-honesty.test.ts` — 6 (chain_segment_broken → unknown KC07; genesis-never-drained → never complete KC14; concurrent reads identical KC04/KC06; row-absence honesty; effect surfacing; unknown-action → `unclassified` never satisfies genesis).
- Plus KC15 over-cap → `400 CASE_RANGE_TOO_LARGE` (in the integration suite).
- **PR5 total: 43/43.**
- **Regression (all pass):** `/audit/bundle` 12 (signAuditBundle factor-out, incl. offline verifier round-trip — no behavior change); PR3 approval lifecycle/negative/boundary + provider-action-service 56; PR3 approval route/concurrency + provider-x-governed-e2e 34; PR4 governed-execution 35.

### Mutation receipts (§9) — `packages/api/scripts/pr5-mutation-proofs.sh`

10 guards, each weakens ONE evidence predicate and must flip a passing test to failing. **Result: 10 killed, 0 survived.**

- M1 manifest hmac==bundle (N43), M2 forged-completeness (N09), M3 fingerprint trust-root (N17), M4 fact-backing (N08), M5 unknown-schema fail-closed (N44), M6 tenant cross-check (N22), M7 events-content digest (N19), M8 seq-bracket (N20), M9 metadata.intentId agreement (N23), M10 workspace scoping (D2 enumeration).

### Invariants held

- **E1/E2:** case id IS `intents.id`; no second identifier, no second ledger. Manifest is a derived index over already-committed audit rows + binding/nonce columns.
- **E3:** authenticity is transitive through the existing Ed25519 checkpoint; no new signature/key.
- **E4/E8:** completeness is mechanical; `outcome_unknown`-unreconciled / awaiting-terminal / broken-chain are honestly `incomplete`/`unknown`, never `complete`; unknown versions fail closed.
- **E5:** data-minimized — hashes/IDs/enums/integer status only; raw provider idempotency key hashed; safe summary re-redacted; denylist scanned.
- **E6/E7:** offline, no network, no Steward secret; operator-holds-both-keys limit stated in verifier output + docs.

### Honest gaps

1. **Governed-execution terminal states (`executing`/`succeeded`/`failed`/`outcome_unknown` via the PROXY path) are not seeded in PR5 unit tests.** The `succeeded`/`failed`/`outcome_unknown` binding states and the v2 execution nonce are produced by PR4's proxy `dispatchGovernedExecution`, which PR5 does not exercise (it's PR4's tested territory + needs the fake transport arriving in PR6). PR5 tests cover: the ALLOWED-stub path (`allowed_stub`→`succeeded`/`failed`), access-denied, pending, approved/execution_ready, plus the manifest's execution-fact plumbing via the verifier synthetic bundles and the terminal-state resolver unit-level. A full governed→executed→evidence e2e lands naturally in PR6 (golden path) which consumes `/evidence` + the verifier. The N25 (raw provider idempotency key) guard is enforced at the SOURCE (PR4 emits only the hash; asserted by the C3-confirmed anchor + the manifest's own hashing), and the leak-scan asserts absence in exported evidence.
2. **REPEATABLE READ isolation is best-effort on non-PGLite:** `SET TRANSACTION ISOLATION LEVEL` inside the tx is wrapped in try/catch (some drivers only honor isolation at BEGIN). The append-only monotonic chain makes a re-read safe regardless, and the manifest's seq+hmac index pins the case events; the evidence bundle read+sign runs just after the snapshot (immutable range) to avoid a write-inside-read-only-tx deadlock on PGLite's single connection. A true real-Postgres REPEATABLE-READ coherence CI assertion (KC01/KC06 on real PG) is deferred to the PR7 real-Postgres harness (the spec routes KC-on-real-PG there).
3. **Verifier `roleForAction`/`requiredRoles` are replicated** (the verifier has ZERO project imports by design). They mirror `@stwd/shared/provider-case.ts`; a drift would be caught by the round-trip integration tests (the API-produced manifest is verified by the script), but there is no compile-time link. A golden-vector cross-check test asserts the API canonicalizer and the verifier agree byte-for-byte on the checkpoint/events digest (via the existing audit-bundle offline round-trip).

### Codex review

Ran `codex review --base origin/develop` iteratively. First pass surfaced 2 P2 findings, both fixed with tests:
1. `/evidence` could export an unbounded audit range for an over-cap case segment (DoS) → now throws `CaseRangeTooLargeError` before the range read; route returns `400 CASE_RANGE_TOO_LARGE`; `/case` still serves the honest `unknown` manifest.
2. Unknown/drifted correlated actions defaulted to the `genesis` role and could falsely satisfy completeness → added a non-satisfying `unclassified` role (shared + verifier mirror); unknown actions never satisfy a required role.

Second pass surfaced 2 more findings (P1 + P2), both fixed with tests:
1. [P1] The verifier derived its role map from EVERY event in the contiguous bundle segment; since the segment can include unrelated same-tenant events (§5.3/§7.5), another case's `exec_authorized` could back a fact or satisfy the forged-completeness guard. Now the verifier derives ALL role/fact reasoning ONLY from the signed events the manifest references (the case's own seqs).
2. [P2] `queue_row_absent_for_approval_path` was gated on `approvalQueueId == null` (the opposite of the corruption case); a case that went through approval whose queue row was deleted was not flagged. Now flags any case with a non-null `approvalQueueId` whose queue row fails to load.

Each fix has a dedicated regression test (§7.5 unrelated-event, deleted-queue-row). Mutation proofs re-confirmed 10/10 killed after each fix round.

DO NOT MERGE — opened for review.

— [sol-orch]


0xSolace:
[sol-orch] FINAL MERGE RECEIPT

**Verdict:** PASS, squash-merged.

- Final exact tip: `46f1462b29d2e9ffa64ea84d4e654e3d60f1385a`
- Merge SHA: `d2bf7ce3a43f258972f87bd2753e0a6526b25c52`
- Exact-tip CI: **13/13 green** on run `29491017719`
  - Lint + Typecheck + PR Metadata
  - Integration Tests (Postgres)
  - E2E Integration (Compose)
  - E2E Smoke (Embedded)
  - auth, db, policy-engine, proxy, redis, sdk, shared, vault, webhooks unit jobs

**Critical gate findings fixed before merge**

1. `629f805`: the case/evidence sub-app consumed `auditOwnerAdminMfaGate` without first running `tenantAuth` on `/v2/provider-actions/:id/{case,evidence}`. The real composed app therefore presented an unpopulated auth context and returned 403 to every caller, including a valid owner. The old mock-router tests injected context and masked the dead production routes. Both concrete paths now use the identical production pattern as approval/execute: `setNoStoreHeaders` + `tenantAuth`, then the shared owner/admin+recent-MFA gate. Real-app JWT tests prove valid owner reaches the handler (404 CASE_NOT_FOUND, not blanket 403) and unauth/member/stale-MFA/agent callers fail closed.
2. `acf73ee`: updated stale source-invariant tests after the audit gate extraction. They now verify both that `audit.ts` wires the shared gate and that the extracted gate still enforces session-jwt + owner/admin + recent MFA + no-store.
3. `dbca87b`: closed the route-wiring test's PGLite client in `afterAll`; removed an exit-99 open-handle failure despite 5/5 assertions passing.
4. `46f1462`: restored the `signAuditBundle` dependency boundary. Importing `API_VERSION` from heavyweight `services/context` made the vault rotation test transitively require unbuilt `@stwd/redis`. `API_VERSION` now lives in dependency-light `services/version.ts` and remains re-exported from context. Verified with `packages/redis/dist` physically absent: rotate-master-password **3/3, 72 assertions**.

**Security/evidence probes**

- Cross-tenant IDOR: valid foreign-tenant admin receives uniform `404 CASE_NOT_FOUND`.
- Completeness honesty: withheld/missing required events remain incomplete/unknown; access-denied genesis-only case is honestly complete.
- Tampered/reordered/removed chain events fail offline verification.
- Manifest omission cannot hide evidence: independent forged-completeness omission probe fails.
- Runtime secret/PII serialization: credential-looking comment absent; raw provider idempotency canary absent, only SHA-256 hash exported.
- Evidence cap is enforced server-side **before** `readAuditBundleData`; over-cap evidence returns `400 CASE_RANGE_TOO_LARGE` while `/case` remains available.
- No migration added; landed provider event anchors and hash-only idempotency claims re-verified.
- `signAuditBundle` remains the single shared signing path for `/audit/bundle` and `/evidence`.
- Byte-corruption sweep clean for all new/changed gate files; the only `third-party` bytes were intentional pre-existing prose.
- Codex supplementary review: two P2 + later P1/P2 findings fixed with dedicated regressions; final dependency-light version change reviewed clean.

**Counts**

- PR5 functional/security suite: **51 tests** (original 43 + 8 gate-hardening tests), all passing in isolated CI execution.
- Focused exact-tip local receipts: case/evidence routes + probes **24/24**, audit source invariants **20/20**, rotate-without-redis **3/3**.
- Independent probes: **2/2** (event-metadata byte flip; required-role manifest omission).
- Mutation proof on final tip: **10/10 killed, 0 survived/invalid**.

Merged via admin squash only after the exact tip and all required checks were green; REVIEW_REQUIRED was the sole remaining block and author==reviewer.

PR #225, opened 2026-07-16
feat(vault): require explicit production local-custody acknowledgement

### LINKED ISSUE #213: gap(security): secrets-at-rest custody-posture guide + production local-custody ack gate
## Gap / doc-and-guard task

Steward's honest posture (VISION.md, README) is that **local and KMS-envelope modes expose plaintext key material to the application at sign time** — only the PKCS#11 wrapping adapter and external-custody interface avoid that. This is the correct, non-inflated claim. But two things are missing operationally:

1. **No secrets-at-rest hardening guide** telling operators how to reach the stronger posture: when to use PKCS#11 vs third-party custody, what threat each removes, and what "plaintext in app memory at sign time" actually means for their threat model. The backends exist (`packages/vault/src/keystore-kms.ts`, `keystore-backend.ts:10` "attested execution, cloud HSM custody can ship their own backend", `third-party-key-custody.ts`) but the operator has no map from threat → backend.
2. **No guard against silently running the weakest custody mode in production.** There is a strong precedent for this exact pattern: the financial adapter registry *fails closed to a DISABLED adapter in production* unless mocks are explicitly opted in (`packages/adapters/src/registry.ts`). Custody has no analogous "you are running local-plaintext custody in production, set STEWARD_ACK_LOCAL_CUSTODY=true to proceed" gate.

## Why it matters
- **Security:** operators will default into local mode and never realize they're one memory-scrape from every key. A deliberate ack (like the adapter mock gate) turns a silent posture into an explicit decision.
- **Buyer:** enterprise security review asks "where does the key exist in plaintext, ever" — the guide is the answer sheet.

## Acceptance criteria
- [ ] `docs/security/custody-posture.md`: threat-model table mapping local / KMS-envelope / PKCS#11 / third-party to what each removes, with the "plaintext at sign time" boundary stated per mode.
- [ ] A production-mode acknowledgement gate for local-plaintext custody, mirroring the adapter-registry mock gate; unacknowledged ⇒ startup warning (or fail-closed under a strict flag).
- [ ] No claim inflation — wording matches VISION's earned boundaries.

## Size
M

Labels: gap, security, compliance

[sol-orch]

0xSolace:
## Summary

Completes issue #213: a secrets-at-rest **custody-posture guide** plus a **fail-closed production local-custody acknowledgement gate**, mirroring the adapter-registry mock gate. Base `develop` (`d2bf7ce`). Two commits: the gate (`760e7fd`, pre-existing on this branch) + the salvage/completion commit.

## Plaintext-boundary findings (traced to code, per mode)

Every built-in signing path calls `this.keyStore.decrypt(...)` and materializes a plaintext private-key **string** in-process before `privateKeyToAccount(...)` (`packages/vault/src/vault.ts`). So:

| Mode | Env selector | Plaintext at rest (Steward DB) | Plaintext at sign time (Steward process) | Verified in |
|------|-------------|-------------------------------|------------------------------------------|-------------|
| **local** | default (no `STEWARD_KMS_PROVIDER`) | effectively **yes** (AES-GCM ciphertext + unwrap material on same host) | **yes** | `keystore.ts` |
| **kms-envelope:aws** | `STEWARD_KMS_PROVIDER=aws` + key id/arn | **no** (data key needs live KMS `Decrypt`) | **yes** — `decrypt()` unwraps the data key then `createDecipheriv("aes-256-gcm", ...)` **in-process** (`keystore-kms.ts:172,117-122,188-194`) | `keystore-kms.ts` |
| **kms-envelope:pkcs11** | `STEWARD_KMS_PROVIDER=pkcs11` + module/pin/label | **no** (HSM `C_WrapKey`/`C_UnwrapKey`) | **yes** — same in-process AES decrypt after HSM unwrap | `keystore-kms.ts` |
| **third-party custody** | `VaultConfig.third-partyKeyCustodyProvider` (code-wired) | **no** — Steward holds an opaque `ExternalKeyHandleDescriptor`, `exportablePrivateKey: false` | **no** — signing delegated via `signTransaction()` returning a signed result; private key never enters Steward's address space | `third-party-key-custody.ts` |

The doc explicitly **corrects the "only PKCS#11 avoids plaintext" shorthand**: PKCS#11 is a key-wrapping adapter under the KMS-envelope backend and still AES-decrypts the private key in-process at sign time. Only third-party custody keeps plaintext out entirely.

**No claim inflation:** no MPC, no threshold/native-HSM signing, no operator-proof/tamper-evident claim. `KeystoreBackend` is documented as an operator-supplied seam, not a bundled capability.

## Gate semantics (fail-closed at the real boundary)

`assertProductionCustodyAcknowledged(mode)` in `vault-factory.ts`, called from `createConfiguredVault()` (the vault composition root):

- `NODE_ENV=production` **and** resolved mode `local` **and** `STEWARD_ACK_LOCAL_CUSTODY !== "true"` ⇒ **throws**.
- Fires at real boot: `packages/api/src/index.ts:212` calls `getConfiguredVault()` "before accepting traffic", so an unacknowledged prod-local deployment **refuses to start**, not warning-only, not lazy-on-first-request.
- Stronger modes (`kms-envelope:*`) boot without the ack (selecting them already requires explicit KMS/HSM config).
- Unknown/unsupported `STEWARD_KMS_PROVIDER` fails closed earlier in `configuredMode()`. `modeExposesPlaintextAtSignTime` treats an unknown mode as plaintext-exposing (fail-closed).
- **No secret leakage:** the error and the boot log line (`plaintext_at_sign_time=… local_custody_acknowledged=…`) never emit master password or KMS key ids (asserted in tests).
- **ACK, not ALLOW:** distinct verb — it acknowledges posture, it does not unblock a capability.

## Tests & mutations

- `vault-factory.test.ts`: **14 tests / 94 assertions**, all pass isolated (CI way: `run-tests-isolated.ts`). Covers fail-closed at the composition boundary (`createConfiguredVault`/`getConfiguredVault`), exact-`"true"` ack, malformed-ack rejection (`TRUE`/`1`/`" true"`/…), dev/test ergonomics unchanged, KMS-mode bypass, plaintext-at-sign-time semantics incl. unknown-mode fail-closed, secret non-leakage, and a "no `new Vault` outside the factory" guard.
- `scripts/custody-ack-gate-mutation-proofs.sh`: **4/4 mutations killed** (no-op gate, inverted prod check, loosened ack comparison, unknown-mode fails open) → each weakens the guard, suite goes RED, restore → GREEN. No `.bak` residue.

## Salvage notes (dead lane recovery)

Prior lane died mid-typecheck. Kept the committed gate + docs + mutation script (all correct). **Fixed:**
1. **Byte-corruption** ("third-" → "third-party" hex `74 68 69 72 64 2d`) in three string/comment references in `vault-factory.ts` introduced by `760e7fd` (FLEET-RULES rule 4). All 4 touched files now byte-clean.
2. **codex P3 (reproduced then fixed):** the gate tests used `STEWARD_KDF_SALT ||= …`, so an invalid/short **ambient** salt (reproduced with `STEWARD_KDF_SALT=short`) made 4 tests fail inside `Vault` salt validation instead of exercising the gate. Now pin a known-valid salt (`beforeEach` + explicit per-test) and add `STEWARD_KDF_SALT` to the saved/restored env keys. Verified passing under both invalid-ambient and clean env.

## Verification run
- `tsc --noEmit` (packages/api): clean
- `biome check` (touched files): clean
- `@stwd/shared` + `@stwd/redis` dist built first (CI way)
- Regression spot-check: `vault-approval-gates`, `vault-signing-gates`, `vault-aggregation-enforcement` pass (one transient PGLite migration flake only when 4 files run concurrently; each passes alone — unrelated to this change).

## Honest gaps / deferred
- **CLI wiring deferred (PR6-owned).** `packages/cli` doctor/init and compose files are owned by stwd-pr6-impl per FLEET-RULES scope fences; **not touched**. Follow-up: surface `STEWARD_ACK_LOCAL_CUSTODY` in `cli doctor` (warn when prod-local + unacked) and in `init` scaffolding once PR6 merges.
- The gate is a **posture acknowledgement, not a mitigation** — setting the ack does not make local mode safer; it records accepted residual risk (stated in the doc's "Honest limitations").
- codex review: first pass returned "implementation sound" + the single P3 above (fixed). A second confirmation pass ran the full suite (14/0) and build clean, then crashed on an internal codex stdin tooling error before emitting a final structured verdict; no new findings were surfaced.

Closes #213

[sol-orch]

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,18 @@ STEWARD_BIND_HOST=127.0.0.1
 # Rotating this requires re-encrypting all keys.
 STEWARD_MASTER_PASSWORD=
 
+# ─── Security — Custody posture acknowledgement ──────────────────────────────
+
+# [REQUIRED in production when running local custody] When no STEWARD_KMS_PROVIDER
+# is set, the vault runs in `local` mode: keys are AES-256-GCM at rest but are
+# decrypted to plaintext in the API process memory at sign time. In production
+# the vault REFUSES TO BOOT in local mode unless you explicitly acknowledge this
+# posture by setting this to exactly "true" (a deliberate, auditable decision;
+# mirrors the STEWARD_ALLOW_MOCK_ADAPTERS gate). To avoid needing the ack, run a
+# stronger backend instead: STEWARD_KMS_PROVIDER=aws|pkcs11 (+ its config).
+# See docs/security/custody-posture.md for the full threat model.
+# STEWARD_ACK_LOCAL_CUSTODY=true
+
 # [REQUIRED in production] Canonical JWT signing/verification secret used by
 # the API, proxy, auth routes, user sessions, and agent-scoped tokens. Must be
 # at least 32 characters in production. Existing deployments using
--- a/deploy/enterprise-reference/.env.example
+++ b/deploy/enterprise-reference/.env.example
@@ -30,6 +30,16 @@ STEWARD_SESSION_SECRET=
 STEWARD_KDF_SALT=
 STEWARD_AUDIT_HMAC_KEY=
 
+# Custody posture acknowledgement. This reference profile sets no
+# STEWARD_KMS_PROVIDER, so the vault runs in `local` mode: keys are AES-256-GCM
+# at rest but materialize as plaintext in the API process memory at sign time.
+# In production the vault REFUSES TO BOOT in local mode unless this is set to
+# `true` (a deliberate, auditable acknowledgement of the posture). The compose
+# file defaults this to `true` for the reference stack; to run a stronger
+# posture instead, set STEWARD_KMS_PROVIDER=aws|pkcs11 (+ its config) and set
+# this to an empty value. See docs/security/custody-posture.md.
+STEWARD_ACK_LOCAL_CUSTODY=true
+
 # STEWARD_EXECUTION_AUTH_SECRET (PR4): key for provider execution
 # authorization v2. SEPARATE from STEWARD_JWT_SECRET (never a fallback).
 # Format: a comma-separated `keyId:secret` rotation list; the FIRST key
--- a/deploy/enterprise-reference/docker-compose.yml
+++ b/deploy/enterprise-reference/docker-compose.yml
@@ -44,6 +44,16 @@ services:
       STEWARD_EXECUTION_AUTH_SECRET: "${STEWARD_EXECUTION_AUTH_SECRET:?required}"
       STEWARD_SESSION_SECRET: "${STEWARD_SESSION_SECRET:?required}"
       STEWARD_KDF_SALT: "${STEWARD_KDF_SALT:?required}"
+      # Custody posture: no STEWARD_KMS_PROVIDER is set, so the vault runs in
+      # `local` mode (plaintext key in app memory at sign time). With
+      # NODE_ENV=production the vault refuses to boot local mode unless the
+      # posture is acknowledged. The value comes from .env (this profile requires
+      # an .env; the template ships STEWARD_ACK_LOCAL_CUSTODY=true as the
+      # reference stack's deliberate, auditable local-custody posture). No
+      # `:-true` default is injected here so an env that omits it fails closed
+      # rather than silently acking. To harden, set STEWARD_KMS_PROVIDER=aws|pkcs11
+      # (+ config) and clear the ack. See docs/security/custody-posture.md.
+      STEWARD_ACK_LOCAL_CUSTODY: "${STEWARD_ACK_LOCAL_CUSTODY:-}"
       STEWARD_AUDIT_HMAC_KEY: "${STEWARD_AUDIT_HMAC_KEY:?required}"
       STEWARD_AUDIT_SIGNING_KEY: "${STEWARD_AUDIT_SIGNING_KEY:?required}"
       STEWARD_PLATFORM_KEYS: "${STEWARD_PLATFORM_KEYS:-}"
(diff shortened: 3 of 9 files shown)


0xSolace:
## Gate verdict: SHIP-WITH-NITS → MERGED

Adversarial gate by [sol-orch] gate-225 lane. 8 probes run independently.

**Findings (all fixed on-branch before merge):**
- **[P1] Reference deploy crash-loop (codex):** the gate fail-closes prod local custody without `STEWARD_ACK_LOCAL_CUSTODY=true`, but the shipped compose stacks (root `docker-compose.yml` + `enterprise-reference`) run `NODE_ENV=production` with no KMS provider → resolve to local → would crash on boot after this lands. Reproduced against both stacks.
- **[P2] Silent auto-ack would defeat the gate (codex):** my first fix defaulted `${STEWARD_ACK_LOCAL_CUSTODY:-true}`, which auto-acks a production local deploy and defeats the explicit-decision intent.
- **Resolution:** `steward init` now emits the ack explicitly (it already commits to prod+local, so it records the posture auditably); compose files pass the ack through with NO `:-true` default (unacked prod fails closed); `.env` templates document the posture + KMS-hardening alternative. E2E CI runs `NODE_ENV=test` so the gate is correctly inert there.

**Clean probes:** no construction-path bypass (all vaults route through the memoized factory + boot assertion at index.ts:212); env semantics match the adapter-registry mock-gate precedent (`NODE_ENV==="production"`, exact-match `"true"`); `modeExposesPlaintextAtSignTime` truth table verified against actual keystore-kms.ts (KMS-envelope DOES expose plaintext in-process — honestly documented); unknown-mode fails closed; all 4 committed files byte-clean (no `third-`/external corruption in the tree); docs claims non-inflated (explicitly disclaims MPC/HSM-signing/operator-proof and corrects the PKCS#11 shorthand).

**Verification (merged tip, against develop 470ac7a):** api tsc clean · 14/14 gate tests · 4/4 mutations killed · compose-plugin-toggle + deploy-artifacts + cli init/doctor green · byte-scan clean · YAML valid · iterative codex converged with zero findings · **13/13 required CI green**.

Squash-merged `7066eb1`, closes #213. REVIEW_REQUIRED-only (author==reviewer) → API/admin fallback per fleet rules. [sol-orch]

PR #245, opened 2026-07-30
feat(vault): no-read-back as a property of the SecretVault custody plane (A2, converged)

0xSolace:
## Pillar A / lane A2 — no-read-back custody (CONVERGED onto the existing plane)

**History:** this PR originally shipped a sibling age-file SecretStore. The eliza
vault audit (`ELIZA-VAULT-AUDIT-2026-07-30`) correctly flagged that a sops/age
layer beside `SecretVault` + custody modes risks becoming **a second, unaudited
secret plane**. Decision (delegated, autonomous mode 2026-07-30): keep #245's
core security property, drop the parallel store. The branch now makes
no-read-back a **property of the existing SecretVault/custody plane**. OFF HOLD.

### The security property: no read-back
Once stored, **no API path returns a secret's plaintext**. Secrets are
write + exercise only:

- **`SecretVault.exerciseSecret(tenantId, secretId, use, { beforeUse })`** —
  the canonical use-only path. Decrypts into a local, hands it to a caller
  closure, returns the closure's *RESULT* (an HTTP response, a signature),
  never the secret. `beforeUse` is a **fail-closed audit chokepoint**: it runs
  after row lookup but *before* decryption, so a failed audit append means the
  secret is never decrypted and the closure never runs.
- **Closed caller set, CI-pinned** (`packages/vault/src/__tests__/secret-no-read-back-inventory.test.ts`):
  the direct-return decrypt methods (`decryptSecret`/`decryptSecretRow`) are
  reachable only from proxy credential injection and provider-x refresh — the
  two pre-existing, classified consumers. A new caller anywhere in
  `packages/` fails CI with a classification instruction pointing at
  `exerciseSecret`. The API routes layer is proven decrypt-free.
- **HTTP surface proof** (`packages/api/src/__tests__/secrets-no-read-back.test.ts`):
  `/secrets` routes never import a decrypt method, request `value` fields are
  input-only (never echoed into any response), the GET surface is pinned to
  `/`, `/routes`, `/:id` (metadata only), and no reveal/export/value route can
  appear without failing CI.

### Why converging is strictly better than the age store
- **One custody plane, one audit surface.** Every op flows through the
  existing fail-closed `writeAuditEvent` + MFA-gated routes — the age store
  had its own unaudited files.
- **Custody strength inherited, not duplicated.** The plane already has
  `local` (prod boot refuses without `STEWARD_ACK_LOCAL_CUSTODY`) and
  `kms-envelope:aws|pkcs11`; Pillar B swaps the master-key source for an
  attestation-gated release with **zero API change**. The age backend's
  "TEE slot-in later" promise is already true here.
- **Per-tenant scoping, versioning, rotation, expiry, route repointing** —
  all already built and tested on this plane.

### CLI salvage (zero-plaintext-transit onboarding)
`steward secret add|rotate` now read the value from **stdin or `--file`** by
default; `--value` still works for backward compat but warns loudly about
shell-history leakage.

### Tests (bidirectional, all green locally)
- Exercise works: sync + async closures, result-only return, no secret in
  serialized results.
- Get is impossible: metadata surfaces proven value-free, `SecretVault`
  method surface pinned (a new plaintext-capable method fails the pin),
  metadata field set pinned.
- Fail-closed: throwing `beforeUse` blocks decryption; missing secret skips
  `beforeUse`; expired secrets refuse exercise.
- Mutation proofs (run locally, reverted): a hypothetical
  `GET /secrets/:id/value` read-back route fails 2 tests; an unauthorized
  `decryptSecret` caller in another package fails the inventory scan.
- vault: 363 pass / 0 fail. cli: 31 pass. api+shared secret suites: 20 pass.
  Typecheck clean (vault, cli, api). Biome clean.

**Removed:** `SecretStoreBackend` / `AgeFileSecretStore` / `steward secrets`
CLI / `migrate-env-secrets` / `age-encryption` dep. Migration off env vars
onboards through the existing `/secrets` + capability grants instead
(A1/#247 + kms-routes/#250 are the joins).

**Secrets:** dummy values only, everywhere.


0xSolace:
converged, ready for chain — all 13 checks green.

Restructured per the vault-audit ruling: age sibling store removed, no-read-back is now a property of the existing SecretVault custody plane (`exerciseSecret` use-only path + fail-closed `beforeUse` audit chokepoint + CI-pinned decrypt-caller inventory + /secrets HTTP surface proof). Bidirectional mutation proofs in the receipt. Merge is the chain lane's call.

PR #250, opened 2026-07-30
feat(api): /v1/kms/* routes implementing the Eliza StewardKmsAdapter contract

0xSolace:
## What

Implements the `/v1/kms/*` endpoint surface that the Eliza production KMS adapter (`elizaos/eliza` `packages/security/src/kms/steward-adapter.ts`) already specifies as its wire contract. The spec is FROZEN on the client side; this PR implements Steward TO that spec (closing gap #1 from the 2026-07-30 vault audit).

### Routes
- `POST /v1/kms/keys` — get-or-create `{ keyId, version }`
- `POST /v1/kms/keys/:keyId/rotate` — `{ keyId, newVersion }`
- `GET  /v1/kms/keys/:keyId/versions` — `{ versions: number[] }`
- `POST /v1/kms/keys/:keyId/encrypt|decrypt` — AES-256-GCM with optional AAD, versioned
- `POST /v1/kms/keys/:keyId/hmac` + `/hmac/verify` — HMAC-SHA256, any-version verify
- `POST /v1/kms/keys/:keyId/sign` + `/verify` — Ed25519 (rsa-pss-sha256 fails closed with 400)
- `GET  /v1/kms/keys/:keyId/public` — raw 32-byte Ed25519 public key

## Design (extend, don't fork)

- **Storage:** key roots are rows in the **existing `secrets` table**, sealed by the **existing `SecretVault`** (AES-256-GCM under `STEWARD_MASTER_PASSWORD`, domain-separated `secret-vault` root). One row per version via SecretVault's native rotation. Historical (soft-deleted) versions stay decryptable so rotation never bricks old ciphertext — standard KMS decrypt-old/encrypt-new semantics.
- **Derivation:** working keys (AEAD / HMAC / Ed25519 seed) are HKDF-SHA256-derived per use-domain from the version root, mirroring Eliza's `LocalKmsAdapter` model, so the root is never used directly.
- **Auth:** agent-token bearer only, over the existing `tenantAuth` plane. Session JWTs, tenant API keys, app secrets, and proxy-scope-only tokens are all DENIED (403). No agent principal → no KMS.
- **Per-agent scoping:** the wire `keyId` is namespaced internally as `kms/<agentId>/<keyId>` using the **verified token's** agent id (never the request). Agent A cannot address, decrypt with, or even observe agent B's material; the same `keyId` across agents yields independent key material.
- **Fail-closed everywhere:** strict canonical base64 (round-trip checked), strict keyId charset, indistinguishable decrypt failures (no AAD/padding oracle), 404 for missing keys (the status the adapter classifies as key-unavailable).
- **Audit:** keys-never-values events (`kms.key.created`, `kms.key.rotated`, `kms.encrypt`, `kms.decrypt`, `kms.hmac`, `kms.sign`) through the existing `writeAuditEvent` pipeline.

## Tests (contract conformance)

`packages/api/src/__tests__/kms-routes.test.ts` drives the real composed app with the **exact request shapes and field names the adapter emits**, and asserts the exact top-level response fields it requires:

- success round-trips: encrypt→decrypt with AAD, hmac→verify, sign→verify **plus independent `node:crypto` Ed25519 verification against `GET /public`** (bidirectional cryptographic evidence, not just self-consistency)
- rotation: version bump, versions listing, decrypt-old-version after rotation, new encrypts on new version
- wrong-agent DENY (404 on every op, cross-agent decrypt fails, public keys differ)
- expired-token DENY, non-agent-principal DENY, proxy-scope-only DENY, missing-bearer DENY
- malformed DENY: bad base64, bad keyId (`../escape`), bad algorithm, bad version, missing fields
- tamper DENY: flipped ciphertext byte and wrong AAD produce identical 400s (no oracle)
- audit rows exist and contain no key material

`15 pass, 0 fail, 77 expect() calls` locally (PGLite). Also green: secrets suite, tenant-auth, agent-jwt-hardening, idempotency; `turbo build --filter=@stwd/api` and biome clean.

`runtime.d.ts`: the ambient Buffer/crypto shims (used because the api package type-checks under workers-types) gained base64 encodings + `alloc`/`concat` — additive only.

## Out of scope
- No prod deploy (branch + PR only).
- OIDC/RS256 bearer variant of the adapter auth (the eliza-cloud JWKS plane, `middleware/agent-jwt.ts`) — routes currently accept Steward's own HS256 agent tokens; wiring the RS256 authenticator onto `/v1/kms/*` is a follow-up decision for the cloud integration lane.
- `rsa-pss-sha256` signing (fails closed, loudly).


0xSolace:
Review pass with one custody-plane fix applied.

What I checked:
- mounted under existing tenantAuth only, no parallel auth plane
- route gate requires verified agent-token principal plus standard agent scope; session JWT, tenant API key, app-secret, proxy-scope-only denied
- keyId is internally namespaced by verified agent id, never request body
- roots live in existing secrets table via SecretVault, no plaintext responses or audit metadata
- contract tests cover Eliza adapter request/response fields, crypto round-trips, wrong-agent deny, malformed/tamper denies, audit no-material

Fix applied before merge consideration: KMS no longer calls SecretVault.decryptSecretRow directly from an API route. Added SecretVault.exerciseSecretRow use-only helper for already-read rows so KMS can decrypt historical rotation rows without expanding the direct-return decrypt caller set. Local: typecheck green; KMS + no-read-back tests 25 pass; targeted biome green.

PR #253, opened 2026-07-30
feat(policy): per-grant policy engine on the capability invoke path (C1)

0xSolace:
## What

Pillar C / lane C1 of the Sovereign Custody plan: a real **policy engine on the capability invoke path**. Today a grant is binary (agent may use capability until `expiresAt`). This PR adds per-grant, invoke-time constraints — which agent, how much, which venue, what approval threshold — enforced server-side, layered UNDER the existing tenant `capability-intent` policy set.

## Design

**Two-layer composition** on `POST /capabilities/:name/invoke` (and the OpenAI-compat adapter, which shares the same core):

```
grant resolution (fail-closed, existing)
  -> GRANT POLICY (new, per-grant jsonb)      deny -> 403
  -> capability-intent composition (existing) deny -> 403
  -> approval from either layer (no deny)     -> 202 (existing pending flow)
  -> allow + allow                            -> proxy delegation (existing)
```

A deny from either layer wins; the grant layer can only **narrow** the tenant layer, never widen it. The agent SDK surface is unchanged: allow / deny / 202.

**Rule vocabulary (v1)** — `@stwd/policy-engine` `grant-policy.ts`, pure/deterministic/non-throwing:
- `rate`: N invokes per trailing window (counted from the append-only `capability_invocations` rows; denied probes consume rate)
- `amount` (value-bearing): declared integer-micros arg field, per-invoke hard cap, rolling-window cumulative cap, and `approvalOverMicros` (over X → 202 human approval)
- `venue`: allowlists of hosts (`*.` wildcard, mirrors proxy matching), methods, segment-boundary path prefixes
- `time`: `notBefore`/`notAfter` + UTC minute-of-day allowed window (wraps across midnight)
- `approval.always`: every invoke → human approval

**Capability classes, fail-closed:**
- `value-bearing` **must** carry amount bounds — a value-bearing policy without them is a config error (deny)
- `plain-secret` with no constraints allows iff the grant is valid — this is the **explicit permissive default** existing grants are migrated to

**Fail-closed everywhere:** unknown keys, malformed/over-retention windows, missing trailing-window counters, unreadable amount args, hostile jsonb values — all deny. A deny is never softened into an approval (deny-first precedence, same discipline as `composeCapabilityIntentDecision`).

**Audit:** every terminal outcome records `verdict_rule` + `verdict_reason` (+ `amount_micros`) on its invocation row — you can always answer *which rule fired*. The rolling amount window sums `allow`+`approval`+`error` rows: pending approvals and post-authorization infra errors **reserve** spend, so parallel invokes cannot overshoot a money cap (over-counting acceptable; under-counting never).

## Migration / rollout (nothing breaks)

- `capability_grants.policy` jsonb: migration 0002 backfills existing rows and defaults new rows to `{"version":1,"class":"plain-secret"}` — exactly pre-policy behavior, but written down instead of implied by NULL
- **Flag-gated strict mode**: `STEWARD_GRANT_POLICY_STRICT=true` makes a NULL policy (rows written outside the migrated path) deny; compatibility mode allows with an audited `no-policy.permissive` verdict
- Grant create accepts an optional `policy` (validated by the **same** fail-closed parser the invoke path enforces with — single source of truth, no drift); new `PUT /capabilities/grants/:grantId/policy` (tenant-admin + recent MFA, the same bar as every other grant mutation) edits it, effective next invoke

## Tests (bidirectional)

- **70 engine tests**: every rule blocks what it should AND allows what it should; malformed-config parse matrix (25+ rejection cases incl. hostile Proxy values); precedence (hard deny beats approval); determinism
- **18 enforcement tests** at the mounted invoke route over real PGLite + plugin migrations: rate arc (N pass, N+1 blocked, rule audited), amount per-invoke cap / approval threshold 202 with reserved spend / rolling window breach incl. pending-approval reservation, venue block+pass, time bounds block+pass, tenant-deny-beats-grant-allow and tenant-deny-beats-grant-approval layering, malformed + stripped policy rows deny, strict-mode flag both ways, verdict audit on every row
- Full plugin suite: 161 pass / 5 fail — the 5 failures (`invoke-e2e` proxy-forward AES decrypt error) **pre-exist on the base commit with a clean tree** (verified via `git stash`: same 5 fail without this diff; an env-dependent keystore issue unrelated to this change)

## Scope notes

- Enforcement is entirely server-side in `plugin-capabilities` invoke (the single shared core used by both the envelope route and the OpenAI adapter). The `packages/proxy` redis rate/spend enforcement is untouched — it remains the per-host transport-layer backstop; this PR is the authorization layer above it.
- Approval quorum >1 and provider-verified amount extraction are follow-ups; the 202 path reuses the existing approval state machine.

Part of SOVEREIGN-CUSTODY-PILLARS-2026-07-30.md (Pillar C, lane C1).

### DIFF
--- a/packages/plugin-capabilities/drizzle/0002_grant_policy.sql
+++ b/packages/plugin-capabilities/drizzle/0002_grant_policy.sql
@@ -1,0 +1,29 @@
+-- @stwd/plugin-capabilities migration 0002: per-grant policy + policy-verdict audit (C1).
+--
+-- 1. capability_grants.policy — the per-grant policy document (jsonb) evaluated
+--    at invoke time by @stwd/policy-engine's grant-policy module. EXISTING grants
+--    are backfilled with the EXPLICIT permissive default
+--    ({"version":1,"class":"plain-secret"}) so rollout changes nothing for them:
+--    a plain-secret policy with no constraints allows iff the grant is valid,
+--    which is exactly the pre-policy behavior — but now it is written down
+--    rather than implied. The column stays NULLABLE (with a DEFAULT for new
+--    rows) so strict mode (STEWARD_GRANT_POLICY_STRICT=true) has a real
+--    fail-closed target: a NULL policy denies under strict mode and allows
+--    (explicitly, audited) under compatibility mode.
+--
+-- 2. capability_invocations gains the policy VERDICT columns: which rule fired
+--    (verdict_rule), the human-readable reason (verdict_reason), and the
+--    extracted per-invoke amount in integer micros (amount_micros). amount_micros
+--    on decision IN ('allow','approval','error') rows is the source for the
+--    rolling-window cumulative amount cap (pending approvals and post-
+--    authorization infra errors RESERVE spend so the window can never be
+--    under-counted).
+ALTER TABLE "capability_grants" ADD COLUMN "policy" jsonb DEFAULT '{"version":1,"class":"plain-secret"}'::jsonb;
+--> statement-breakpoint
+UPDATE "capability_grants" SET "policy" = '{"version":1,"class":"plain-secret"}'::jsonb WHERE "policy" IS NULL;
+--> statement-breakpoint
+ALTER TABLE "capability_invocations" ADD COLUMN "verdict_rule" text;
+--> statement-breakpoint
+ALTER TABLE "capability_invocations" ADD COLUMN "verdict_reason" text;
+--> statement-breakpoint
+ALTER TABLE "capability_invocations" ADD COLUMN "amount_micros" bigint;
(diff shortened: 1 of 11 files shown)


lalalune:
Closing as unsafe at the current authorization boundary. Exact review found caller-controlled args drive amount policy while a separate body is forwarded, so an agent can declare amountMicros=1 and execute a larger operation. Rate/amount checks are read-then-record and record only after upstream execution, allowing concurrent cap overshoot; audit failures are swallowed after side effects. The approval path only inserts an invocation row with no approve/deny/resume consumer, so 202 is a dead end. Capability class is policy self-declared, and strict-mode typos silently become permissive. The parser and deny precedence are strong, but the tests are sequential and never prove forwarded-body binding or atomic reservation. A replacement needs a server-owned canonical extractor from exact forwarded bytes, atomic durable pre-dispatch reservation/finalization, canonical exact-request approval integration, server-owned capability class, concurrency/multi-instance tests, and fresh exact-head CI.

PR #254, opened 2026-07-30
feat(sealed-state): attestation-bound agent state at rest (C2)

0xSolace:
### Placement audit
This belongs in new `@stwd/sealed-state`, not `@stwd/vault`. Vault is tenant credential/signing custody with database and exercise policy semantics. Agent memory/runtime snapshots are opaque blobs released by runtime measurement. The package consumes measurement types from `@stwd/attestation` without conflating the custody planes.

### What ships
- AES-256-GCM envelope encryption: random DEK per blob, KEK-wrapped DEK, authenticated provider/purpose/image/config metadata.
- `dstack-tdx` backend: `/Info` measurement equality is mandatory, then `/GetKey` derives a measurement/purpose-separated KEK from dstack's application key. No fallback.
- loudly named `INSECURE-noop-dev`, forbidden in production.
- 64-case measurement mismatch property proof; payload/header corruption rejection.
- `docs/SEALED-STATE.md` threat model and explicit non-goals.

### Honest production consumer
Exact live path: `packages/api/src/embedded.ts`, executed by root `bun run start:local` via `scripts/start-local.ts`. With `STEWARD_SEALED_PGLITE_PATH`, PGLite runs only in memory, restores from an authenticated envelope, and seals an atomic 0600 snapshot before database shutdown. First boot is explicitly forced in-memory. Shutdown persistence errors exit non-zero. Existing plaintext directories are not silently migrated or deleted.

This is opt-in because graceful-shutdown snapshots are not yet an HA database profile. It is a real production entry path behind a flag, not a harness.

### Proof
- `bun test packages/sealed-state/src/__tests__/sealed-state.test.ts`: same M passes; 64 distinct M fail; corruption fails; dev backend production enable fails.
- `bun test src/__tests__/sealed-pglite-snapshot.test.ts` from `packages/api`: real PGLite dump -> seal -> restore -> query passes; envelope excludes plaintext.
- shutdown hook success/failure tests pass.
- `bun run typecheck`: 34/34 packages green.
- `bun run lint`: clean for this diff; repo reports only pre-existing warnings.

### Non-goals
Same-measurement authorized clones/hosts, use-time runtime compromise, access-pattern hiding, crash-consistent continuous journaling, automatic legacy plaintext deletion.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ COPY packages/adapters/package.json          packages/adapters/package.json
 COPY packages/agent-trader/package.json      packages/agent-trader/package.json
 COPY packages/api/package.json               packages/api/package.json
 COPY packages/attestation/package.json       packages/attestation/package.json
+COPY packages/sealed-state/package.json       packages/sealed-state/package.json
 COPY packages/auth/package.json              packages/auth/package.json
 COPY packages/db/package.json                packages/db/package.json
 COPY packages/eliza-plugin/package.json      packages/eliza-plugin/package.json
@@ -84,6 +85,7 @@ COPY packages/adapters/package.json          packages/adapters/package.json
 COPY packages/agent-trader/package.json      packages/agent-trader/package.json
 COPY packages/api/package.json               packages/api/package.json
 COPY packages/attestation/package.json       packages/attestation/package.json
+COPY packages/sealed-state/package.json       packages/sealed-state/package.json
 COPY packages/auth/package.json              packages/auth/package.json
 COPY packages/db/package.json                packages/db/package.json
 COPY packages/eliza-plugin/package.json      packages/eliza-plugin/package.json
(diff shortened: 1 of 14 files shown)


lalalune:
Closing as cryptographically unsafe at exact head. The decrypt path accepts truncated AES-GCM tags; review reproduced successful plaintext recovery with 4-, 8-, 12-, 13-, 14-, and 15-byte prefixes of the 16-byte tag, downgrading authentication to as little as 32 bits for both payload and wrapped-DEK tags. The envelope needs strict canonical base64/schema validation, 12-byte IVs, 16-byte tags with authTagLength:16, 32-byte wrapped DEK, size bounds, and malformed/truncation tests. Additional blockers: missing snapshot silently becomes first boot and may overwrite state; shutdown/snapshot lacks proven write quiescence and fsync durability; measurement changes have no rewrap/upgrade protocol; dstack HTTP has no timeout/size bound or production-provider tests; sealed-state lacks a dedicated CI job. The architecture is directionally useful, but this implementation must not ship.

PR #256, opened 2026-07-31
feat(api): workload-scoped secret contract (/v1/workload-secrets) for container env sealing

0xSolace:
## Why

elizaos/eliza#17432 (successor to the closed elizaos/eliza#17301) needs a **machine** secret path: a container control plane seals caller-supplied env secrets into Steward, and the container resolves them at boot. The human `/secrets` CRUD is intentionally owner/admin-session + recent-MFA and rejects machine credentials — correct, and untouched here. What was missing is a workload-scoped contract where **no machine credential can enumerate tenant inventory** and **no human session can operate the machine plane**.

## Contract (mounted `/v1/workload-secrets`, behind tenantAuth)

| principal | credential | may |
|---|---|---|
| writer | tenant API key (control-plane side, never inside a workload) | `POST /workloads` (register + P-256 key; re-POST = capability rotation, prior signers revoked in-tx) · `PUT /workloads/:id/secrets/:name` (upsert/rotate, versioned via SecretVault-native rotation) · `DELETE /workloads/:id` (revoke capability + soft-delete namespace) |
| reader | short-lived agent token from the public keypair-only `/agent-enroll` (A1, #247), signed by the per-workload P-256 key | `POST /resolve` — values in the token-derived namespace `workload/<agentId>/<name>` only |

- Humans (any role, any MFA state) are **denied on every route**.
- The writer can never read a value back; there is **no list endpoint** anywhere.
- Cross-workload/cross-tenant addressing is impossible **by construction** (no request field can carry a foreign namespace).
- Capability lifecycle: issue → exercise (enroll → 5m token → resolve, renew = re-enroll) → rotate (old key dead in-tx) → revoke (enroll dies instantly; an outstanding ≤TTL token resolves an already-emptied namespace).

## Built on shipped primitives — no new planes

- Identity: `agents` (`walletType="workload"`) + `agent_signers` (`signerType/subjectType="workload"`, p256). **No new tables, no migration.**
- Enrollment: A1 `/agent-enroll` challenge-response (#247), unchanged.
- Storage: existing `secrets` table + SecretVault; reads via `exerciseSecretRow` with a **fail-closed audit `beforeUse` chokepoint** — nothing joins the pinned direct-return decrypt caller set (#245 inventory test still green).
- Audit: keys-never-values events (`workload.register/capability.rotate/secret.write/secret.rotate/secrets.resolve/revoke`).

## Evidence

`packages/api/src/__tests__/workload-secrets.test.ts` — **24 pass** against the REAL app (real tenantAuth, real enrollment crypto, pglite): writer/reader auth matrices (incl. owner **with recent MFA** denied, writer-cannot-read), full lifecycle, value rotation, capability rotation (old key enrollment dies immediately), revocation (live token resolves nothing), tenant isolation (409 takeover / 404 no-oracle), workload namespace isolation, no-inventory 404s, malformed/oversized 400s, keys-never-values audit, ciphertext-at-rest.

Regression: kms-routes + agent-enroll-route + secrets-no-read-back + secrets-audit-order (28 pass), vault decrypt-caller inventory (3 pass). `turbo build --filter=@stwd/api` clean; biome clean on the diff.

## Consumer

The elizaOS side (control-plane sealing + boot resolver, flag-gated default-off) follows as a separate PR referencing elizaos/eliza#17432; its cross-repo suite runs against THIS real API.


0xSolace:
elizaOS consumer PR is up: elizaos/eliza#17440 (writer + boot resolver, flag-gated default-off). Its cross-repo suite boots THIS branch's real API (embedded pglite entry) and proves seal → enroll → resolve, writer-cannot-read, foreign-workload isolation, both rotations, revocation, and outage fail-closed — 7/7 green against this contract.

lalalune:
Closing as architecturally unsafe despite green exact-head CI. workloadId permits underscore but revocation uses an unescaped SQL LIKE prefix, so deleting foo_bar also matches fooXbar. DELETE leaves the workload agent row live, allowing immediate PUT repopulation; outstanding reader tokens can then access new values. Resolve decrypts a row loaded before concurrent delete/rotation without revalidation, registration checks existing state outside the transaction (allowing multiple active signers), and namespace writes/replace plus audit are not atomic. The claimed Eliza consumer PR is closed with changes requested and its review found related lifecycle/exfiltration/fail-open gaps, so it is not production evidence. A replacement needs an explicit revoked lifecycle/generation bound into tokens, exact workload_id ownership (not LIKE), locked/unique active signer state, atomic staged replace/activation and audit outcome, resolve-time revocation revalidation, adversarial race tests, and a pinned always-on consumer contract.

PR #259, opened 2026-08-09
fix(security): bump hono across the workspace and protobufjs override for GHSA advisories

NubsCarson:
## What

Resolves the remaining fixable `bun audit` findings on develop. **Scope note:** this branch originally also bumped nanoid, pdfjs-dist, and js-yaml — those were fixed upstream while it was in review, so the rebase defers to the upstream choices (including `pdfjs-dist ^6.2.108`) and this PR now touches only what upstream still lacks:

| Dependency | Change | Advisory |
|---|---|---|
| **hono** | `^4.12.18 → ^4.12.34` in all 8 consuming package manifests + root override (resolves **4.13.1**) | CORS ReDoS + `memo()` cross-user SSR disclosure — hono is the API framework, so these sat on the serving path |
| **protobufjs** | root override `^7.6.0 → ^7.6.5` | advisory bump |

**Remaining `bun audit` findings, deliberately not addressed:** `@hono/node-server` (via `@modelcontextprotocol/sdk`, distinct package from hono), `image-size` ×2 (via react-native), `elliptic` (no fixed upstream release; already pinned). Candidates for the threat-model's accepted-CVE list.

## Verification (post-rebase, against current develop)

- Lockfile resolves `hono@4.13.1`, `protobufjs@7.6.5`, and `pdfjs-dist@6.2.108` (upstream's pin, unchanged by this branch)
- `bun install` clean; `packages/auth` (hono consumer): **182 pass / 0 fail**
- `bun audit`: **4 findings remain** (2 high, 1 moderate, 1 low — all in the out-of-scope list above)
- Pre-rebase, the full sweep also ran clean against the bumped versions: lint 0 errors, `turbo test` 28/32 with only the known develop-baseline failures; `plugin-trading` 95/0 + `plugin-capabilities` 148/0 isolated

Independently reviewed by an adversarial pass (diff scope, lockfile churn, npm integrity hashes).


NubsCarson:
Note for maintainers: the one red check (E2E Integration Compose) is the known-flaky job — it timed out on develop itself before this branch existed (see the audit in the PR body; unit suites pass 182/0 against the bumped deps, full turbo green locally). A re-run should clear it; happy to rebase if develop moves again.

0xSolace (changes requested):
Two blocking scope/correctness findings from the dependency-fallout review:

1. The root `"hono": "^4.12.34"` resolves `hono@4.13.1`, not the security-only 4.12.34 patch. Hono 4.13.0 changes the default CORS allow-methods from `GET, HEAD, PUT, POST, DELETE, PATCH` to include `QUERY`. This repo's proxy uses `app.use("*", cors())` followed by authenticated `app.all("*", handleProxy)`, so the upgrade silently expands which methods browsers may preflight and forward through the catch-all. 4.13 also changes RegExpRouter registration behavior. The green unit suites prove broad compatibility, but do not make that security-surface expansion appropriate for this narrowly scoped advisory patch. Please pin the root resolution to exact `4.12.34` (while package manifests may retain `^4.12.34` as their published minimum), or explicitly preserve the old CORS method set and separately justify/test 4.13.

2. `bun.lock` also upgrades the entire optional native `@napi-rs/canvas` family from 1.0.3 to 1.0.5. No manifest in this PR asks for that change, and the PR body says only Hono/protobuf remain in scope. Please restore those unrelated lock entries so this security patch does not also ship an unreviewed native-binary upgrade.

I separately verified the red Compose job is infrastructure flake, not evidence against the dependency patch: PR #259 and develop run 29505605525 both fail `Start compose stack` after about five minutes with exit 143; recent develop run 31305902120 and multiple adjacent develop runs pass the same job. Still require a green rerun after the branch amendment before merge.

PR #235, opened 2026-07-16
fix(web): client-safe @stwd/shared/client entrypoint unbreaks cf:build

0xSolace:
## Summary

Closes #231.

`bun run cf:build` (OpenNext/Cloudflare Worker build) failed on every web-touching PR because `web/src/lib/utils.ts` imported the top-level `@stwd/shared` barrel for two pure helpers, and the barrel re-exports `provider-execution-auth.js` (PR4 #199) which imports `node:crypto`. webpack/OpenNext cannot resolve `node:crypto` in the Worker bundle:

```
Module build failed: UnhandledSchemeError: Reading from "node:crypto" is not handled by plugins.
  node:crypto
  ../packages/shared/dist/provider-execution-auth.js
  ../packages/shared/dist/index.js
  ./src/lib/utils.ts
  ./src/app/dashboard/page.tsx
```

Repro receipt preserved at commit base 24ae042 before any change (log kept by the lane, tail matches the issue trace exactly, exit 1).

## Fix (issue option 1, preferred)

- **New client-safe entrypoint `@stwd/shared/client`** (`packages/shared/src/client.ts` + `./client` in the exports map). It exposes ONLY modules whose full transitive graph is free of node builtins and server-only code: the chains registry, the token registry, and `describeThrown`/`UNPRINTABLE_THROWN_VALUE`.
- **Web import audit + migration.** Every `@stwd/shared` import under `web/src` was audited; exactly two modules import it (`web/src/lib/utils.ts`, `web/src/lib/chains.ts`) and both import only client-safe symbols (`getNativeDecimals`, `getNativeSymbol`, `CHAIN_PROVIDERS`, `getChainProviderByNumeric`). Both migrated to `@stwd/shared/client`.
- **No polyfills, no stubs, no webpack fallback.** `node:crypto` and `provider-execution-auth` remain fully available from the server entrypoint (`.`), which is untouched. API/server imports unchanged.
- **No Dockerfile change needed**: the new entrypoint lives inside the existing `packages/shared` workspace path and ships in `dist/` from the same `tsc` build.

## Contract test (fail-closed)

`packages/shared/src/__tests__/client-entrypoint.test.ts`:

[...]

PR #260, opened 2026-08-09
fix(test): isolate plugin-trading/plugin-capabilities suites and close CI unit-matrix gaps

NubsCarson:
## What

Three related test-infrastructure fixes (3 files, +21/−3):

1. **`--isolate` for `plugin-trading` and `plugin-capabilities` test scripts.** Both suites fail under plain `turbo test` from cross-file mock pollution (same class as #59) but pass 100% isolated. Flag placement matches the repo's existing `bun test --isolate` usage in CI.
2. **CI unit-matrix coverage:** `plugin-trading`, `plugin-capabilities`, and `trade-sessions` had **no CI unit job at all** — added to the `unit` matrix. (`signer-frost` deliberately excluded: its harness runs `cargo build` for the Rust sidecar and CI has no Rust toolchain — documented inline.)
3. **`STEWARD_AUDIT_HMAC_KEY` in the unit job env.** Found by adversarial review running the *exact CI invocation*: `bunfig.toml` `[test]` preloads only load from bun's cwd, and CI runs `bun test` from the repo root — so `plugin-trading`'s preload (which supplies this key) never runs in CI, deterministically failing 2 tests. Job-level env var with the preload's exact value shape (64-char dummy), with the root cause documented in the workflow.

Not done deliberately: `--timeout 30000` was **not** added to the shared CI run step (would raise the budget for all 12 matrix packages; the slow files already self-manage via in-code `setDefaultTimeout(30000)`, which survives any invocation cwd). `pr.yml`'s matrix has the same coverage gap (plus missing `plugin-wxmr`) — left as a follow-up to keep this diff reviewable.

## Verification

Baseline repro: `turbo test` filtered to the two plugins → plugin-trading **84 pass / 11 fail** (pollution). After:

- CI-style from repo root (`STEWARD_MASTER_PASSWORD` + `STEWARD_AUDIT_HMAC_KEY`, exact CI command): `plugin-trading` **95/0 ×2** (determinism), `plugin-capabilities` **148/0**, `trade-sessions` **21/0**
- Baseline *without* the new env var reproduces CI's failure (93/2) — proving the var is what fixes it
- `trade-policy-audit.test.ts` (manages/deletes the env var itself) re-verified CI-style: 9/0; `proxy`/`vault` suites confirmed unaffected (hard-assign or exclude the key)
- `turbo test` filtered to both plugins: 2/2 tasks, 0 fail — pollution gone
- YAML parses; matrix = 12 entries; lint 0 errors; no lockfile churn

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,13 @@ jobs:
           - packages/db
           - packages/shared
           - packages/plugin-wxmr
+          - packages/plugin-trading
+          - packages/plugin-capabilities
+          - packages/trade-sessions
+          # packages/signer-frost is intentionally absent: its test harness
+          # builds the Rust sidecar via `cargo build --release` and this job
+          # has no Rust toolchain (and a release build would strain the
+          # 10-minute timeout). Add it alongside a rust-toolchain step.
 
     steps:
       - name: Checkout
@@ -83,6 +90,11 @@ jobs:
           fi
         env:
           STEWARD_MASTER_PASSWORD: "ci-test-secret"
+          # bunfig.toml [test] preloads only load from bun's cwd, and this job
+          # runs `bun test` from the repo root — so per-package preloads (e.g.
+          # plugin-trading's, which supplies this key) never run. 64-char dummy
+          # matching the preload's "a".repeat(64).
+          STEWARD_AUDIT_HMAC_KEY: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 
   unit-redis:
     name: Unit Tests (packages/redis)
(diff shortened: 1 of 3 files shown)


0xSolace (approved):
Read the full 3-file diff and verified the test-runner semantics and CI coverage.

- `--isolate` changes execution isolation only; it does not skip files, filter assertions, relax timeouts, or mask failures. It directly addresses cross-file mock/global pollution in the two affected suites.
- The new unit-matrix entries are real coverage additions (`plugin-trading`, `plugin-capabilities`, `trade-sessions`). The documented `signer-frost` exclusion is honest because that harness invokes a Rust release build while this job provisions no Rust toolchain.
- Supplying the deterministic 64-byte dummy `STEWARD_AUDIT_HMAC_KEY` at job scope matches the package preload and compensates for Bun loading preloads from the root cwd. It is non-secret test material and does not weaken production validation.
- No lockfile/runtime/source churn and no timeout inflation.

No blocking findings. Approved. I will merge only after #235's develop CI verdict, then verify the combined head rather than relying solely on the pre-#235 run.

Issue #231, opened 2026-07-16
web cf:build fails: @stwd/shared barrel pulls node:crypto into the Cloudflare Worker bundle


0xSolace:
## Problem
The `Build check (dry run)` (`.github/workflows/deploy-web-cloudflare.yml`, `bun run cf:build`, OpenNext/webpack) **fails to compile** on any web-touching PR:

```
Module build failed: UnhandledSchemeError: Reading from "node:crypto" is not handled by plugins.
Import trace:
  node:crypto
  ../packages/shared/dist/provider-execution-auth.js
  ../packages/shared/dist/index.js   (barrel)
  ./src/lib/utils.ts                 (imports getNativeDecimals/getNativeSymbol from @stwd/shared)
  ./src/app/dashboard/page.tsx
```

## Root cause (pre-existing, NOT PR6)
`packages/shared/src/index.ts` barrel-exports `provider-execution-auth.js` (added by PR4 #199), which `import { createHmac, hkdfSync, timingSafeEqual } from "node:crypto"`. `web/src/lib/utils.ts` imports the barrel for unrelated helpers, dragging `node:crypto` into the Cloudflare Worker bundle, which webpack/OpenNext cannot resolve.

## Proof it is develop-level, not a specific PR
Removing ALL of PR6's web pages and re-running `cf:build` **still fails** identically. The workflow is path-filtered to `web/**`, so PR5 (#200, no web change) never triggered it; PR6 (#226) is merely the first web-touching PR since PR4 broke it.

## Options
1. Add a **client-safe shared entrypoint** (`@stwd/shared/client`) that excludes node-crypto modules; point `web` imports at it. (preferred)
2. Or webpack `resolve.fallback`/third-partys config for `node:crypto` in the OpenNext build, or mark server-only shared modules.
3. Or move `provider-execution-auth` out of the top-level barrel so `getNativeDecimals`/`getNativeSymbol` don't transitively pull it.

## Acceptance criteria
- `bun run cf:build` in `web/` compiles clean on develop.
- The `Build check (dry run)` PR check is green on a web-touching PR.
- No server-only crypto is bundled into the client Worker.

## Size: M. Labels: web, build, tech-debt. Priority: blocks the web Build check on every web PR.

PR #261, opened 2026-08-09
ci: build @stwd/web on develop so shared-package regressions cannot land unseen

NubsCarson:
## What

Adds a `web-build` job to **both** `ci.yml` and `pr.yml`, mirroring `deploy-web-cloudflare.yml`'s `build-check` step-for-step (checkout → bun → node 22 → frozen install → turbo build of `@stwd/react`/`sdk`/`shared` → `cf:build` → `wrangler deploy --dry-run`) — **no deploy step, no secrets, no `continue-on-error`**.

Both files need it: `ci.yml` triggers only on *pushes* to develop/main, so a `ci.yml`-only guard fires post-merge; the `pr.yml` copy is what actually blocks a regression from landing.

## Why

The `@stwd/web` build broke on develop (#231) via a `packages/shared` change **without any CI catching it** — `deploy-web-cloudflare.yml` only runs on `main` pushes / PRs touching `web/**`, so shared-package regressions sail through. This closes that exact gap.

## ⚠️ Sequencing: merge #235 first (or together)

This job **will fail on current develop until #235 lands** — that is the guard doing its job, not a defect (`continue-on-error` deliberately not set; a guard that cannot fail is decoration).

**We independently verified #235 is safe to merge** (green CI is not proof, so we built it): control build on develop fails with exactly the #231 `node:crypto UnhandledSchemeError`; treatment build on the PR head exits 0 (OpenNext + wrangler dry-run both pass); the new client entrypoint's full transitive graph (src + dist) has **zero** `node:` imports, env access, or server-only crypto reaching the browser bundle; `packages/shared` tests 354/0 isolated; and a manual mutation (re-exporting `provider-execution-auth` from `client.ts`) is caught by the new contract test — the guard has teeth.

## Verification

- `actionlint` on both workflows: exit 0; YAML parses; job present with 7 steps in each file
- Behavioral proof the guard catches the real regression: ran the job's exact commands locally on develop → turbo build 3/3, then `cf:build` **exit 1** with the #231 error (`Reading from "node:crypto" is not handled by plugins`, import trace through `packages/shared/dist/provider-execution-auth.js`)
- No secrets used; inherits each workflow's existing `permissions: contents: read`
- Known duplication: on PRs touching `web/**`, this overlaps `build-check` (~10 min) — accepted cost; the guard's purpose is the `packages/*` changes that path filter misses

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,52 @@ jobs:
       - name: Typecheck
         run: bun run typecheck
 
+  web-build:
+    # Guard for the @stwd/web Worker build. deploy-web-cloudflare.yml only
+    # triggers on main pushes / PRs touching web/**, so a packages/* change
+    # (e.g. the @stwd/shared barrel leaking node:crypto into the client
+    # bundle, issue #231) can break the web build without any check firing.
+    # Build the Worker here — no upload, no secrets — so those regressions
+    # surface immediately. Mirrors build-check in deploy-web-cloudflare.yml.
+    name: Web Build (@stwd/web)
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: "1.3"
+
+      # OpenNext tooling needs Node; pin to 22 (Node 24 breaks its CLI deps).
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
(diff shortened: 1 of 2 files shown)


0xSolace (approved):
Read both workflow diffs against the existing CI, PR, and Cloudflare dry-run workflow.

- The new job is least-privilege and build-only: no secrets, upload, deploy, write permission, `continue-on-error`, or conditional bypass.
- It closes both sides of the gap: PR-time prevention in `pr.yml`, plus post-merge detection on develop/main pushes in `ci.yml`.
- The build order is correct for workspace `dist` imports, Node 22 is explicitly provisioned for OpenNext, and the final Wrangler invocation is a dry run to a temporary outdir.
- Commands match the established build-check path; the Bun `1.3` selector also matches these workflows' existing toolchain policy.

No code-review findings. Approved subject to the required sequencing proof: refresh/re-run on current develop after #235 and require `Web Build (@stwd/web)` to pass before merge.

PR #281, opened 2026-08-16
feat(vault): add AWS KMS external custody signer

### LINKED ISSUE #214: gap(security): reference third-party-custody signing backend + HSM/KMS pluggability guide
## Gap

Custody has a documented seam for operator-supplied backends (`packages/vault/src/keystore-backend.ts:10` — "attested execution, cloud HSM custody can ship their own backend by …"), an `third-party-key-custody.ts` interface, and a contract test (`packages/vault/src/__tests__/third-party-key-custody-contract.test.ts`). But there is **no reference third-party-custody adapter and no operator guide for authoring one**. The extension point exists in principle; nobody outside the repo can exercise it.

For the "bring your own custodian" positioning (VISION: "integrate third-party identity, secrets, and custody systems without replacing them") to be real, an operator must be able to plug a KMS-signing or HSM-signing backend without patching Steward.

## Why it matters
- **Buyer:** "can I keep signing in *my* HSM/KMS and let Steward do policy+approval+evidence around it" is the enterprise custody question. The seam exists but is unproven and undocumented.
- **Differentiator:** Steward's pitch vs Turnkey/Fireblocks is "we don't replace your custodian, we govern around it." That only lands if the custody seam is real and documented.

## Proposal
- Publish the third-party-custody backend interface as a documented, versioned contract (it partly exists as a test — promote it to a doc + typed public interface).
- Ship one reference adapter (e.g. an AWS KMS *asymmetric signing* backend that never exposes the private key to the app — distinct from the existing KMS *envelope* mode which does).
- A "write your own custody backend" guide running the contract test.

## Acceptance criteria
- [ ] Public, versioned third-party-custody signing interface with the contract test as its conformance suite.
- [ ] One reference signing backend where the private key never enters app memory (KMS asymmetric or PKCS#11 signing), clearly distinguished from envelope mode.
- [ ] Operator guide: implement + register + pass the contract test.
- [ ] Claims discipline: reference adapter's posture stated precisely; no blanket "keys never in memory" claim for envelope modes.

## Size
L

Labels: gap, security

[sol-orch]

lalalune:
## Summary

- add a versioned external-custody v1 contract and reusable provider conformance probe
- add an EVM-only AWS KMS asymmetric signer using `ECC_SECG_P256K1`, `SIGN_VERIFY`, `ECDSA_SHA_256`, and `MessageType=DIGEST`
- bind every registration and signature to the KMS-derived Ethereum address and configured region
- strictly decode DER, validate scalar ranges, normalize low-s, derive recovery parity, and verify the registered address before return/broadcast
- wire external custody through a separate `STEWARD_EXTERNAL_CUSTODY_PROVIDER=aws-kms` registry path without changing envelope-KMS semantics
- document setup, IAM boundary, unsupported modes, and the precise custody posture

No private key, mnemonic, seed phrase, plaintext key, or key material is accepted by the external-custody boundary. Steward stores only an opaque KMS handle and verified public address.

## Security behavior

The adapter fails closed on unsupported chain/key modes, region/address mismatch, malformed/non-canonical/out-of-range signatures, wrong-key recovery, high-s output (normalized), semantic RPC mutation, missing RPC/KMS responses, and broadcast-before-verification. Solana, Bitcoin, messages, typed data, user operations, and arbitrary hash signing are not exposed by this reference provider.

## Exact-head validation

Validated on `e44793809a63b89c23f55b175cb6f90222af3990`:

- `bun test packages/vault/src/__tests__/aws-kms-external-custody.test.ts packages/vault/src/__tests__/external-key-custody-contract.test.ts packages/api/src/__tests__/vault-factory.test.ts` — 29 pass, 0 fail, 153 assertions
- `bun test packages/vault/src/__tests__/external-key-custody.test.ts -t "delegates transaction signing|backend binding mismatch|still routes to the provider"` — 3 pass, 0 fail, 15 assertions
- `bunx turbo run build --filter=@stwd/vault --filter=@stwd/api` — 21/21 tasks successful
- `git diff --check` — clean

AWS interactions are hermetically exercised through a mock SDK client; live AWS credentials and key creation are intentionally not required by this test suite.

Closes #214

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,15 @@ STEWARD_MASTER_PASSWORD=
 # See docs/security/custody-posture.md for the full threat model.
 # STEWARD_ACK_LOCAL_CUSTODY=true
 
+# Optional external signing custody. This is deliberately DISTINCT from
+# STEWARD_KMS_PROVIDER=aws, which is envelope encryption and still decrypts
+# wallet private keys in the Steward process. `aws-kms` stores only an opaque
+# asymmetric key handle in Steward and asks an ECC_SECG_P256K1 SIGN_VERIFY KMS
+# key to sign EVM transaction digests. The private key never enters Steward.
+# See docs/security/external-custody.md.
+# STEWARD_EXTERNAL_CUSTODY_PROVIDER=aws-kms
+# STEWARD_EXTERNAL_CUSTODY_AWS_REGION=us-east-1
+
 # [REQUIRED in production] Canonical JWT signing/verification secret used by
 # the API, proxy, auth routes, user sessions, and agent-scoped tokens. Must be
 # at least 32 characters in production. Existing deployments using
--- a/docs/introduction.mdx
+++ b/docs/introduction.mdx
@@ -60,7 +60,7 @@ Configured provider calls and wallet actions then enter Steward through authenti
 - **Credential proxy:** configured provider requests receive credentials server-side without returning them to the supported agent caller.
 - **Audit evidence:** HMAC chaining, Ed25519 checkpoints, bundle export, and offline verification, with a separate out-of-band check of the bundled public key.
 - **Self-hosting:** Docker with PostgreSQL and Redis, or embedded PGLite for local and desktop use.
-- **Custody choices:** local encrypted storage, AWS KMS envelope wrapping, a PKCS#11 wrapping adapter, and an external custody interface. Steward does not claim MPC or native HSM signing.
+- **Custody choices:** local encrypted storage, AWS KMS envelope wrapping, a PKCS#11 wrapping adapter, a versioned external-custody interface, and an EVM-only AWS KMS asymmetric signing reference provider. Steward does not claim MPC or threshold signing.
 
 ## Direction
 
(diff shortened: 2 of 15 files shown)


lalalune:
Closing as unsafe/incomplete at the governed custody boundary. The KMS crypto and key/address verification are largely strong, but a prepared transaction with to: undefined bypasses the recipient equality guard and becomes a KMS-signed contract creation. RPC-supplied gas/gasPrice are also not bounded when omitted by the caller, allowing a hostile/misconfigured RPC to inflate fees with no authorized total-fee cap. More fundamentally, the canonical governed vault/approval gateway still rejects third-party-custody and authorizations are bound to local-vault, so #214’s policy+approval+evidence acceptance path is not delivered. A replacement should require/pin recipient, bind explicit gas/fee maxima, test hostile RPC mutation before any Sign/broadcast, extend execution-authorization commitments to external custody including provider/key/address identity, preserve TOCTOU re-resolution, and add direct plus approval gateway E2Es. The DER/low-s/recovery/KMS validation work is worth preserving.

lalalune:
Re-closing: exact head e44793809a63b89c23f55b175cb6f90222af3990 is unchanged from the audited head. It still signs a prepared transaction after RPC removes transaction.to, still leaves RPC-selected gas/fees unbounded, and still cannot traverse the canonical governed external-custody execution/approval path. A CI rerun does not remediate these security/acceptance blockers.

PR #282, opened 2026-08-16
fix(deploy): serialize and pin staging images

lalalune:
## Summary
- cancel superseded Docker builds for the same ref so stale builds cannot overwrite mutable branch tags
- publish a full immutable `sha-<commit>` image tag from every Docker build
- skip staging deploys for Docker workflow completions that are no longer the `develop` tip
- serialize staging mutations and deploy the exact upstream SHA tag
- upgrade `actions/checkout` and `actions/setup-node` to their official Node 24 majors across all workflows

## Failure/race evidence
Consecutive develop Docker runs `31975671184` and `31975679428` spawned overlapping staging runs `31976043328` and `31976050487` against the same mutable `develop` tag. Both happened to succeed, but they mutated the same Railway service concurrently and did not prove which commit was deployed. Exact-head PR checks also emitted GitHub deprecation warnings for every Node-20 `actions/checkout@v4` / `actions/setup-node@v4` invocation.

## Local validation
- `actionlint .github/workflows/*.yml`
- `bun run lint`
- `git diff --check`
- verified official v7 action manifests use `node24`

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -36,13 +36,24 @@ permissions:
   contents: read
   deployments: write
 
+# Railway deploys mutate one shared staging service. Serialize manual and
+# automatic requests so two close develop pushes cannot race each other.
+concurrency:
+  group: deploy-staging
+  cancel-in-progress: false
+
 jobs:
   deploy-staging:
     runs-on: ubuntu-latest
-    # On workflow_run, only proceed if the upstream Docker build succeeded.
+    # On workflow_run, deploy only a successful Docker build for the commit
+    # that is still at the tip of develop. GITHUB_SHA for workflow_run is the
+    # default-branch tip when this downstream run is created; without this
+    # equality check, a slower Docker run for an older push can deploy after a
+    # newer commit has already landed.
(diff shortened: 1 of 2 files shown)


lalalune:
Closing because the deployment serialization/pinning claims are not yet true. workflow_run.head_sha is compared with the event’s captured github.sha, not a live develop ref after the concurrency wait, so commit A can queue, B can land, and A still passes A==A and deploys stale. Also sha-<commit> is a mutable registry tag, not a content-addressed image pin; rebuild/rerun can move it, provenance is disabled, and the exact build digest is discarded. The concurrency additions are directionally useful, but a replacement should resolve refs/heads/develop live immediately before mutation, pass and verify the upstream build digest, deploy repo@sha256:... where supported, record digest/commit provenance, and include queued-A/new-tip-B plus same-commit-retag workflow tests and exact-head CI.

lalalune:
Re-closing after live refresh: the head is unchanged at 65e464b, so the stale-event and mutable-tag findings above remain unresolved. Please reopen only with a new head that performs a live tip check and deploys a verified content digest with regression evidence.

Issue #218, opened 2026-07-16
gap(dx): SDK + Eliza plugin support for v2 provider-action lifecycle


0xSolace:
## Gap

Neither the Eliza plugin nor the TypeScript SDK supports the v2 governed-provider-action flow.

- **Eliza plugin** (`packages/eliza-plugin/src/actions/`) has `call-governed-api`, `sign-transaction`, `transfer`, `request-approval`, `list-approvals`, `check-spend` — but grep for `provider-action` / `v2/provider` across `packages/eliza-plugin/src` returns nothing. `call-governed-api` is the legacy capability-proxy path, not the new PR1-PR6 workspace-scoped provider-action path with exact-request approval + case evidence.
- **SDK** (`packages/sdk/src`) — grep for `provider-action` / `v2/provider` returns nothing. There are no typed client methods for `POST /v2/provider-actions`, the approval endpoints, or `/case` / `/evidence`.

## Why it matters
- **DX/adoption:** the whole authority-plane thesis is unreachable from the two first-party integration surfaces developers actually use. An Eliza agent cannot take a governed workspace-scoped provider action through the plugin; SDK consumers must hand-roll HTTP.
- **Dogfood:** Steward's own agent story (Sol's credentials → vault → governed access, issue #155) depends on these surfaces existing.

## Proposal
- **SDK:** typed client methods for the v2 provider-action lifecycle (invoke, get, approval get/post, execute, case, evidence) with request/response types generated from or matched to the OpenAPI (`docs/openapi.json` already has 18 `v2/provider` refs, so the API contract exists).
- **Eliza plugin:** a `provider-action` action + provider that surfaces pending approvals and case status into agent state, distinct from the legacy `call-governed-api`.

## Acceptance criteria
- [ ] SDK exposes typed methods for the full v2 provider-action lifecycle; types match OpenAPI.
- [ ] Eliza plugin action for invoking a governed provider action + surfacing approval/case state, separate from `call-governed-api`.
- [ ] Example in `docs/sdk/` and `docs/sdk/eliza-plugin.mdx`.
- [ ] No credential handling client-side; blocked/pending surfaced.

## Size
M

Labels: gap, dx

[sol-orch]

PR #280, opened 2026-08-16
feat(api): add agent-scoped provider-action status

lalalune:
## Summary

- add exact agent-JWT `GET /v2/provider-actions/:id` status route
- bind reads to the persisted provider-action tenant and submitting agent
- return one uniform 404 for malformed, absent, cross-agent, and cross-tenant ids
- share the canonical redacted intent serializer with `GET /intents/:id`
- repoint MCP status and document the least-privilege boundary
- retain separate human-session and MFA gates on approval, case, and evidence routes

## Security validation

- response parity test against `GET /intents/:id`
- credential canaries across authorization details, payload, and execution result
- legitimate `tokenAddress` metadata remains visible
- mutation-style same-tenant foreign-agent and cross-tenant denial proofs
- exact suffix-route canary proves agent JWT still receives 403 on approval, case, and evidence
- missing agent JWT receives 401

## Verification

- `bunx turbo run build --filter=@stwd/api --filter=@stwd/mcp` (22/22 tasks)
- `bun test packages/api/src/__tests__/provider-actions-route.test.ts` (10 pass)
- `bun test packages/mcp/src/__tests__/provider-tools.test.ts packages/mcp/src/__tests__/tools.test.ts` (36 pass)
- `bun test --timeout 15000 packages/api/src/__tests__/openapi-contract.test.ts` (8 pass)
- Biome check on every touched TypeScript source (clean)
- `git diff --check` (clean)

Closes #233

PR #283, opened 2026-08-16
feat(dx): add SDK and Eliza provider-action lifecycle

lalalune:
## Summary

- add typed `StewardClient.providerActions` methods for invoke, agent-scoped status, human approval get/decide, safe resume, case manifest, and signed evidence
- match the real raw and enveloped API response contracts and preserve structured authorization/MFA failures
- register a first-class `STEWARD_PROVIDER_ACTION` Eliza action, distinct from the legacy capability proxy
- track submitted action ids and expose agent-readable lifecycle/case state through an Eliza provider
- document the full lifecycle and publish complete OpenAPI contracts for every route

## Authorization boundary

- invoke and status use the agent JWT surfaces from #280
- the Eliza state provider polls only agent-scoped status; it never fetches approval detail, case manifests, or evidence
- approval get/decide remain eligible-human plus recent-MFA routes
- case/evidence remain owner/admin plus recent-MFA routes
- the Eliza action rejects credential-shaped arguments and never accepts actor or tenant substitution fields
- safe resume remains server-authorized from persisted action ownership and exact approval state; the plugin does not auto-execute

## Verification

- `bun test packages/sdk/src/__tests__` (248 pass)
- `bun run --cwd packages/eliza-plugin test` (39 pass, 24 opt-in live tests skipped)
- `bun test --timeout 15000 packages/api/src/__tests__/openapi-contract.test.ts` (8 pass)
- `bunx turbo run build --filter=@stwd/api --filter=@stwd/sdk --filter=@stwd/eliza-plugin` (22/22 tasks)
- Biome check on all touched TypeScript and navigation JSON (clean)
- `git diff --check` (clean)

Stacked on #280 because the agent-readable status method intentionally depends on its scoped `GET /v2/provider-actions/:id` route. Merge #280 first, then retarget this PR to `develop`.

Closes #218


lalalune:
Closing exact head 1bb069fb0e454889b515984f7dccf48eb916fd12 after independent review. It is stacked on incorrect #280 and has no exact-head develop CI. Independently, Eliza generates random retry keys (duplicate writes after lost responses), persisted denials are mislabeled unsubmitted and not tracked, the lifecycle provider reports generic intent rather than provider binding state, execute accepts then ignores its idempotency key, and poll errors silently erase tracked state. Focused SDK/Eliza suites passed 4/4 each but do not cover these failures. Please restack a repaired implementation on develop with stable retry identity, typed persisted-denial handling, a least-privilege real lifecycle DTO, honest execute semantics, outage-visible tracking, and full exact-head CI.

PR #284, opened 2026-08-16
feat(cli): add operational integrity doctor checks

### LINKED ISSUE #211: gap(dx): expand steward doctor (migration tip, audit-chain, redis, governed-route sanity, clock skew)
## Gap

`steward doctor` (`packages/cli/src/doctor.ts`) checks env-var presence, Ed25519 key format, `/health`, `/ready`, and two strict-mode env checks. It does **not** verify any of the operational-integrity properties an operator most needs confirmed before trusting a deployment:

- **DB migration currency** — is the schema at the expected journal tip (currently 0082)? A behind-migrations DB silently breaks governed paths.
- **Audit-chain continuity** — does the latest checkpoint verify against the chain head? (This is *the* health signal for the evidence product.)
- **Redis reachability** — the proxy/nonce path depends on it; absent Redis is a governed-execution outage, not surfaced.
- **Governed-route sanity** — are there `governed_v2` routes with a null `provider_operation_id`, or routes reachable by both legacy and governed paths (the exact release-blocker class from MASTER-PLAN §10.1)?
- **Clock skew** between API/proxy/DB — MASTER-PLAN PR7 calls skew a tested failure mode; doctor should warn.

## Why it matters
- **Operator:** a green `/health` today can coexist with a stale schema, a broken audit chain, or a governed route with a legacy bypass — all invisible.
- **Security:** the governed-route/legacy-bypass check is the cheapest continuous guard against the P0.4 "second authority source" regression.

## Acceptance criteria
- [ ] `doctor` adds: migration-tip check, audit-checkpoint-verifies check, Redis-reachability check, governed-route inventory check (no null operation, no dual-mode route), clock-skew warning.
- [ ] Each check fail-closed under `--strict`.
- [ ] No secret material in any check output (preserve `describeSecret` discipline — presence/length only).
- [ ] Governed-route check reuses the same inventory logic PR7's static route guard will use (no drift).

## Size
M

Labels: gap, dx, security

[sol-orch]

lalalune:
Closes #211

## Summary
- derive the expected migration tip from the checked-in Drizzle journal and compare it to the live DB journal
- expose Redis, DB/API/proxy clock, and shared governed-route inventory diagnostics through readiness
- add an owner/admin + recent-MFA audit-integrity endpoint that verifies the live HMAC chain and requires a valid Ed25519 checkpoint at the current head
- make `steward doctor --strict` fail closed for every operational diagnostic while retaining non-strict warning behavior
- add the proxy server time and generated OpenAPI contract

## Validation
- API/DB/proxy/CLI filtered build (22/22 tasks after type fixes)
- `bun test packages/cli/src/__tests__/doctor.test.ts` (6/6)
- `bun test packages/db/src/__tests__/migration-status.test.ts` (1/1)
- `bun test packages/api/src/__tests__/audit-bundle.integration.test.ts` (12/12)
- `bun test packages/api/src/__tests__/openapi-contract.test.ts` (8/8)
- focused Biome check and `git diff --check`

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -67602,6 +67602,185 @@
         }
       }
     },
+    "/audit/integrity": {
+      "get": {
+        "tags": [
+          "Audits"
+        ],
+        "summary": "Verify audit chain and latest checkpoint at the current head",
+        "description": "Requires an owner/admin browser session with recent MFA. Used by steward doctor to verify the live HMAC chain and the newest persisted Ed25519 checkpoint without returning signing or HMAC key material.",
+        "security": [
+          {
+            "bearerAuth": []
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "JSON response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "required": [
+                    "valid",
+                    "chainValid",
+                    "checkpointPresent",
+                    "checkpointValid",
+                    "checkpointAtHead",
+                    "checkpointSeq",
+                    "chainHeadSeq"
+                  ],
+                  "properties": {
+                    "valid": {
+                      "type": "boolean"
+                    },
+                    "chainValid": {
+                      "type": "boolean"
+                    },
+                    "checkpointPresent": {
+                      "type": "boolean"
+                    },
+                    "checkpointValid": {
+                      "type": "boolean"
+                    },
+                    "checkpointAtHead": {
+                      "type": "boolean"
+                    },
+                    "checkpointSeq": {
+                      "type": [
+                        "integer",
+                        "null"
+                      ]
+                    },
+                    "chainHeadSeq": {
+                      "type": [
+                        "integer",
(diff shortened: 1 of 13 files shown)


lalalune:
Closing because the doctor cannot currently serve as a fail-closed operational gate. --strict prints ok:false but exits 0; structured /ready 503 diagnostics are discarded by the non-2xx API client; explicit --api-url/--tenant-id/--token/--tenant-key context is built but not passed to doctor; and route-inventory SQL compares only identical strings, missing actual wildcard host/path/method overlap bypasses. /audit/integrity is also unbounded despite a bounded claim, and making Redis globally readiness-required conflicts with supported no-Redis modes. A replacement should assert nonzero strict exit with spawned-CLI tests, parse structured 503 bodies, pass the authenticated CLI context, reuse real proxy overlap semantics with broad/narrow wildcard tests, bound integrity verification, and make Redis readiness mode-aware.

PR #285, opened 2026-08-16
fix(policy): make provider reservations crash-durable

### LINKED ISSUE #240: follow-up(policy): crash-durable cumulativeSpend reservation handles (persist + C2 reconcile) (#206)
## Gap

`cumulativeSpend` / `maxCalls` reservation handles (#206, PR #237) are **in-memory to the original `createProviderAction` call** and are NOT persisted on the binding. A crash between the binding commit and the finalize (settle/release) call — and the replayable-audit-fail path — leaves the reservation as `outcome_unknown`: it ages out at the window edge (fail-closed; never frees maybe-spent budget).

Worst case is a **bounded transient over-count for one window** on a genuinely-failed replay: a deny-side error (safe), never allow-side. But it is an imprecision worth closing for exactly-once accounting.

## Anchors (develop @ post-#237)
- `packages/api/src/services/provider-action-service.ts` — `CumulativeSpendReservationHandle` / `WindowedInvokeReservationHandle` are returned from `evaluatePolicy` in-memory; `finalizeCumulativeSpend` / `finalizeWindowedInvoke` are the settle/release points; the replay path (`outcomeFromBinding`) cannot re-derive them.
- `packages/redis/src/cumulative-spend-tracker.ts` — reservation entries carry `reservationId|amount|reserved`.

## Why it matters
Precision of the money-cap accounting across process crashes. Current behavior is fail-closed (safe) but can transiently over-count one window, wrongly denying unrelated invokes until ageout.

## Acceptance criteria
- Reservation handles persisted per binding (e.g. on `providerActionBindings`), reconciled by the C2 crash-recovery sweeper.
- A crash between commit and finalize is reconciled to the correct settle/release, not left to age out.
- Requires a migration (out of scope for #237, which added none).
- Mutation-proven / crash-injection tested.

## Size
L

## Labels
policy, spend-caps, durability, follow-up

lalalune:
Closes #240

## What changed
- persist a versioned, immutable cumulative-spend/maxCalls reservation handle document in the same transaction as each provider binding
- add a guarded `pending -> settled|released` reconciliation lifecycle and partial pending index in migration 0084
- extend the C2 recovery pass to settle/release terminal bindings using idempotent Redis operations followed by a status-bound database CAS
- keep `allowed_stub`, executing, and `outcome_unknown` reservations pending/fail-closed until a known terminal outcome exists
- make windowed-invoke release propagate Redis errors so reconciliation can never record false completion
- enable the existing API Redis-gated suite in PR integration CI, which already provisions Redis

## Crash/race guarantees
- crash before Redis: next sweep repeats the operation
- crash after Redis but before DB CAS: next sweep repeats the idempotent operation safely
- concurrent workers may both issue the idempotent operation, but exactly one wins `pending -> terminal`
- reservation handles are frozen by the binding trigger and contain no raw action/policy text or credentials

## Verification
- `bunx turbo run build --filter=@stwd/api...` (21/21)
- direct builds: `packages/db`, `packages/redis`, `packages/api`
- migration/invariant test: 12 pass, 0 fail
- provider action service: 10 pass, 0 fail
- approval concurrency with execution auth secret: 12 pass, 0 fail
- Redis package without daemon: 21 pass, 46 expected gated skips
- root lint: exit 0 (pre-existing warnings/info only)
- Biome changed-file check and `git diff --check`: clean

The three new real-Redis crash/race E2Es are explicitly enabled in the PR Postgres integration job; local execution could not reach a Redis daemon and Docker was not running, so exact live Redis proof is delegated to that required CI service.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -302,6 +302,9 @@ jobs:
           STEWARD_EMAIL_CODE_SECRET: "ci-email-code-secret-must-be-at-least-32-chars-long"
           STEWARD_EXECUTION_AUTH_SECRET: "v2-ci-1:ci-execution-auth-secret-must-be-at-least-32-bytes-long"
           STEWARD_AUDIT_HMAC_KEY: "fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210"
+          # Exercise the API's real Redis-backed reservation/crash-recovery E2Es;
+          # this job already provisions a healthy Redis service above.
+          STEWARD_REDIS_TESTS: "1"
           SIWE_ALLOWED_DOMAINS: "steward.fi,localhost"
           # The whole api suite drives this one server from a single IP; the
           # default 100-req/60s per-IP gate 429s the sharded test flood.
@@ -345,6 +348,7 @@ jobs:
           STEWARD_EMAIL_CODE_SECRET: "ci-email-code-secret-must-be-at-least-32-chars-long"
           STEWARD_EXECUTION_AUTH_SECRET: "v2-ci-1:ci-execution-auth-secret-must-be-at-least-32-bytes-long"
           STEWARD_AUDIT_HMAC_KEY: "fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210"
+          STEWARD_REDIS_TESTS: "1"
           SIWE_ALLOWED_DOMAINS: "steward.fi,localhost"
 
       - name: Stop API test server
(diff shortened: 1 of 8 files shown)


lalalune:
Closing current implementation after exact-head audit (live head changed from 548c192 to ae6c367 during review). The durable reservation primitive is promising, but the PR does not meet crash-recovery correctness yet: there is no autonomous startup/periodic sweep, approval can race release and later falsely settle the released original handle, reconciliation failures are silently swallowed with no operational evidence, and unsafe integers can be reserved then become unparsable/unreconcilable. It also needs restacking on merged #278, which changes the same immutable request envelope/service, plus binding tests and fully green live-head CI. Please replace with a bounded recovery worker, explicit reservation generations for approval, observable retry/dead-letter behavior, safe-integer validation, race/fault tests, and exact-current-base evidence.

PR #241, opened 2026-07-16
feat(provider): config-driven generic-http governed provider profile + fail-closed profile registry (#201)

0xSolace:
Closes #201.

## What

A config-driven **`generic-http.provider-action.v1`** profile: an operator can govern a NEW public HTTPS host via config only (no bespoke adapter package), flowing through the identical pipeline as github/x (access -> policy -> approval -> execution -> evidence). Plus a **fail-closed profile registry** replacing the closed `github|x` union.

### 1. Profile registry (`packages/shared/src/provider-profile-registry.ts`)
- Frozen, code-side, compile-time registry keyed by stable profile string. `assertRegisteredProfile` / `isRegisteredProfile` / `isConfigDrivenProfile`.
- Replaces `canonicalProfile: "github.provider-action.v1" | "x.provider-action.v1"` (provider-approval.ts) with a registry-validated `string`. **github/x string VALUES are unchanged**, so the golden digest corpus does not move (verified).
- **Fail-closed at every enumerated consumption site**: store/binding write (`provider-action-service`), approval-arm commitment (`provider-approval`), and proxy dispatch (`governed-execution`). An unregistered profile is rejected (`CANON_PROFILE_UNSUPPORTED` / stale-dependency at the proxy boundary), never stored, never dispatched.

### 2. Generic-http profile (`packages/shared/src/generic-http-provider-action.ts`)
- **Operator descriptor** (stored per `provider_operations.request_profile.operationDescriptor`, no new column), STRICTLY validated at authoring time (`validateGenericHttpDescriptor`, `DESCRIPTOR_*` fail-closed codes):
  - origin: absolute https, real DNS host, no IP literal / userinfo / wildcard / non-443 port / raw-unicode homograph;
  - method allowlist; path template with **typed named segments** (string+anchored-regex / int / uuid); typed allowlisted query params; allowlisted request headers (**credential/hop-by-hop/forwarding/method-override headers rejected at descriptor time**, so a credential header can never enter the digest); JSON body schema (scalar fields); projection of validated scalars into `policyArgs` + `safe_summary` (never raw JSON).
- **Per-request canonicalization** (`buildGenericHttpAction`) reuses the ONE shared JCS/hash/normalize surface -> **byte-identical framing** to github/x. Dynamic segments are `encodeRfc3986`-encoded and re-run through `normalizePath` (traversal-proof: `/`, `.`, `..`, `%`, backslash all escaped/denied pre- and post-decode).
[...]

PR #287, opened 2026-08-16
feat(provider): secure config-driven generic HTTP operations

### LINKED ISSUE #201: gap(provider): config-driven generic-HTTP governed provider adapter + profile registry
## Gap

Steward has exactly two governed provider adapters: `github` and `x` (`packages/provider-github/`, `packages/provider-x/`). Every new provider today requires (a) a new package, (b) hand-written per-operation argument schemas + canonical action construction (`packages/provider-github/src/operations.ts:1-253`), and (c) editing a **hardcoded profile union** in shared code (`packages/shared/src/provider-approval.ts:39` — `canonicalProfile: "github.provider-action.v1" | "x.provider-action.v1"`). There is no way for an operator to govern a REST endpoint that Steward's authors have not pre-integrated.

Buyers evaluating "the open execution authorization layer for enterprise agents" expect to point Steward at *their* internal/partner HTTP API and get scoped grants + exact-request approval + evidence without waiting for a bespoke adapter package. This is the single biggest breadth gap.

## Proposal: a `generic-http.provider-action.v1` config-driven adapter

A declarative provider profile an operator authors (YAML/JSON, stored per `provider_operations` row) that specifies:
- allowed origin (public-HTTPS/SSRF constraints reused from the proxy)
- an operation template: method, path template with typed named segments, allowlisted query params (typed), allowlisted headers (never credential headers), body JSON-schema
- which validated scalar fields become `policyArgs` (never raw JSON — mirror `operations.ts` discipline)
- a `safe_summary` projection

The canonicalizer (`canonicalizeRawInternalAction`) is already generic; the missing piece is a **profile registry** so `canonicalProfile` is an open string keyed to a registered, versioned profile descriptor instead of a closed union.

## Why it matters
- **Buyer:** unblocks self-serve provider onboarding — the core self-hostable-product promise.
- **Architecture:** forces the profile-union hardcoding (`provider-approval.ts:39`, `provider-action.ts`) to become a registry, which every future provider needs anyway.

## Competitive
- **Parity:** OpenKey / OAuth-for-agents patterns assume the operator governs arbitrary downstream APIs, not a fixed vendor list. Turnkey/Privy policy engines are wallet-signing-scoped; Steward's differentiator is *arbitrary provider HTTP* — but only if it's config-driven, not adapter-per-vendor.

## Acceptance criteria
- [ ] `canonicalProfile` is an open registered string; the union at `provider-approval.ts:39` is removed in favor of a profile registry with a stable key + version + descriptor.
- [ ] An operator can register a governed operation for a new HTTPS host via config only (no new TS package), and it flows through access → policy → approval → execution → evidence identically to `github.issue.list`.
- [ ] Path segments, query, headers, body validated fail-closed with stable `CANON_*` codes; credential headers provably excluded from the digest (golden vector).
- [ ] SSRF/public-HTTPS constraints from the proxy apply unchanged.
- [ ] Golden canonicalization corpus for the generic profile shared by API + proxy.

## Size
L

Labels: gap, provider

[sol-orch]

lalalune:
Closes #201

Supersedes the closed, incomplete #241 with the production authoring and dispatch gaps repaired.

## What is now proven
- operator registration accepts only a validated `generic-http.provider-action.v1` descriptor, derives the credential host/method/path constraints, and rejects widening before commit
- the canonical descriptor is strict at every object layer, forbids credential headers and prototype pollution, limits collection/input sizes, and admits only a bounded finite-time regex subset (no lookaround/backrefs/unbounded or nested repetition)
- profile consumption remains registered/fail-closed across storage, approval binding, dispatch, and evidence
- governed internal dispatch can use the exact DB-bound public DNS host without a second env allowlist; legacy/direct proxy still requires `STEWARD_PROXY_ALLOWED_HOSTS`
- private/IP/invalid hosts, traversal, descriptor/route mismatch, and unknown profiles fail closed
- a real encrypted credential route is selected, injected, forwarded through the DNS-vetted proxy seam, terminal state is recorded, and both dispatched/succeeded evidence events are asserted

## Exact local validation
- generic shared + proxy alias suites: 55/55
- generic API operator/policy/approval E2E: 8/8
- generic governed proxy dispatch/evidence test: 1/1
- shared and proxy builds pass
- prior focused API dependency build and migration suite pass
- Biome focused check and `git diff --check` clean

Migration is numbered 0085 because open crash-reconciliation PR #285 owns 0084.


lalalune:
Closing exact head a1411517650dc32ff6c30fcbb08935bcecafba15 after independent production-path review. The security work is substantial, but the real operator flow is still unusable: registerOperation never promotes/binds the secret route to governed_v2/providerOperationId; only test helpers do, and dispatch rejects before decrypt without that binding. Dynamic descriptor paths also cannot compose with the single concrete secret route, registration omits actor/credential/binding checks, and patterns such as ^admin$|user$ bypass the intended full anchoring (eviluser passes). The proxy test also does not assert actual credential header injection/non-leak. Local builds passed 21/21, shared tests 39/39 with 8/8 mutations killed, and targeted API/proxy tests 61/61, but their DB seeding bypasses the missing production wiring. Preserve the strict schema/DNS/pinned-forwarder/governed allowlist work in a replacement that proves registration → policy → approval → real route binding → credential injection → dispatch → non-leaking evidence through public APIs, with dynamic-path and regex adversarial cases. Migration numbering must also be rebased because 0085 currently assumes closed #285’s 0084.

Issue #220, opened 2026-07-16
gap(security): provider-action profile registry + shared canonicalization conformance harness


0xSolace:
## Gap (test/architecture debt noticed while reading)

The canonical-profile identifier is a **closed union type** duplicated across the codebase rather than a registry, which is both an extensibility blocker (see the generic-HTTP issue) and a source of drift/test debt:

- `packages/shared/src/provider-approval.ts:39` — `canonicalProfile: "github.provider-action.v1" | "x.provider-action.v1"`.
- `packages/shared/src/provider-action.ts:657` — `GITHUB_PROVIDER_ACTION_PROFILE` const; `x-provider-action.ts:45` — `X_PROVIDER_ACTION_PROFILE`.
- Golden-vector tests are per-profile (`provider-action-golden.test.ts`, `x-provider-action.test.ts`) with no **shared conformance harness** asserting that *every registered profile* satisfies the invariants (credential headers excluded from digest, duplicate-key rejection, path-traversal rejection, JCS determinism).

## Why it matters
- **Architecture:** every new provider (Slack, Google, AWS, generic-HTTP) must edit the union in shared — a scaling wall and a merge-collision magnet across provider lanes.
- **Security test debt:** the canonicalization safety invariants (MASTER-PLAN §9) are re-asserted ad hoc per profile. A new profile can silently omit a safety property because there's no harness that runs *all* invariants against *all* profiles.

## Proposal
- Introduce a profile registry: `registerProviderActionProfile({ key, version, descriptor })`; `canonicalProfile` becomes a validated registered string, not a union.
- A **shared profile-conformance test harness** that, for every registered profile, asserts: credential headers never enter the digest, duplicate JSON/query keys rejected, path traversal/encoded-traversal rejected, JCS output deterministic, unsupported content types denied.
- Existing github/x profiles register through it; their bespoke golden tests remain but the invariant suite is shared.

## Acceptance criteria
- [ ] Profile registry; union at `provider-approval.ts:39` replaced by a registered-key type.
- [ ] Shared conformance harness runs the §9 safety invariants against every registered profile.
- [ ] github + x profiles migrated to the registry with no golden-vector byte changes.
- [ ] Adding a profile that violates an invariant fails the shared harness (mutation-proven).

## Size
M

Labels: gap, security, dx

[sol-orch]

lalalune:
[...]

PR #288, opened 2026-08-16
test(security): enforce provider profile conformance

lalalune:
Closes #220

Stacked on #287; retarget to `develop` after #287 merges.

## Summary
- attach a deeply immutable conformance fixture/build callback to every registered profile
- enumerate github, X, and generic HTTP through the same determinism and adversarial invariant harness
- require credential exclusion, duplicate-key/query rejection, plain/encoded traversal rejection, and unsupported-content-type rejection for every profile
- add a registered-profile key union and preserve the private fail-closed registry
- prove the harness itself is load-bearing with a deliberately violating fixture that fails every invariant
- retain the existing byte-exact github/X/generic golden corpus and DB/API/proxy boundary checks from #287

## Validation
- shared profile suite: 41/41
- `@stwd/shared` build passes
- focused Biome check and `git diff --check` clean

### DIFF
--- a/packages/shared/src/__tests__/generic-http-provider-action.test.ts
+++ b/packages/shared/src/__tests__/generic-http-provider-action.test.ts
@@ -32,8 +32,11 @@ import {
   GOLDEN_VECTORS,
   genericDescriptorAllowsExactPath,
   genericHttpCanonicalActionBytes,
+  getProfileDescriptor,
   isRegisteredProfile,
+  type ProviderProfileDescriptor,
   REGISTERED_PROFILES,
+  runProfileConformance,
   UnregisteredProfileError,
   validateGenericHttpDescriptor,
   X_GOLDEN_VECTORS,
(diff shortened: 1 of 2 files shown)


lalalune:
Closing exact head 3eb58ed9bac9eea23dc773070ee3982a2e0e4470. It is stacked on closed unsafe #287, and the conformance harness is self-attested rather than production-bound: registry entries supply handwritten build/reject callbacks unrelated to the real provider adapters/routes/credential proxy/approval/evidence path, so unsafe production code can coexist with safe fixtures and pass. A direct adversarial probe confirmed runProfileConformance returns no violations for that split. Any exception counts as a good rejection, credential exclusion tests only one literal canary, the duplicate-key case is duplicate header pairs rather than JSON keys, and the hardcoded profile union remains duplicated. Exact shared tests pass 41/41 (143 assertions), but this stacked base triggers no required PR workflow. Replace from develop with one typed profile spec that drives production builders and tests, stable error codes, digest/metamorphic/credential-noninterference checks, mutation against production functions, and full public API → binding → approval → credential injection → forward → redacted evidence E2Es.

PR #290, opened 2026-08-16
feat(audit): durable per-tenant retention archives

### LINKED ISSUE #216: gap(compliance): audit retention policy + chain-safe pruning + bulk archival export
## Gap

There is no **audit/evidence retention policy** and no **bulk compliance export** surface. `/audit/bundle` exports correlated evidence per case/range (`packages/api/src/routes/audit.ts`), but:
- There is no configurable retention (how long audit_events / approvals / execution records are kept, and how they're purged without breaking the hash chain).
- There is no bulk/scheduled export in a compliance-friendly format for archival (the current bundle is per-request, verifier-oriented, not "hand my auditor 12 months of events in a standard format").

## Why it matters
- **Compliance buyer:** retention + defensible deletion is a baseline audit requirement (and a GDPR/data-minimization requirement — you cannot keep everything forever, but you also cannot break the audit chain by deleting arbitrarily).
- **Ops:** unbounded audit growth is an operational problem with no documented answer.

## Proposal
- A configurable retention policy per tenant (min/max windows), with a **chain-safe truncation** procedure: sealed checkpoints let you archive+prune events below a checkpoint while keeping the checkpoint chain verifiable (archive the pruned range as a signed bundle first).
- A bulk export command (`steward audit export --from --to --format`) producing signed bundles suitable for archival, distinct from the interactive verifier bundle.

## Acceptance criteria
- [ ] Configurable retention window; documented chain-safe pruning that archives-then-prunes below a sealed checkpoint.
- [ ] Pruning provably does not invalidate remaining-chain verification (test: prune, then verify head).
- [ ] Bulk export command producing signed archival bundles; format documented.
- [ ] GDPR/data-minimization note in docs.

## Size
M

Labels: gap, compliance

[sol-orch]

lalalune:
Closes #216

## Summary
- replace the global destructive audit-retention env switch with explicit per-tenant policy (disabled by default) behind the existing owner/admin + recent-MFA audit gate
- durably seal versioned Ed25519-signed JSONL chunks and a manifest receipt before any source deletion
- verify the live HMAC source, receipt digest/signature/identity, contiguous floor, exact ending HMAC, and exact delete count; atomically advance floor + delete + mark pruned
- make archive and prune resumable/idempotent, tenant-isolated, and bounded (50k events/run, 10k/chunk)
- add CLI export plus offline verifier with independently trusted signing-key fingerprint support
- document durability, trust, backup, privacy, and non-repudiation limits
- use migration 0086 (0084 is active in PR #289; 0085 is reserved by the coordinated #201 repair lane)

## Exact-head verification
Head: `8ace53989f63de4d5a57f6498cf23c58d600e6ca`

- `bun test --timeout 120000 packages/api/src/__tests__/audit-retention-archive.test.ts` — 3 pass, 0 fail, 20 assertions; includes migration application, policy isolation/bounds, archive-before-delete, signed offline verification, tampered JSONL rejection, idempotent resume/prune, receipt signature corruption rejection, contiguous-floor refusal, and post-prune chain verification
- `bun test --timeout 120000 packages/api/src/__tests__/audit-chain-strict.test.ts packages/api/src/__tests__/audit-retention-archive.test.ts packages/api/src/__tests__/retention.test.ts` — 5 pass, 5 DATABASE_URL-only skips, 0 fail, 24 assertions
- `bun test packages/api/src/__tests__/auth-audit-hardening.test.ts packages/api/src/__tests__/secret-bearing-cache-control.test.ts` — 20 pass, 0 fail, 228 assertions (owner/admin session, recent MFA, and no-store audit gates)
- `bunx turbo run build --filter=@stwd/api --filter=@stwd/cli` — 22/22 tasks successful
- Biome on all changed TS/JS files and `git diff --check` — clean

## Security properties
There is no boolean/environment attestation that permits audit deletion. A sealed status alone is insufficient: prune re-hashes the canonical manifest, verifies Ed25519 and archive/tenant/range identity, matches the live ending HMAC, deletes the exact range, and advances the chain floor in one transaction. A crash before prune leaves source rows; a prune failure rolls back deletion and floor movement together.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -343,12 +343,8 @@ STEWARD_AGENT_TOKEN=
 # indefinitely as a ledger record. Default: 365 days.
 # STEWARD_RETENTION_FAILED_TX_DAYS=365
 
-# audit_events — tamper-evident SOC2 audit chain. By default, audit events
-# are NEVER deleted, because SOC2 CC2 / CC7 expect at least one year of
-# retention and most assessors prefer 3+ years. Set this env var to opt in
-# to deletion; a warning will be logged each sweep if the value is below
-# 365 days. For values below 365, also set
-# STEWARD_RETENTION_AUDIT_ARCHIVE_CONFIRMED=true after archiving the audit
-# chain prefix externally.
-# STEWARD_RETENTION_AUDIT_EVENTS_DAYS=
-# STEWARD_RETENTION_AUDIT_ARCHIVE_CONFIRMED=false
+# audit_events — no destructive environment switch exists. Events are never
+# deleted by default. An owner/admin with recent MFA must enable the tenant's
+# retention policy through /audit/retention-policy; Steward then durably seals
+# signed JSONL chunks before a separate transactional floor+delete step. See
+# docs/security/audit-retention-archives.md.
--- a/docs/security/audit-retention-archives.md
+++ b/docs/security/audit-retention-archives.md
@@ -1,0 +1,27 @@
+# Audit retention archives
+
+Steward audit events are retained indefinitely unless an owner or administrator explicitly enables a tenant retention policy. Retention is tenant-scoped, requires a recent-MFA session, and never treats an environment flag as proof of backup.
+
+## Safe workflow
+
+1. Configure `STEWARD_AUDIT_SIGNING_KEY` with an Ed25519 private key and publish the SHA-256 fingerprint of its SPKI public key through an independent channel.
+2. An owner or administrator with recent MFA calls `PUT /audit/retention-policy` with `enabled`, `retentionDays` (30–3650), and `archiveChunkSize` (1–10,000).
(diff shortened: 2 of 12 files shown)


lalalune:
Closing exact head 8ace53989f63de4d5a57f6498cf23c58d600e6ca after independent destructive-path testing. Prune verifies only the signed manifest row/live end HMAC, not that archive chunks still exist or match. Exact reproduction: seal seq 1, delete its only audit_archive_chunks row, then prune succeeds and deletes the live source—leaving zero archived bytes. The same mutable DB and self-stored public key are also trusted for the destructive decision; policy/audit and prune/completion evidence are not atomic; scheduled archives cannot be listed/exported; installed CLI verification is cwd-relative/non-resumable; stale policy/concurrent and tenant-starvation paths are unhandled. Targeted tests pass 5 with 5 DB skips, builds 22/22, but miss the reproduced data-loss case. Preserve the tenant locking/HMAC/Ed25519/offline-verifier work in a replacement that verifies the full chunk set under lock, requires trusted immutable external durability acknowledgement before deletion, binds a trusted key id, transactionally records authorization/outcome, provides list/resumable export/restore, isolates tenant failures, adds real Postgres races, and uses current migration numbering.

PR #329, opened 2026-08-17
fix(security): wave-2b plugins batch — deferred LOW/INFO findings (SEC-095/096/109/112/170/171/172/173)

lalalune:
# Security wave 2b — plugins batch (LOW/INFO deferred findings)

Deferred LOW/INFO remainder for the **plugins** batch. Builds on the merged wave-2 HIGH/MEDIUM fixes; no re-litigation of earlier decisions.

## Status

| SEC ID | Severity | Status | Summary |
| --- | --- | --- | --- |
| SEC-095 | LOW | FIXED | `submit-trade.ts` `stewardApiUrl()` now applies the shared `assertSecureApiUrl` guard (was reading `STEWARD_API_URL` raw, sending the agent JWT over plaintext for non-localhost `http://`). `validate` fails closed (returns false) on an insecure URL; the handler surfaces it as a config error instead of "venue error". |
| SEC-096 | LOW | FIXED | Capability audit `ipAddress` no longer records raw `X-Forwarded-For`. New `client-ip.ts` mirrors the core's `trustedClientIp` (same env contract: `STEWARD_TRUSTED_PROXY_HOPS` / `STEWARD_TRUST_CLOUDFLARE` / legacy `STEWARD_TRUST_PROXY_HEADERS`, right-most-N read, `isIP` validation). No trust configured ⇒ NULL recorded, never a spoofable value. |
| SEC-109 | LOW | FIXED | `decodeJobCreatedId` now skips receipt logs whose emitter isn't the configured `agenticCommerce` address (case-insensitive), matching the erc8004 register filter. A lookalike `JobCreated` from another contract can no longer yield a jobId that later `setBudget`/`fund` calls would fund. |
| SEC-112 | LOW | FIXED (docs) | Documented the single-RPC trust assumption at the trust boundaries: `REGISTRY_CONFIGS` header, `RegistryConfig.rpcUrl`, both `verified` result docs (erc8004), and `ERC8183ChainConfig.rpcUrl` (erc8183, already operator-registered-only). Production operators must supply their own `RegistryConfig.rpcUrl`/`publicClient`. |
| SEC-170 | INFO | PARTIAL/DECISION | Documented the plugin trust bar prominently in operator-facing `docs/plugins/README.md` and `packages/api/src/plugin.ts`: plugins run in-process with full trust (raw `vault` + `db`); fail-closed guarantees cover contribution collisions, not privilege. The policy-gated signing facade is a considered future hardening, not implemented — sandboxing would be a subsystem rewrite, out of scope for a LOW-wave fix. |
| SEC-171 | INFO | VERIFIED | `StewardProxyClient` already signs whenever a signing secret is configured. Added a regression test proving signed proxy calls with explicit enforcement off; clarified that `signingRequired` only governs the separate fail-closed rule when a required secret is absent. |
| SEC-172 | INFO | FIXED | Broadened MCP `sanitizeProviderPayload` with the shared keyed classifier, safe accessor/cycle handling, JWT and provider-token shapes, modern `sk-proj` keys, AWS access-key IDs, and armored private keys. Request/HMAC signatures are redacted, while generic public transaction `signature` fields remain intact for provider compatibility. |
| SEC-173 | INFO | FIXED (docs) | Chose the audit's "loud comment" option over gating: the route exists to prove the plugin seam dependency-free, and gating would change the example's character + its e2e test. Added unmissable warnings at the mount site, the file header, the package README, and `docs/plugins/README.md` quickstart: deliberately unauthenticated, do not copy the pattern, real routes must use ctx auth middleware. |

Every assigned ID appears above; no omissions.

## Trade-offs / decisions

- **SEC-096**: the plugin cannot import `@stwd/api`'s `trustedClientIp` (the core↔plugin one-directional dependency rule), so the logic is mirrored in `client-ip.ts` reading the same env vars. If the core's resolver changes, the mirror must be updated — noted in the file header.
- **SEC-171**: the audit concern was already satisfied: `StewardProxyClient` signs iff a secret is configured. The regression test now observes the signature with enforcement off; the service comment distinguishes this from the production/opt-in rule that rejects an absent secret.
- **SEC-172**: redaction remains best-effort, but keyed classification now reuses the shared credential vocabulary and traversal fails closed on accessors, cycles, and excessive depth. Free-text secret shapes remain conservative to avoid redacting transaction hashes and ordinary identifiers. `session` is still not treated as a key by itself.
- **SEC-170 / SEC-112**: docs-only per the audit's own fix options ("document the trust bar/assumption").

## How tested

Targeted suites per touched package (all green):

- `packages/eliza-plugin`: `bunx vitest run` — 52 passed / 24 skipped (7 files); new tests pin SEC-095 (plaintext URL rejected before any fetch; loopback http still allowed) and SEC-171 (signs with secret present even when enforcement flag is off). `tsc --noEmit` clean.
- `packages/plugin-capabilities`: `bun test --isolate` full suite — 163 pass (12 files), incl. new `client-ip.test.ts` (7 tests: no-trust ⇒ NULL, hops=1/2 positional reads, envoy port stripping, garbage rejection, legacy flag). `tsc --noEmit` clean.
- `packages/erc8183`: `bun test src/__tests__` — 7 pass, incl. lookalike-emitter-first and foreign-emitter-only regressions. `tsc --noEmit` clean.
- `packages/erc8004`: `bun test src/__tests__` — 3 pass. `tsc --noEmit` clean. (comment-only change)
- `packages/mcp`: original full-package run was green; the rebased exact-head focused `provider-tools.test.ts` run is 17/17 green, including adversarial key, free-text, modern provider-token, armored-key, accessor, cycle, and public-transaction-signature compatibility coverage. `tsc` clean. The existing mutation-proof script's target lines were left structurally intact.
- `packages/plugin-example`: `bun test src/__tests__` — 2 pass. `tsc --noEmit` clean. (comment/README-only change)
- `packages/api`: comment-only change to `plugin.ts`; parsed by biome (no `tsc` run — pure docblock addition).
- `bunx biome check` clean on all changed files (one format fixup commit).

Pre-existing issue noted (not caused by this batch): running the plugin-capabilities suite WITHOUT `--isolate` fails 2 `session-broker-e2e` tests (vault decrypt "unable to authenticate data" — cross-file env pollution; the package's own `test` script mandates `--isolate`, under which all 163 pass). Re-verified on this branch: the same 2 failures occur with this batch's new `client-ip.test.ts` removed from the run, and the file passes standalone. All other packages show no failures.

## Review-wave scrutiny notes

- SEC-096 changes what gets written to the capability audit trail: with no trusted-proxy env configured, `ipAddress` is now NULL where it previously recorded a client-supplied header value. That is the intended fail-closed direction; operators who want audit IPs set `STEWARD_TRUSTED_PROXY_HOPS` exactly as for the core auth limiter.
- SEC-109 changes receipt decoding only by narrowing which logs qualify; the "event not found" error path is unchanged.
- SEC-095 can turn a previously-working (but insecure) `http://<remote>` eliza deployment into a config error at trade time — intended; loopback http is unaffected.

## Independent final audit repairs

The final independent review rebased this series onto current `develop` and added four corrections:

- Multi-hop proxy attribution now fails closed when positional `X-Forwarded-For` is absent, short, or invalid. It never falls back to `X-Envoy-External-Address`, which identifies the adjacent proxy in that topology. Core and plugin copies have hostile regression tests.
- MCP redaction now covers modern `sk-proj` keys, AWS `AKIA`/`ASIA` access IDs, and armored private keys, while preserving public transaction signatures and still removing request/HMAC signatures.
- The full-RCE/raw-vault plugin trust boundary is visible in operator-facing plugin documentation.
- ERC-8183 receipt decoding skips a runtime-malformed log address rather than aborting before examining later authoritative logs.

Exact repaired head: `713a10a94aefe81cd86a83c79429dd45965ee36d`, rebased onto `develop` `a71b778420c838d438f419bc475863f1ab61a5dc`. Final local verification: focused dependency build 24/24 tasks; 33 plugin-capabilities/MCP/ERC tests, 24 auth/PGLite tests, 5 real proxy-signing verifier tests, and 10 submit-trade Vitest tests all green; Biome and `git diff --check` clean. Fresh exact-head GitHub CI is required before merge.

### DIFF
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -68,6 +68,8 @@ mount hono routes/middleware onto the core app. `ctx` is injected: it carries th
 
 ```ts
 register(app, ctx) {
+  // NOTE: /example/ping is deliberately unauthenticated (constant payload,
+  // hello-world only). Any route touching real state MUST be gated:
   app.get("/example/ping", (c) => c.json({ ok: true }));
   // gate a route: app.post("/example/do", ctx.requireAgentJwt, handler);
 }
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -41,6 +41,12 @@ migrations run at boot before the server accepts traffic. that's it - the core i
 
 ## door 2 - write a plugin (author)
 
+> **Security boundary:** plugins execute in-process with full host trust. The
+> injected context includes the raw database and vault, so a plugin can read or
+> rewrite tenant state and sign outside the policy engine. There is no sandbox.
+> Register only code you would trust with root access and master signing keys;
+> pin and audit its dependency tree, and never load an untrusted runtime plugin.
+
 a plugin is one object. you import everything from a single package:
 
 ```ts
--- a/packages/api/src/__tests__/auth-client-ip.test.ts
+++ b/packages/api/src/__tests__/auth-client-ip.test.ts
@@ -263,10 +263,16 @@ describe("trustedClientIp", () => {
     expect(
       await probeIp({ "cf-connecting-ip": "9.9.9.9", "x-forwarded-for": "198.51.100.7" }),
     ).toBe("9.9.9.9");
-    // Flag on + invalid CF value falls through to the XFF path.
-    expect(await probeIp({ "cf-connecting-ip": "junk", "x-forwarded-for": "198.51.100.7" })).toBe(
-      "198.51.100.7",
-    );
+    // Flag on makes CF authoritative. Missing/invalid values fail closed and
+    // cannot fall through to client-controlled proxy headers.
+    expect(
+      await probeIp({
+        "cf-connecting-ip": "junk",
+        "x-forwarded-for": "198.51.100.7",
+        "x-envoy-external-address": "203.0.113.9",
+      }),
+    ).toBeNull();
+    expect(await probeIp({ "x-envoy-external-address": "203.0.113.9" })).toBeNull();
   });
 });
 
--- a/packages/api/src/__tests__/auth-client-ip.test.ts
+++ b/packages/api/src/__tests__/auth-client-ip.test.ts
@@ -225,13 +225,14 @@ describe("trustedClientIp", () => {
         "x-forwarded-for": "203.0.113.5, 198.51.100.7, 10.0.0.9",
       }),
     ).toBe("198.51.100.7");
-    // ...but Envoy still rescues when the positional read cannot parse.
+    // A missing positional entry is a topology failure. Envoy names the
+    // adjacent proxy here and must not be accepted as the external client.
     expect(
       await probeIp({
         "x-envoy-external-address": "10.0.0.9",
         "x-forwarded-for": "203.0.113.5",
       }),
-    ).toBe("10.0.0.9");
+    ).toBeNull();
   });
 
   it("parses ip:port and [ipv6]:port in the trusted XFF position; bare IPv6 is never mangled", async () => {
--- a/packages/api/src/plugin.ts
+++ b/packages/api/src/plugin.ts
@@ -43,6 +43,23 @@
  *
  * that way `@stwd/api` itself stays trading-free for anyone importing it, while a
  * deploy that wants trading opts in by registering the plugin.
+ *
+ * PLUGIN TRUST MODEL (SEC-170)
+ * ----------------------------
+ * plugins run IN-PROCESS with FULL trust — there is no sandbox. the injected
+ * ctx hands a plugin the live `vault` (it can sign arbitrary transactions for
+ * any agent/tenant, bypassing the policy engine) and the raw drizzle `db`
+ * handle (it can rewrite `agent_policies`, insert `secret_routes`, read
+ * ciphertext). the host's fail-closed guarantees cover CONTRIBUTION COLLISIONS
+ * (duplicate plugin names, missing deps, cycles, migration conflicts), NOT
+ * privilege: a malicious or compromised plugin package is full RCE with
+ * signing authority. therefore:
+ *   - only register first-party plugins or third-party plugins you would
+ *     trust with root on the host AND the master signing keys;
+ *   - pin + audit plugin dependencies exactly as you would the core itself;
+ *   - never register a plugin supplied at runtime by an untrusted party.
+ * a policy-gated signing facade in place of the raw vault handle is a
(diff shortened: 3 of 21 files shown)


lalalune:
Audit follow-up at exact head `63b4ecf39ce1ee07f5fffd820689000bd4566be4` (base `b63d2867f3265e7b8732c0bd6e4ba10f4de3c955`):

- Found and repaired a SEC-172 boundary gap: the MCP scrubber had a narrower, duplicated key vocabulary and could invoke getters while traversing an untrusted value.
- It now reuses `@stwd/shared`'s `isSensitiveCredentialKey`, retains MCP-specific `passwd`/signature handling, covers the omitted free-text carriers, and fails closed on accessors, cycles, and excessive depth.
- Added adversarial coverage for `clientSecretValue`, `cookieHeader`, `accessKeyId`, opaque JWT/signature text, getters, and cycles.

Local validation:
- `packages/mcp`: 65 passed / 0 failed; `tsc --noEmit`; Biome clean.
- `packages/eliza-plugin`: 52 passed / 24 skipped; `tsc --noEmit`; Biome clean.
- `packages/plugin-capabilities`: focused client-IP suite 7 passed / 0 failed; typecheck/Biome clean; full isolated package suite running separately.
- `packages/erc8183`: 7 passed / 0 failed; typecheck/Biome clean.
- `packages/erc8004`: 3 passed / 0 failed; typecheck/Biome clean.
- `packages/plugin-example`: 2 passed / 0 failed; typecheck/Biome clean.
- `git diff --check`: clean.

The prior 19/19 CI result was for superseded head `bc6cb4ce`; replacement exact-head run `32062411818` is in progress. No merge performed.

lalalune:
Final audit correction: exact head is now `58e37c16d2580bf1332b9e434033f0168bc96b1a`.

SEC-171's added `|| Boolean(config.proxyRequestSigningSecret)` term was tautological at the only check (`signingRequired && !config.proxyRequestSigningSecret`) and therefore did not enforce anything. The underlying `StewardProxyClient` already signs whenever a secret is supplied. I removed the misleading no-op, clarified the two distinct invariants in-code, and retained the regression test that observes an actual signature with explicit enforcement disabled (5/5 focused tests green).

The full workspace build completed 34/34 successfully. Any CI result for `63b4ecf3` is superseded; await replacement exact-head CI for `58e37c16` before merge.

PR #333, opened 2026-08-17
fix(security): wave 2b — deferred LOW/INFO remainder (batch misc)

lalalune:
# Security wave 2b — deferred LOW/INFO remainder (batch: misc)

Final wave-2 batch covering the deferred LOW and INFO findings assigned to the
"misc" batch. Wave-2 HIGH/MEDIUM fixes are already merged; this builds on them.

## Status

| SEC ID | Severity | Status | Summary |
|---|---|---|---|
| SEC-087 | LOW | **FIXED** | Production now REJECTS `sslmode=require` (postgres-js treats it as TLS without peer verification) and steers to `verify-full`/`verify-ca`; `STEWARD_ALLOW_UNVERIFIED_DB_TLS=true` is the explicit risk-acknowledgement escape hatch for legacy providers. (Escalated from the initial warn-only fix during PR review — commit 909c84d7.) |
| SEC-089 | LOW | **FIXED** | Audit metadata is normalized through the same `canonicalJsonValue` for both the HMAC preimage and the INSERT, so a stored row can always re-canonicalize to its written HMAC (previously `undefined`-valued keys broke chain verification from that seq onward). |
| SEC-090 | LOW | **FIXED** | PGLite data directory created `mode: 0o700`, and existing dirs are chmodded to owner-only. |
| SEC-097 | LOW | **FIXED** | Outbound header blocklist extended: `x-client-ip`, `cf-connecting-ip`, `true-client-ip`, `fastly-client-ip`, `x-cluster-client-ip`, `x-original-forwarded-for`. |
| SEC-098 | LOW | **FIXED** | `set-cookie` stripped from upstream responses on credential-injected routes so agents never receive credential-derived session tokens. |
| SEC-099 | LOW | **FIXED** | `x-steward-request-timestamp` / `x-steward-request-expires-at` stripped before forwarding. |
| SEC-100 | LOW | **FIXED** | Proxy limit env vars (`STEWARD_PROXY_RESPONSE_BYTES`, stream duration, in-flight caps) reject `0`/invalid values at startup with a clear error — limits can no longer be silently disabled; default response cap lowered to 25 MB. |
| SEC-166 | INFO | **FIXED** | CLOUDFLARE.md reconciled with reality: the audit chain *does* take `pg_advisory_xact_lock` on every non-PGLite append. On Workers, drizzle's neon-http driver throws `No transactions support in neon-http driver` from `db.transaction()`, so audited writes reject before lock or INSERT — they **fail closed**, never silently skip the chain. Pinned by a Workers-path test that deliberately breaks if a drizzle upgrade adds neon-http transactions (forcing re-review). |
| SEC-167 | INFO | **FIXED** | `isAuditSequenceConflict` no longer retries on the bare `23505` SQLSTATE or generic duplicate-key text — only serialization failures (`40001`) and conflicts naming `audit_events_tenant_seq_idx`. A caller's own unique violation inside the audited transaction now propagates immediately instead of 5 wasted retries. |
| SEC-168 | INFO | **FIXED** | `settleReservedSpend` decrements `reserved` through a Lua clamp (`max(0, current − release)`) — a double-settle can no longer drive it negative and silently free budget. Atomic with the read; `IoredisPipelineLike.eval` added (Upstash mapping covered by test). |
| SEC-169 | INFO | **PARTIAL / DECISION** | Tenant isolation remains application-layer only (no Postgres RLS). Full RLS enablement halfway is worse than none; the four preconditions (per-request `app.tenant_id` GUC discipline, `FORCE RLS` + break-glass role, cross-tenant-read policy matrix, PGLite/Workers parity) are now documented at the top of `schema.ts` as the rollout contract. No schema change. |
| SEC-174 | INFO | **FIXED** | Wildcard CORS removed — the proxy sends no CORS headers unless `STEWARD_PROXY_CORS_ORIGINS` lists explicit origins (agents are server-side callers). `/health` no longer discloses version or the route alias list; the API readiness probe only reads `serverTime` (verified). |
| SEC-175 | INFO | **FIXED** | Fail-closed by default in every environment: request signing, Redis rate limiting, and the shared replay store are REQUIRED unless `STEWARD_PROXY_DEV_MODE=true` explicitly opts into the soft dev posture (set by `docker-compose.dev.yml`, documented in `.env.example`). `NODE_ENV=production` overrides the dev flag, so a stray opt-in can never weaken production; `STEWARD_ALLOW_PROXY_REDIS_SOFT_FAIL` keeps its pre-existing explicit opt-out semantics. |
| SEC-176 | INFO | **FIXED** | Unknown `injectAs` now fails closed with a 400 config error instead of forwarding upstream with no credential. |
| SEC-191 | INFO | **FIXED** | JCS `serializeObject` reads members via `Object.getOwnPropertyDescriptor` and rejects accessor properties with `CANON_RUNTIME_VALUE_UNSUPPORTED` — getters are never invoked, matching the documented contract (prevents mint-vs-verify digest desync from value-changing getters). |
| SEC-192 | INFO | **FIXED** | `describeContributions()` builds its result with `Object.create(null)`; a plugin literally named `__proto__` no longer vanishes from JSON diagnostics. |
| SEC-194 | INFO | **FIXED** (docs) | Documented on `normalizeSensitiveKey`: non-ASCII lookalike key names (e.g. Cyrillic `secrеt`) evade the credential-carrier heuristic — accepted residual risk for a defense-in-depth canary, with the actual credential boundary (forbidden credential headers, vault-allowlisted injection) noted. |

## Trade-offs / decisions

- **SEC-087 (warn-not-reject):** `sslmode=require` is still accepted in production because hard rejection breaks existing deployments; the warning steers operators to `verify-full`. Cutover to rejection is a product decision, flagged in code.
- **SEC-175 (behavior change):** deployments running the proxy outside `NODE_ENV=production` *without* Redis/signing config will now fail closed at request time. This is the intended posture change from the audit; the explicit escape hatch is `STEWARD_PROXY_DEV_MODE=true` (dev compose updated). `NODE_ENV=production` always wins over the dev flag.
- **SEC-174 (CORS off by default):** any browser client calling the proxy cross-origin now needs `STEWARD_PROXY_CORS_ORIGINS`. No in-repo consumer does (agents and the API probe are server-side).
- **SEC-169 (documented, not implemented):** RLS requires a coherent migration (GUC discipline + FORCE + break-glass role + runtime parity); shipping half is worse than none. Documented as the rollout contract instead.

## How tested

- `packages/db`: **92/92 pass** (`bun test`), incl. new `audit-chain-retry.test.ts` (SEC-167: no retry on caller's own 23505; retries on seq-index race/40001; retry-then-commit) and `audit-chain-workers.test.ts` (SEC-166: neon-http transactions throw → audited writes fail closed). `tsc --noEmit` clean.
- `packages/redis`: **81/81 pass** with a live throwaway Redis (`STEWARD_REDIS_TESTS=1`, container removed after), incl. new double-settle floor-at-zero regression (SEC-168) and Upstash pipeline-`eval` arg-mapping test. Gated suites skip cleanly without a server. `tsc --noEmit` clean.
- `packages/shared`: **429/429 pass**, incl. new SEC-191 accessor-rejection test (getter never invoked) and SEC-192 `__proto__` plugin-name tests. `tsc --noEmit` clean.
- `packages/proxy`: suite **192 pass / 35 fail — identical to the pre-existing baseline on `origin/develop`** (same 35 failures, confirmed by stash-compare before/after; they are cross-file env-pollution artifacts — e.g. PR4/query-preservation/spend-limit files — plus `proxy-approval-lifecycle` / `proxy-approval-expiry-audit-atomicity`, which fail in isolation on develop too). New SEC-175 regression test added; the 4 strict-posture test files now set `STEWARD_PROXY_DEV_MODE` explicitly. `tsc --noEmit` clean.
- `packages/api`: `audit-chain-strict` (SEC-089) and `plugin-host` (SEC-192 consumer) pass; `tsc --noEmit` clean.
- `bunx biome check` on all 31 changed TS files: clean.

### Pre-existing failures (not caused by this PR)

The 35 proxy full-suite failures above pre-date this branch on `origin/develop`
(verified by running the suite with all branch changes stashed). Individual
files pass in isolation except `proxy-approval-lifecycle` and
`proxy-approval-expiry-audit-atomicity`, which fail on develop as-is.

## Post-review follow-ups on this branch

After the initial push, the automated review loop landed additional hardening
on top of the commits above (all assigned-ID statuses unchanged):

- `909c84d7` fix(db): require authenticated production TLS — SEC-087 escalated
  from warn to reject (row above updated).
- `b8c70e83` fix(security): complete proxy fail-closed hardening — extends
  SEC-175 with config-security test coverage and PGLite permission assertions.
- `60722d0e` fix(proxy): bound every resource-limit setting — extends the
  SEC-100 startup validation to the remaining proxy limit env vars.
- `749b8d58` / `f9ed3ec8` — test/CI posture wiring for the fail-closed
  defaults (explicit STEWARD_PROXY_DEV_MODE in CI jobs and fixtures).
- Branch merged with develop (`f6a0c614`), pulling in the other wave-2b batch
  PRs (#328 infra/scripts, #329 plugins, #330 vault).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -325,6 +325,11 @@ STEWARD_PROXY_PORT=8080
 # NODE_ENV=production overrides this flag. docker-compose.dev.yml sets it.
 # STEWARD_PROXY_DEV_MODE=true
 
+# Per-agent, per-upstream-host Redis rate limit. Both values must be positive
+# integers; invalid or zero values fail closed at startup.
+# STEWARD_PROXY_RATE_LIMIT_MAX=60
+# STEWARD_PROXY_RATE_LIMIT_WINDOW_MS=60000
+
 # Comma-separated browser origins allowed CORS access to the proxy. Since
 # SEC-174 the proxy sends NO CORS headers by default (agents are server-side
 # callers; a wildcard origin would let any website drive probes, and use a
--- a/.env.example
+++ b/.env.example
@@ -122,8 +122,13 @@ STEWARD_EXECUTION_AUTH_SECRET=
 # ─── Database ─────────────────────────────────────────────────────────────────
 
 # [REQUIRED for full/Postgres mode] PostgreSQL connection string.
-# Example: postgres://steward:password@localhost:5432/steward
+# Remote production example: postgres://steward:password@db.example:5432/steward?sslmode=verify-full
 DATABASE_URL=
+
+# Production rejects sslmode=require because postgres-js does not authenticate
+# the database peer in that mode. Prefer verify-full. If a legacy provider only
+# supports sslmode=require, this separate acknowledgement opts into the MITM risk.
+# STEWARD_ALLOW_UNVERIFIED_DB_TLS=false
 
 # [REQUIRED when using the bundled Postgres container in docker-compose.yml]
 # Password for the managed Postgres container. The production compose files
--- a/.env.example
+++ b/.env.example
@@ -330,6 +330,11 @@ STEWARD_PROXY_PORT=8080
 # NODE_ENV=production overrides this flag. docker-compose.dev.yml sets it.
 # STEWARD_PROXY_DEV_MODE=true
 
+# Per-agent, per-upstream-host Redis rate limit. Both values must be positive
+# integers; invalid or zero values fail closed at startup.
+# STEWARD_PROXY_RATE_LIMIT_MAX=60
+# STEWARD_PROXY_RATE_LIMIT_WINDOW_MS=60000
+
 # Comma-separated browser origins allowed CORS access to the proxy. Since
 # SEC-174 the proxy sends NO CORS headers by default (agents are server-side
 # callers; a wildcard origin would let any website drive probes, and use a
--- a/.env.example
+++ b/.env.example
@@ -318,6 +318,20 @@ REDIS_URL=
 # Port the proxy gateway listens on (default: 8080).
 STEWARD_PROXY_PORT=8080
 
+# Explicit opt-in to the soft DEVELOPMENT posture: unsigned agent requests,
+# permissive in-process rate limiting when Redis is down, and a per-process
+# replay store. Since SEC-175 the proxy fails CLOSED by default in every
+# environment — leave this UNSET anywhere real traffic flows. An explicit
+# NODE_ENV=production overrides this flag. docker-compose.dev.yml sets it.
+# STEWARD_PROXY_DEV_MODE=true
+
+# Comma-separated browser origins allowed CORS access to the proxy. Since
+# SEC-174 the proxy sends NO CORS headers by default (agents are server-side
+# callers; a wildcard origin would let any website drive probes, and use a
+# leaked token cross-origin from browser JS). Set only for genuine browser
+# clients.
+# STEWARD_PROXY_CORS_ORIGINS=https://dashboard.example.com
+
 # ─── Migrations ───────────────────────────────────────────────────────────────
 
 # Set to "true" to skip automatic migrations on API startup.
--- a/.env.example
+++ b/.env.example
@@ -323,6 +323,20 @@ REDIS_URL=
 # Port the proxy gateway listens on (default: 8080).
 STEWARD_PROXY_PORT=8080
 
+# Explicit opt-in to the soft DEVELOPMENT posture: unsigned agent requests,
+# permissive in-process rate limiting when Redis is down, and a per-process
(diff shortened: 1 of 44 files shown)


lalalune:
Independent audit note: SEC-169 remains documentation-only and is now tracked in #343. This PR must not be represented as implementing database RLS or closing #343.

PR #342, opened 2026-08-17
fix(vault): complete SEC-162 and SEC-164 boundaries

lalalune:
## Summary

Follow-up to #330 after independent security review.

- make `tenantId` mandatory for every in-process `Vault.createWallet` call so omitting it cannot preserve the cross-tenant venue-slot-squatting path from SEC-162
- update all production and test call sites to pass their authenticated/known tenant
- make legacy-secret re-encryption updates compare-and-set all authenticated ciphertext fields, so a stale migration read cannot overwrite a concurrent master-password rotation
- add a deterministic concurrency regression proving the competing ciphertext wins and the migration reports the row as failed

## Validation

- `bunx biome check` on all 10 changed files
- `bunx turbo build --filter=@stwd/vault --filter=@stwd/api` (21/21)
- SEC-082/162/163/164 Vault suites: 37 pass, 0 fail
- concurrent secret-rotation regression: 1 pass, 0 fail
- pre-follow-up #330 exact-head CI: all 19 required checks passed before merge

No related standalone issue was found; this is a correctness follow-up to #330.

### DIFF
--- a/packages/api/src/__tests__/vault-bitcoin-psbt-signing.test.ts
+++ b/packages/api/src/__tests__/vault-bitcoin-psbt-signing.test.ts
@@ -164,27 +164,32 @@ describe("vault Bitcoin PSBT signing", () => {
     await vault.createAgent(TENANT_ID, TAPROOT_AGENT_ID, "Bitcoin PSBT Taproot Agent");
     const wallet = await vault.createWallet({
       agentId: AGENT_ID,
+      tenantId: TENANT_ID,
       chainType: "bitcoin",
       bitcoin: { network: "testnet", addressType: "p2wpkh" },
     });
     const changeWallet = await vault.createWallet({
       agentId: AGENT_ID,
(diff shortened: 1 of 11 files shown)


lalalune:
Exact-head audit follow-up:

- The pre-fix head `733b0e73109cfc448e12ace75f766e5d793a9781` missed one direct Vault call in `packages/plugin-trading/src/__tests__/trade-polymarket-order.test.ts`; its plugin-trading CI job failed after `tenantId` became required.
- Repaired that call with the authenticated test tenant and guarded-pushed only after confirming the remote still matched the reviewed SHA.
- New exact head: `0ed166bf46f888253d5289797d8da7feef289b78`; exact base and merge-base: `a71b778420c838d438f419bc475863f1ab61a5dc`.
- Fresh exact-head workflow: https://github.com/Steward-Fi/steward/actions/runs/32070275382 (currently queued). Older green jobs are stale for merge purposes.

PR #349, opened 2026-08-17
fix(ci): minimize release workflow permissions

lalalune:
## Summary

- remove unused `packages: write` from the release job
- retain only `contents: write`, required for GitHub Release creation
- keep npm publishing on `NPM_TOKEN`; GHCR remains isolated to the Docker workflow

This preserves the least-privilege repair found while auditing #337 without taking the now-policy-blocked routine major action upgrade. Supersedes #337.

## Validation

- `actionlint .github/workflows/release.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,11 +9,10 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
-    # contents:write is required to create GitHub Releases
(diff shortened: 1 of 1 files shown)


lalalune:
Closing as superseded by #351. The only functional change here—removing release-job packages:write—is included verbatim in #351, which also carries the audited pinned release-action upgrade and the related Docker action upgrades on a newer base. The expanded comment is explanatory only and adds no unique behavior.

PR #352, opened 2026-08-17
fix(security): close post-merge #318 auth proxy gaps

lalalune:
## Post-merge audit of #318

This corrective closes the remaining browser auth-proxy, refresh-rotation, and CSP gaps found after #318.

### Repairs

- Refuse credential-bearing upstream redirects; bound refresh and revoke responses to 10 seconds and 1 MiB.
- Whitelist the exact refresh success/error response and compact-JWT shape; never reflect upstream errors or refresh-token fields into browser JavaScript.
- Persist access tokens only after the HttpOnly refresh-token deposit succeeds.
- Serialize refresh, tenant switch, revoke, and cookie deletion with one origin Web Lock. A shared localStorage logout epoch and storage events make sign-out win across tabs; proxy rotation fails closed when lock or shared epoch storage is unavailable.
- Validate tenant switch ids and atomically lock/recheck the active user and target membership rows before issuing the target-tenant JWT and rotated refresh token. Unauthorized, removed, or deactivated switches fail closed.
- Correct Coinbase Wallet CSP origins while keeping optional analytics blocked.

### Exact-head verification

- Head: 67411ce30dd793bac415a31b8c476d7eb84e6dde
- Base: 7779c556d400e985a50ba95ad731d70ceb0cfb48
- Focused SDK, API-source, auth-route, proxy, and CSP regressions: 72 passed, 0 failed; authority-lock regression: 4 passed, 0 failed.
- Dependency-aware targeted build: 21 of 21 packages successful.
- Biome on all affected source and test files plus git diff --check: clean.
- Deterministic two-instance race proves an in-flight refresh cannot resurrect the access token or rotated cookie after another tab signs out; provider mutations remain ordered under the shared lock.

Exact-head CI is required before merge.


lalalune:
Re-audit repair at exact head `80d4714e9cb1b446552f3ce63055283d5a916826` on current `develop` `47e4856de2886ffa53946e224916e8dd6e455990`.

The prior implementation serialized refresh/revoke/cookie deletion, but the initial HttpOnly refresh-token deposit remained outside that origin mutation lock. A `signOut()` racing a successful deposit could therefore clear the old state and then have the delayed sign-in persist both the new cookie and access token, resurrecting the session. The repair serializes deposits under the same Web Lock and checks the instance generation plus shared logout epoch before and after custody transfer; stale deposits delete the cookie while still holding the lock and never persist the access token. A deterministic regression also proves two client instances serialize proxy refreshes through the origin lock.

Validation on the rebased tree:
- focused SDK/API/web/CSP/auth-proxy suites: 74 passed, 0 failed
- SDK TypeScript build: green
- API source/auth authority regressions: green

The broader web dependency build is left to exact-head CI because the isolated worktree dependency links were incomplete under the current runner contention. Keep draft until that exact CI is green.

PR #353, opened 2026-08-17
fix(security): make trading idempotency atomic and harden venue execution

lalalune:
## Corrective for merged #347

#347 merged before its concurrency and fill-accounting audit completed. This PR consolidates the correct, non-duplicate hardening from #353/#354/#357 on exact develop `0c6b96fd`, while replacing the unsafe late-claim and integer-response heuristics.

### Correctness and security
- atomic two-phase idempotency claim (`SET PX NX`) with token-bound Lua completion/release across replicas
- pending claims fail closed after crashes or completion-write failures; deterministic validation/policy outcomes remain replayable
- atomic coverage at all external mutation boundaries, including operator recovery routes
- opaque collision-safe Redis/memory keys; production without Redis fails closed unless explicitly acknowledged single-instance
- Polymarket `makingAmount`/`takingAmount` interpreted in the API response human units, including exact integer strings; no magnitude/order-size guessing
- AES-GCM cached CLOB credentials bound to tenant/agent/wallet/endpoint cache key via AAD
- strict Polymarket builder config: complete enablement, non-empty auth token, HTTPS signer transport (loopback HTTP only), no URL credentials
- Hyperliquid builder config rejects partial address/fee configuration
- current-develop vault fixtures supply tenant binding and fake Redis implements NX/CAS, fixing SEC-108 failures seen on current branches

### Exact-head validation
Head: `233dd879ae35472b131b33336a68f54d2a7b11a8`
Base: `0c6b96fda905871359d1e334d7d1416b46efde58`

- focused combined run: 139 passed; one concurrent PGLite migration hook timed out
- isolated rerun of that suite with repository-appropriate timeout: 16/16 passed, including concurrent close-all (one external call)
- encrypted Polymarket credentials/AAD transplant suite: 2/2 passed, 21 assertions
- Hyperliquid + Polymarket venue unit coverage passed, including direct decimal and >=1M-share response-unit cases
- `turbo run build` for plugin-trading, venue-hyperliquid, venue-polymarket, agent-trader: 13/13 tasks passed
- Biome: 7/7 changed source/test files clean; `git diff --check` clean

### Supersession
- #354 is fully contained here (tenant-bound SEC-108 fixtures plus faithful NX/CAS Redis double).
- #357 overlap and valid cache-key AAD hardening are contained here. Its integer-string scaling heuristic is intentionally excluded: an exact integer human-unit fill is valid and must not be divided by 1e6.

PR #354, opened 2026-08-17
test(trading): restore tenant-bound Polymarket cache setup

lalalune:
## Summary

- pass the seeded tenant ID into both SEC-108 Polymarket cache wallet fixtures
- restore compatibility with the tenant-bound vault API landed in #342

## Failure fixed

Current exact `develop` integration failed both `trade-polymarket-creds-cache` cases with `Agent ... not found for tenant undefined` in PR Checks job `95547162882`.

## Validation

- `biome check` passes for the changed test
- `git diff --check` passes
- exact-head PR matrix is authoritative for the full package run


lalalune:
Closing as superseded by #353. The tenant-bound Polymarket cache fixture repair from this test-only branch is included in #353 together with the production atomic idempotency, credential-cache, signer-transport, fill-accounting, and Hyperliquid configuration fixes.

PR #357, opened 2026-08-18
fix(security): close PR #347 idempotency and fill-accounting gaps

lalalune:
## Security follow-up to #347

PR #347 merged while its security audit was still in progress. This follow-up fixes release-blocking defects in the merged SEC-043/SEC-108/SEC-187 implementation.

- replace check-then-execute-then-store with an atomic Redis `SET NX PX` processing claim at the last safe pre-execution boundary
- suppress concurrent, cross-replica, restart, crash, and completion-write-failure retries for 24h instead of permitting a duplicate fund movement
- fail closed on malformed durable records and use an owner-checked Lua completion CAS
- mirror atomic processing claims in the bounded single-process fallback
- bind encrypted Polymarket L2 credential cache blobs to tenant/agent/wallet/endpoint via AES-GCM AAD, preventing ciphertext transplantation between cache slots
- follow Polymarket POST /order wire semantics: integer amount strings are 6-decimal fixed math; decimal strings are the human-unit variant, avoiding both tiny-fill and million-order misaccounting

## Validation

- repository dependency build: 34/34 tasks green, including plugin-trading and both venue packages
- `bunx tsc -p packages/plugin-trading/tsconfig.json --noEmit`
- `bunx tsc -p packages/venue-polymarket/tsconfig.json --noEmit`
- idempotency store: 9/9
- venue-polymarket: 59/59
- encrypted credential cache: 2/2
- Polymarket order lifecycle: 28/28
- operator recovery suites completed green before the final narrow rebase
- Biome and `git diff --check` clean

The original #347 head documented concurrent duplicate execution and post-execution store loss as accepted residuals; those are not acceptable for fund-moving routes and are eliminated here.

### DIFF
--- a/packages/plugin-trading/src/__tests__/idempotency-store.test.ts
+++ b/packages/plugin-trading/src/__tests__/idempotency-store.test.ts
@@ -2,19 +2,12 @@ import { describe, expect, it } from "bun:test";
 import type { IoredisLike } from "@stwd/redis";
 import { DurableIdempotencyStore } from "../routes/idempotency";
 
-/**
- * SEC-043: trade/operator idempotency records must be DURABLE — backed by
- * Redis when a client is available so a restart or a second replica still
- * dedups a retried withdraw/order — with the bounded in-memory map as the
- * dev/single-replica fallback.
- */
-
 type TestRecord = { status: number; body: unknown };
 
-/** Minimal in-memory IoredisLike double (records raw set args for assertions). */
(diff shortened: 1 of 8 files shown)


lalalune:
Closing as superseded by consolidated #353. The valid AES-GCM AAD and opaque idempotency-key hardening from this branch were carried into #353, while its unsafe integer-string fill heuristic was replaced with direct human response-unit handling and its remaining atomic lifecycle gaps were repaired.

PR #358, opened 2026-08-18
fix(auth): centralize public OIDC destination checks

lalalune:
## Summary

- replace duplicated JWKS and authorization-code token endpoint classifiers with one fail-closed public Internet destination classifier
- reject noncanonical IPv4 text, IPv4-compatible/translated IPv6, non-global IPv6, IANA special-purpose IPv4 blocks, credential-bearing URLs, scoped IPv6, single-label/search-domain hosts, and local/special-use DNS suffixes
- deny every blocked DNS suffix at both its apex and subdomains after trailing-dot/case normalization
- apply the same classifier to issuer, JWKS, authorization, and token URLs at configuration time
- revalidate legacy authorization URLs immediately before issuing the browser redirect
- validate URL literals and every connect-time DNS candidate on JWKS and token transports, failing closed where connect-time pinning is unavailable

## Security impact

This corrects the post-merge review finding in #332. The prior duplicated classifiers admitted IPv4-compatible IPv6 and special-purpose ranges, while suffix-only checks missed apexes such as home.arpa. The exact-head follow-up also closes ambiguous resolver address forms: empty octets, signs, exponent notation, whitespace, and leading-zero octets are rejected rather than coerced by Number(). Mixed Happy Eyeballs results fail when any candidate is non-public.

The application redirect remains separately exact-allowlisted and receives only a one-time exchange code. OIDC state, provider, tenant, PKCE, and nonce bindings are unchanged. Provider configuration remains owner/admin or platform-only; tenant API keys are rejected.

## Exact-head validation

Base: `47e4856de2886ffa53946e224916e8dd6e455990`
Head: `4dca457438df5567d2499b81ef80ea76fafd2b36`

- focused public endpoint, JWKS, provider-config, OIDC JWT, and authorization-code suites: 39 passed, 262 assertions
- filtered auth/API build graph: 21/21 tasks
- strict malformed IPv4 and mixed public/private DNS candidate regressions pass
- Biome and `git diff --check`: passed
- exact-head GitHub CI and required review remain required before merge

Refs #332


lalalune:
Deep-review repair pushed at exact head `6ec80232a61fc4194777105bbe994845e074791e`. The original transport was not connect-compatible with modern Node: HTTPS requests request `lookup({ all: true })`, but the hook forced a scalar result, producing `ERR_INVALID_IP_ADDRESS`. Both JWKS and token transports now share an all-aware lookup hook that validates every Happy-Eyeballs candidate immediately before connection. Cloudflare Workers explicitly fail closed because Workers does not implement `dns.lookup` and ignores the HTTPS `lookup` option; silently using that path would discard the claimed connect-time SSRF boundary. The portable URL/IP classifier no longer imports `node:net`, and the omitted IANA `2620:4f:8000::/48` special-purpose prefix is blocked. URL credentials/non-HTTPS/local names/special literals remain denied; redirects remain rejected; custom public HTTPS ports remain supported. Focused auth/API tests: 22 pass / 180 assertions; auth and API TypeScript checks pass; live public/private connector probes passed.

PR #308, opened 2026-08-17
feat(capabilities): issue upstream-derived credential leases

lalalune:
Refs #150. Implements the upstream lease-mode code portion; the tracker remains open for live Railway, GitHub App, and zero-raw-credential acceptance evidence.

## Summary

Replaces the misleading GitHub `token` mode with real GitHub App installation-token leases and a fail-closed delivery lifecycle:

- binds issuance to tenant, workspace, authenticated agent, live grant/capability, issuer configuration digest, repository scope, and permission subset
- validates exact GitHub identifier, permission, UTF-8 token-byte, and one-hour lifetime contracts with bounded no-redirect transport
- durably claims idempotency before provider I/O; concurrent or replayed requests cannot mint or re-deliver credentials
- encrypts the exact provider token with tenant+lease AAD before any contract rejection, enabling exact cleanup and recovery even for malformed successful responses
- delivers once as `delivery_pending`; the holder must prove possession through ACK before activation
- autonomously revokes abandoned, interrupted, expired, or authority-invalid leases through globally deadline-ordered bounded recovery
- commits authority tombstones before exact-token teardown; per-lease failures do not strand siblings
- preserves truthful `needs_attention` state and the sealed recovery handle after ambiguous provider outcomes
- atomically couples terminal/private lifecycle events with tenant-bound core append-only audit evidence
- prevents provider tokens from outliving grants, and fails issuance closed in Workers or server configurations unable to honor the recovery deadline
- exposes bounded issue, ACK, and revoke contracts through the agent SDK

Migration `0093_upstream_credential_leases` follows merged budget migration 0092 contiguously.

## Validation

Exact head: `f8b60c99ccae179dee9b6d3fdedb678d30a89aac`
Exact base: `47e4856de2886ffa53946e224916e8dd6e455990`

- frozen install: unchanged
- forced dependency-aware API/plugin build: 21/21 tasks, zero cache hits
- exact full upstream lease suite: 38/38 tests, 228 assertions
- exact manifest routing, GitHub issuer, scheduler, Worker, and migration suites: 23/23 tests, 55 assertions
- SDK base transport + agent client from the patch-identical predecessor: 33/33 tests, 86 assertions
- Biome and `git diff --check`: clean
- range-diff preserves all 28 feature/hardening commits; the rebased tree is byte-identical to the locally validated candidate
- migration chain applies 0091, 0092, then 0093 in PGLite
[...]

PR #361, opened 2026-08-18
fix(capabilities): broker GitHub until native token mint

lalalune:
## Summary

Post-merge audit follow-up for #319.

- classify GitHub capability issuance as broker mode until Steward implements a real GitHub App installation-token exchange
- never return an internal Steward session JWT as though it were a provider-native GitHub credential
- update provider-mode documentation and route/core regression coverage

## Security rationale

GitHub App installation tokens can be scoped and short-lived, but the current issuance callback uses `signAgentToken`, producing a Steward HS256 JWT that GitHub cannot consume. Broker mode is the fail-closed honest behavior until a provider-native minter exists.

## Validation

- `bunx turbo run build --filter=@stwd/plugin-capabilities` (9/9)
- focused provider-mode, issuance, manifest-route, and rate-limit tests
- Biome and `git diff --check` clean

Follow-up to #319.


lalalune:
Closing as superseded by #308, which has now merged the production GitHub-native installation-token lease flow, including bounded issuance, delivery acknowledgement/revocation, recovery, evidence integrity, and exact tenant/agent/workspace/grant revalidation. This broker-only stopgap is no longer the implementation path and should not be merged alongside the landed lease architecture.

PR #365, opened 2026-08-18
fix(ci): pin PR metadata diff to event SHAs

lalalune:
## Summary

- retain merge-base only for changed-file attribution
- read package versions from the immutable pull_request event base SHA and head SHA
- accept package.json metadata only when the head version is a strict SemVer increase
- reject unchanged versions, downgrades, malformed versions, and numeric prerelease identifiers with leading zeroes

Exact base: 3ca87c1cef516b93051304e07070988c91a4d1aa
Exact head: 75bc7813e9ae915fc14059d4b3a2abf3487fd2a6

## Validation

- metadata checker: 5 passed, 0 failed, 15 assertions
- actionlint: passed

The PR remains draft pending exact-head CI.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -65,21 +65,13 @@ jobs:
           # Compare the immutable commits captured by the pull_request event.
           # The base branch can advance while this job runs; diffing its live
           # remote tip would then attribute unrelated merged changes to the PR.
-          git fetch --no-tags origin "${BASE_SHA}" "${HEAD_SHA}"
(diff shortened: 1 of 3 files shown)


lalalune:
Adversarial follow-up at exact `4293f3e5`: the event-SHA diff now uses the two commits already present in the full pull-request checkout and verifies both with `git cat-file`. It no longer asks the base repository to fetch a fork-only head SHA. `actionlint`, an exact event-style three-dot diff reproduction, and `git diff --check` pass. Fresh exact-head CI is authoritative.

PR #367, opened 2026-08-18
test(security): align fixtures with fail-closed contracts

lalalune:
## Summary
- configure an explicit JWKS URL in mocked agent-token suites now that the implicit development trust anchor fails closed
- give the PGlite-backed token expiry suite enough setup time
- assert the sanitized recovery conflict response instead of the raw vault exception

## Validation
- EVM swap prepare: 14/14 passing
- agent token expiry monitoring: 5/5 passing
- user wallet recovery setup: 17/17 passing
- `bunx turbo run build --filter=@stwd/api...`

### DIFF
--- a/packages/api/src/__tests__/user-wallet-recovery-setup.test.ts
+++ b/packages/api/src/__tests__/user-wallet-recovery-setup.test.ts
@@ -898,6 +898,7 @@ describe("user wallet recovery setup", () => {
     expect(response.headers.get("Cache-Control")).toContain("no-store");
     const text = await response.text();
     expect(text).not.toContain(mnemonic);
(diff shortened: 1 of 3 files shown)


lalalune:
Superseded by #368, which carries the same security-fixture corrections plus environment restoration and exact isolated verification.

lalalune:
Closed as a duplicate of #368 after exact patch comparison. #368 retains the required explicit default-JWKS opt-in for suites that deliberately mock the built-in development JWKS, restores the caller's environment after each suite, and now incorporates this PR's stronger exact sanitized JSON assertion and 30-second agent-suite timeout. No production behavior differs.

PR #377, opened 2026-08-18
fix(test): isolate approval principal RPC seams

lalalune:
## Summary
- stub the native-recipient code lookup used by the approval-principal route test
- stub transaction signing because this suite verifies attribution and separation of duties, not signer implementation
- remove dependence on public Base RPC availability from the Postgres integration matrix

## Verification
- local PostgreSQL 16, exact focused test: 3 passed, 0 failed
- 21-package API dependency build: passed
- Biome and git diff checks: passed

Fixes the sole failure in develop CI run 32091754732 at exact SHA 47e4856de2886ffa53946e224916e8dd6e455990.

### DIFF
--- a/packages/api/src/__tests__/approval-principals.test.ts
+++ b/packages/api/src/__tests__/approval-principals.test.ts
@@ -1,4 +1,4 @@
-import { afterAll, beforeAll, describe, expect, it } from "bun:test";
+import { afterAll, beforeAll, describe, expect, it, spyOn } from "bun:test";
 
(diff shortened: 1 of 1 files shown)


lalalune:
Closing as superseded by #378, which already landed the approval-principal isolation fix on develop in 6168e740. #378 stubs the same EOA RPC seam but correctly intercepts GovernedVault.signTransactionAuthorized (the actual approval path) and asserts exactly one authorized signing call; this PR instead stubs Vault.signTransaction and provides no additional behavior or coverage.

PR #379, opened 2026-08-18
test(api): isolate approval principal RPC fixture

lalalune:
## Summary
- replace the approval-principal fixture dependency on a public Base Sepolia RPC with strict in-process vault seams
- keep execution/auth/audit secrets and mutable vault methods isolated and restored
- use collision-resistant wallet identities and a standalone-safe timeout
- assert only the distinct approver reaches signing

## Regression evidence
- reproduced develop CI run 32091754732 failure locally against PostgreSQL 16
- focused real-Postgres suite: 10 pass, 0 fail, 77 expects
- @stwd/api dependency build: 21/21
- Biome and git diff checks clean

The production route is unchanged; this repairs deterministic coverage for the existing four-eyes authorization contract.


lalalune:
Closing as superseded by merged #378 (6168e740), which already fixes the exact develop/Postgres failure by stubbing the EOA lookup and the authoritative GovernedVault.signTransactionAuthorized approval boundary, with an explicit one-call assertion. This branch reimplements the same fixture isolation from the pre-#378 base, conflicts with the landed test, and moves the signing seam down to mutable singleton Vault methods; its timeout/environment/address fixture changes do not cover a distinct product behavior or unresolved failure. The related alternative #377 was closed for the same reason.

lalalune:
Closing again: the production-code fixture correction already landed via #378. This revision still replaces the authoritative GovernedVault.signTransactionAuthorized seam with mutable singleton Vault.signTransaction, which weakens what the four-eyes test proves and can bypass the governed authorization boundary under test. The extra timeout/env cleanup and randomized identities are test hygiene, but they do not constitute a distinct unresolved product behavior and do not justify regressing the signing seam. If any hygiene remains desirable, it should be proposed separately atop current develop while retaining #378's governed seam.

PR #386, opened 2026-08-18
fix(api): suppress trailing credential replay paths

lalalune:
Follow-up to #382 after exact review of the credential response replay boundary.

The secret-response classifier covered issue and renew routes without a trailing slash, while the capability route exposure contract accepts an optional trailing slash. This patch keeps those equivalent paths out of the idempotency response store too.

Regression coverage verifies both unversioned issue and versioned renew trailing-slash forms return the first credential once, then reject the duplicate with 409 instead of persisting or replaying its body.

Intentionally draft pending exact-head CI and maintainer review.

### DIFF
--- a/packages/api/src/__tests__/idempotency-middleware.test.ts
+++ b/packages/api/src/__tests__/idempotency-middleware.test.ts
@@ -366,6 +366,14 @@ describe("idempotencyMiddleware", () => {
       count += 1;
       return c.json({ ok: true, data: { token: `ghs_v1_renew_${count}` } });
     });
+    app.post("/capabilities/manifest/github/issue/", async (c) => {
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after independent exact-head audit: production Hono routing is strict and registers only the non-trailing-slash issue/renew paths. Real router probes return 404 for both unversioned and /v1 trailing-slash variants; encoded/doubled-slash variants do not bypass routing, and normalized dot paths are already covered by the exact existing classifier. The PR tests manufactured slash routes that are not part of the production exposure contract, so the stated vulnerability premise is false. The regex change is harmless but unnecessary and would leave misleading security coverage.

PR #396, opened 2026-08-18
test(provider): isolate governed rate-limit fixture

lalalune:
## Summary

- make the provider-authority E2E explicitly permit its deterministic fake dispatch through the production rate-limit seam
- restore the real Redis enforcement function after the suite
- keep production fail-closed behavior unchanged

## Why

The suite claims to exercise the governed terminal forwarder, but the newer fail-closed Redis rate limiter returns 429 before the fake transport when Redis is intentionally absent. This made all positive dispatch assertions fail with zero forwards and hid the boundary the test is supposed to prove.

## Exact validation

Head `aee0855bcd52439e2f68026d327099515187c80f`, base `8d7da9e26bc1020a0d89c20955d3f3cab7d10f94`:

- isolated provider-authority E2E: 12/12 passed
- API dependency build: 21/21 tasks passed
- Biome and `git diff --check`: passed

### DIFF
--- a/packages/api/src/__tests__/provider-authority-e2e.test.ts
+++ b/packages/api/src/__tests__/provider-authority-e2e.test.ts
@@ -67,6 +67,7 @@ type ProxyMod = typeof import("@stwd/proxy/src/handlers/proxy");
 type DispatchMod = typeof import("@stwd/proxy/src/handlers/governed-execution");
 let proxyMod: ProxyMod;
 let dispatchGovernedExecution: DispatchMod["dispatchGovernedExecution"];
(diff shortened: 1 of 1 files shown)


lalalune:
Independent exact-head audit at `70f8150d0b51a7b9b909e3271ce627240763dd19` on `a82d7914edcbc4d3809f94aa4e371d10fc631a15`: fixture-only change, deterministic allow is installed through the existing test seam and the exact production Redis checker is restored in `afterAll`; no production behavior or assertions are weakened. Isolated E2E, 23-package dependency build, Biome, and diff-check are green.

PR #400, opened 2026-08-18
fix(proxy): preserve terminal summon replay semantics

lalalune:
## Summary

Follow-up to merged #369 and #394:

- return persisted terminal and `outcome_unknown` dispositions before revalidating short-lived X summon evidence
- preserve final-boundary summon verification for every fresh, dispatchable execution
- ensure terminal replay never reaches the atomic claim, credential decryption, or upstream forwarding path
- retain #394’s truly-absent-provenance compatibility behavior

## Security rationale

Summon attestation expiry or adapter-key rotation after a completed dispatch must not rewrite an immutable idempotent replay into `EXEC_AUTH_SUMMON_ATTESTATION_INVALID`. The persisted dispatch disposition is authoritative once the nonce is consumed or dispatch state has advanced. Fresh executions still revalidate exact audience, tenant, agent, account, source post, action scope, idempotency hash, signature, expiry, and request-envelope digest before claim/decrypt/fetch.

## Validation

Exact base: `8d7da9e26bc1020a0d89c20955d3f3cab7d10f94`
Exact head: `66f997093bd3ab919032916421ca8affca607ac7`

- focused governed execution + X summon suites: 49 passed, 189 assertions
- `@stwd/proxy` TypeScript build: passed
- Biome on all relevant files: passed
- `git diff --check`: passed

Regression matrix covers both `succeeded` and `outcome_unknown` replays after both attestation expiry and signing-key removal, asserting zero upstream dispatch.

### DIFF
--- a/packages/proxy/src/__tests__/governed-execution.test.ts
+++ b/packages/proxy/src/__tests__/governed-execution.test.ts
@@ -1262,6 +1262,39 @@ describe("PR4 dispatchGovernedExecution claim + dispatch", () => {
     expect(captured).toBeNull();
   });
 
+  it("P26: invalid summon evidence on a fresh execution is denied before claim", async () => {
+    const summonDigest = `sha256:${"8".repeat(64)}`;
+    const { intentId, authorizationId } = await seedExecutionReady({
+      requestEnvelope: {
+        schemaVersion: "steward.provider-request.v1",
+        xSummonAttestationDigest: summonDigest,
+      },
+      safeSummary: {
+        xSummonAttestation: {
+          schemaVersion: "steward.x-summon-attestation.v1",
+          keyId: "adapter-removed",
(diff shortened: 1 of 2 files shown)


lalalune:
Final exact-current-base evidence at `7494dcc8670b25f1e24ea88b0d07c60234d08b94` on `dc4279efcc8b8db8323b9ca7a497c9339fbc1fc1`: terminal succeeded/outcome_unknown results remain stable across attestation expiry/key removal; fresh invalid evidence still denies before claim/decrypt/fetch. The integration was updated to preserve the newly landed bounded-response invariant: an oversized credential-bearing open stream fails closed and is cancelled, releasing the slot. Governed suite 47/47 (186 assertions), proxy graph build 9/9, Biome/diff green.

PR #404, opened 2026-08-18
fix(proxy): fail closed on Slack response errors

lalalune:
## Security finding and repair

Post-merge review of #302 found three residual response-boundary failures:

- an interrupted Slack response left a disturbed stream for later credential scanning
- Slack `ok:true` was audited before credential-reflection checks
- Slack semantic parsing and the generic credential scanner used `arrayBuffer()` before checking size, so absent or false `Content-Length` could allocate beyond the intended inspection bound

This draft preserves the disturbed-stream and audit-order fixes, and adds one independent fixed 2 MiB bounded reader for every inspectable credential-bearing response. It counts actual streamed bytes, validates declared length, cancels on overflow/error, discards partial bytes, and exposes only fixed errors. The same resulting bytes are reused by Slack semantics, LLM spend parsing, credential reflection scanning, and the downstream response—no second read and no contradictory Slack success audit.

Hostile regressions cover interrupted bodies, missing `Content-Length`, lying-small `Content-Length`, chunk overflow/cancellation, split credential reflection, audit ordering, and slot release.

Exact validated head: `56fd28a7e91d96eb3d43816b0855e62a0d7a459f`
Base: `93322dd1a809cdcc298be3a15ec450b9f16d66e7`

- Slack credential routes: 8/8 (39 assertions)
- generic proxy spend/SSRF/credential suite: 32/32 (174 assertions)
- proxy TypeScript build, Biome, and diff check: clean

Fresh exact-head CI is required before merge; keep draft until green.

### DIFF
--- a/packages/proxy/src/__tests__/slack-credential-route.test.ts
+++ b/packages/proxy/src/__tests__/slack-credential-route.test.ts
@@ -126,6 +126,65 @@ describe("Slack narrow credential route", () => {
     expect(audit.at(-1)?.reason).toBe("slack-api-error:channel_not_found");
   });
 
+  test("fails closed on an interrupted Slack body and releases the proxy slot", async () => {
+    const { tenantId, agentId } = await fixture();
+    proxyMod.__setProxyInFlightCapsForTests({ perAgent: 1, perTenant: 1 });
+    try {
+      proxyMod.__setForwardProxyRequestForTests(
+        async () =>
+          new Response(
+            new ReadableStream<Uint8Array>({
+              start(controller) {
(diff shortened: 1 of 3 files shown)


lalalune:
Final exact-current-base validation at `ca9d2fa64d139db1ce9928a7f0593486cea22f1b` on `57a42be52f2fa16e254d6949c373f64f37ef4e5c`: bounded 2 MiB actual-byte reader is reused across Slack semantics, LLM spend parsing, generic credential scanning, and downstream response; missing/lying Content-Length overflow cancels and fails closed; disturbed streams release slots; Slack success audit remains after credential-reflection rejection. Exact proxy graph build 9/9; Slack + generic hostile suites 40/40, 221 assertions; TypeScript, Biome, and diff-check green.

PR #407, opened 2026-08-18
test(proxy): isolate query preservation fixture

lalalune:
## Finding

The query-preservation integration suite relied on an ambient `STEWARD_PROXY_DEV_MODE=true`. From a clean environment all 11 cases stopped at the production fail-closed proxy gate with 401, so the test never exercised query preservation.

## Fix

Opt into the explicit local-test proxy posture before importing the handler and restore the caller's original environment afterward. Production defaults and assertions remain unchanged.

## Validation

- clean-env query-preservation suite: 11 pass, 65 expectations
- proxy dependency build graph: 9/9 pass
- Biome and git diff check: pass

### DIFF
--- a/packages/proxy/src/__tests__/proxy-query-preservation.test.ts
+++ b/packages/proxy/src/__tests__/proxy-query-preservation.test.ts
@@ -88,6 +88,10 @@ beforeAll(async () => {
 });
 
 afterAll(async () => {
+  // This file replaces DNS and the final network forwarder. Restore the real
+  // production functions so a shared-process run cannot leak its permissive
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head audit at `fe4dbc3ffb8e0fef2217d4d5ccbef04a762ae636` on `4186de53817717c6eae4f7568a8923c99700bebb`: the fixture snapshots/restores dev mode and now resets every mutable proxy handler seam back to the exact production rate/spend/DNS/forward functions. Query preservation assertions are unchanged; isolated suite, dependency build, Biome, and diff-check are green.

PR #414, opened 2026-08-18
fix(security): W3-D re-audit batch — policy/trading fail-closed residuals

lalalune:
Security re-audit residuals for plugin-trading.

- Durable idempotency fallback fails closed instead of evicting live or in-flight records when bounded local storage is saturated.
- Operator recovery rate-limit fallback likewise fails closed when all bounded windows are live.
- Redis-backed production behavior and existing durable reservation semantics remain unchanged.

Exact head: 37c988cb0012fc308d67964d950afbe11264eabb
Included develop: a7b1d566b6e09136ac7f6e723fa5e053203b3907

- focused idempotency and operator policy suites: 31/31, 86 assertions
- plugin-trading dependency build: 12/12
- Biome and diff-check: green
- independent semantic review: accept

Fresh hosted CI and required review remain merge gates.

### DIFF
--- a/packages/plugin-trading/src/__tests__/idempotency-store.test.ts
+++ b/packages/plugin-trading/src/__tests__/idempotency-store.test.ts
@@ -172,6 +172,52 @@ describe("DurableIdempotencyStore (SEC-043)", () => {
     expect(await store.check("scope", undefined, "hash-1")).toEqual({});
   });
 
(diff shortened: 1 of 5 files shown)


lalalune:
Temporarily restored Draft while the exact-current-head adversarial policy/trading audit is still running. Do not merge until that audit and fresh exact-head CI complete.

PR #415, opened 2026-08-18
test(proxy): align stream fixture with response inspection

lalalune:
## Summary
- update the governed stream regression for the fail-closed credential-response inspection merged in #404
- assert an oversized credential-bearing stream returns the stable 502 dispatch failure
- preserve the original concurrency-slot safety proof by asserting the hostile stream is cancelled

## Validation
- `STEWARD_AUDIT_HMAC_KEY=<test> bun test packages/proxy/src/__tests__/governed-execution.test.ts --test-name-pattern P36` — 1/1
- `bunx turbo run build --filter=@stwd/proxy...` — 9/9
- Biome and `git diff --check` green

Follow-up to #404.

### DIFF
--- a/packages/proxy/src/__tests__/governed-execution.test.ts
+++ b/packages/proxy/src/__tests__/governed-execution.test.ts
@@ -1583,17 +1583,22 @@ describe("PR4 dispatchGovernedExecution claim + dispatch", () => {
     expect(captured).toBeNull();
   });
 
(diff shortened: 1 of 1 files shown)


lalalune:
Closed as redundant: merged #393 independently updated the same P36 regression to the correct fail-closed oversized credential-response semantics while this follow-up was in flight. Current develop already asserts the 502 failure and hostile-stream cancellation, so retaining this PR would duplicate the test change.

PR #422, opened 2026-08-18
fix(security): preserve ambiguous cross-rail counters

lalalune:
Post-merge security follow-up to #401.

Two cross-rail accounting gaps remained:

- operator admission ignored core transactions in `outcome_unknown`, allowing a possibly-landed on-chain transfer to disappear from operator USD-spend and rate counters;
- core vault/intent admission counted only core transactions for rate-limit rules, so pending/final operator transfers did not consume the same hourly/daily policy budget.

This keeps ambiguous core outcomes charged until reconciliation, and includes active operator reservations in core transaction-rate counters. Released reservations remain excluded. USD-micro spend stays separate from chain-native units.

Exact head: `aa94690160fa1d302c695e92c7077a41d7ba96aa`
Included develop: `d8293bc4733f6d11d0235758f69c3cccfc02e646`

Fresh exact-head evidence:

- operator policy gates: 18/18, 61 assertions (60-second timeout for the 1,000-entry saturation case)
- transaction stats: 8/8, 18 assertions
- advisory-lock structural suite: 1/1; 2 real-PostgreSQL cases skipped because no `DATABASE_URL` is available locally
- API/plugin dependency build: 23/23
- Biome and diff-check: green

Kept draft for fresh hosted CI and real-PostgreSQL validation.

### DIFF
--- a/packages/api/src/__tests__/transaction-stats-chain-scope.test.ts
+++ b/packages/api/src/__tests__/transaction-stats-chain-scope.test.ts
@@ -150,8 +150,9 @@ describe("getTransactionStats chain scoping (issue #110)", () => {
     const combined = (BigInt(ETH_SPEND) + BigInt(SOL_SPEND)).toString();
     expect(stats.spentToday.toString()).toBe(combined);
     expect(stats.spentThisWeek.toString()).toBe(combined);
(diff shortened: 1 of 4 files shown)


lalalune:
Returned to draft only to refresh onto current develop after #429 landed; the exact cross-rail accounting patch itself is locally green.

PR #447, opened 2026-08-18
fix(security): enforce public webhook address parity

lalalune:
## Summary
- reject non-global and omitted special-purpose IPv4/IPv6 webhook destinations at registration and delivery
- correct the post-#445 assumption that adjacent special-use space is public
- fail closed on malformed or DNS-family-confused connection-time answers

## Validation
- 44 focused webhook/API tests (254 assertions)
- Biome check on all five changed files
- @stwd/webhooks TypeScript build
- standalone strict TypeScript check for webhook URL validator
- @stwd/api TypeScript build on the prior exact base before the accounting-only restack

### DIFF
--- a/packages/api/src/__tests__/webhook-url-validation.test.ts
+++ b/packages/api/src/__tests__/webhook-url-validation.test.ts
@@ -62,9 +62,25 @@ describe("webhook URL validation", () => {
     ]) {
       expect(validateWebhookUrl(url)).toBe("url host must be public");
     }
-    // Adjacent prefixes are not accidentally widened by the exact word masks.
-    expect(validateWebhookUrl("https://[2001:2:1::1]/hook")).toBeNull();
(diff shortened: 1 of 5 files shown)


lalalune:
BLOCKING AUDIT: current head overblocks addresses that the authoritative IANA special-purpose registries mark Globally Reachable=True. This includes IPv4 192.31.196/24, 192.52.193/24, 192.175.48/24 and IPv6 2001:3::/32, 2001:4:112::/48, 2001:20::/28, 2001:30::/28, 2620:4f:8000::/48 (plus three 2001:1 anycast /128s). IANA footnote for 2001::/23 explicitly says more-specific allocations override the false default. Do not merge until both registration and connect-time classifiers implement exact longest-prefix exceptions and regressions while retaining truly non-global blocks.

PR #455, opened 2026-08-18
fix(audit): redact best-effort writer failures

lalalune:
## Security finding

The best-effort audit path logged the raw thrown database/canonicalization object. A hostile metadata getter or lower-layer failure could therefore place token, credential, or connection diagnostics into production logs.

This routes the failure through the shared bounded diagnostic sanitizer and logs only the error class plus an explicitly allowlisted platform error code. The regression injects a secret canary through a throwing metadata getter and proves it never reaches `console.error`.

## Exact-head evidence

- `bun test packages/api/src/__tests__/audit-chain-strict.test.ts`: 4 pass, 0 fail, 10 assertions
- `bunx turbo run build --filter=@stwd/api...`: 24/24 packages
- Biome: clean
- `git diff --check`: clean

Keep draft until exact-head CI is green.

### DIFF
--- a/packages/api/src/__tests__/audit-chain-strict.test.ts
+++ b/packages/api/src/__tests__/audit-chain-strict.test.ts
@@ -2,7 +2,7 @@ import { afterAll, beforeAll, describe, expect, it } from "bun:test";
 import { closeDb, getDb, tenants } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
 import { eq, sql } from "drizzle-orm";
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head review: ACCEPT. `trackAuditEvent` now logs only the fixed error class and allowlisted transport code from `redactedThrownDiagnostics`; hostile `instanceof`/code getters are already guarded in the shared helper, and the new audit regression exercises a throwing metadata getter without exposing its canary. Exact local: audit-chain-strict 4/4 (10 assertions), API dependency build 24/24, Biome/diff clean. Merge remains gated on hosted exact-head checks and required review.

PR #456, opened 2026-08-18
fix(provider-x): recover rotating refresh outcomes

lalalune:
Refs #195

Follow-up to merged #438.

## Summary

- commit a tenant/account/revision-bound refresh intent before spending X's single-use refresh token
- encrypt the exact successful provider response before validating access token, rotated refresh token, scope, token type, or expiry
- adopt a valid staged response after crashes without another provider call; disable the exact account revision on ambiguous or invalid outcomes
- retain invalid/scope-widened staged credentials as encrypted revocation handles, autonomously revoke the exact rotated credential, and retry ambiguous revocation without replaying refresh
- run bounded autonomous recovery for abandoned intents, staged responses, and pending revocations in both Bun and Cloudflare Worker runtimes
- block reconnect while an exact staged handle is unresolved; allow reconnect only after proven revocation or safe handle-less supersession
- validate connect-time scope/token type/expiry before identity lookup or state consumption
- classify the new tenant table in the SEC-169 RLS inventory

## Security invariants

- the old rotating refresh token is never posted twice
- malformed/scope-widened/time-invalid successful responses retain encrypted recovery evidence and disable execution
- a pre-provider crash is detected by autonomous recovery; a valid post-stage crash adopts from escrow
- a staged invalid credential is revoked through X's fixed revocation endpoint; failure retains the exact encrypted handle for a later retry
- credential-bearing OAuth calls refuse redirects, and only RFC 7009 HTTP 200 counts as synchronous revocation proof
- Worker cron recovery is enabled only when provider credentials and the master password are present
- reconnect and late recovery are revision-bound, so stale work cannot disable or overwrite the new credential
- raw OAuth material never enters lifecycle or audit rows

## Exact-head evidence

Head: `2fd139dcb7f340f4631f61f6320670af9fd8ecde`
Base: `f4671bf3f705c7db72cf6c28f632284513a5511a`

- combined X lifecycle + actual Worker recovery + migration-tip suite: 40/40, 200 assertions
- migration journal-tip validation: 2/2, 3 assertions
- dependency-aware API build: 24/24 on the exact live head
- Biome and diff-check: green
- frozen dependency install: green

The revocation regressions prove reconnect is denied before cleanup, successful cleanup removes the transient handle, an upstream 503 retains it for a second exact-token attempt, a grammar-rejected rotated refresh string is still revoked exactly as returned, and provider refresh count remains exactly one throughout.

### DIFF
--- a/packages/api/src/__tests__/provider-x-connect.test.ts
+++ b/packages/api/src/__tests__/provider-x-connect.test.ts
@@ -856,7 +856,7 @@ describe("refresh", () => {
     expect(reconnected.completed.reconnected).toBe(true);
   });
 
-  test("failed staged revocation retains the exact handle and retries without refresh", async () => {
+  test("non-200 2xx revocation retains the exact handle and retries without refresh", async () => {
     const { completed } = await connectHappy(new MemoryConnectStore());
     const invalid = installFakeX({
       refreshResponses: [
@@ -869,7 +869,7 @@ describe("refresh", () => {
           },
         },
       ],
-      revokeStatuses: [503, 200],
+      revokeStatuses: [202, 200],
     });
     await expect(
       refreshXProviderCredential({
@@ -1357,11 +1357,14 @@ test("XConnectError carries code + http status", () => {
 
 test("real X transport rejects oversized provider responses before parsing", async () => {
   const originalFetch = globalThis.fetch;
-  globalThis.fetch = (async () =>
-    new Response("{}", {
+  let redirect: RequestRedirect | undefined;
+  globalThis.fetch = (async (_input, init) => {
+    redirect = init?.redirect;
+    return new Response("{}", {
       status: 200,
       headers: { "content-type": "application/json", "content-length": "1048577" },
-    })) as typeof fetch;
+    });
+  }) as typeof fetch;
   try {
     await expect(
       __runDefaultXForwardForTests({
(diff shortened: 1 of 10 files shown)


lalalune:
Exact-head blocker at `842508015021a787606368145b8cc4e52149fc06`: the new handle-preservation rule correctly prevents reconnect from discarding a live staged response, but it leaves no operational recovery path. `runXCredentialLifecycleSweep` selects only `inflight` / `credential_staged`; a malformed or scope-widened staged response is moved to `needs_attention` with `credentialSecretId`, reconnect now rejects it, and no route/scheduler terminalizes or revokes it. The account is permanently disabled even though Steward retains the exact encrypted response.

Required closure: durably claim exact-handle revocation, revoke the bounded staged refresh/access token at X without any refresh repost, retain the handle on ambiguous revoke failure for retry, and atomically mark revoked/delete the transient secret on proven success. Add a regression covering invalid staged response -> needs_attention -> autonomous exact-token revoke -> reconnect succeeds, with provider refresh count remaining one.

lalalune:
Deep security re-audit update on exact head `520968593abc3e3f7c841781df3bb5d22cd65648` (kept draft pending exact-head CI): closed connect-code cross-replica double exchange, post-exchange orphan-grant revocation, redirect credential replay, malformed/unbounded connect state and identity inputs, credential/account identity confusion, slow healthy refresh false-disable races, in-memory rotated-grant staging failure, lifecycle DB shape/revision integrity, and scheduler activation without X credentials. Preserved prior exact staged-token revocation proof. Regression evidence on the exact pushed tree: 46 focused provider-X/Worker tests, 234 assertions, 0 failures; dependency-aware API graph build 24/24 from zero cache; Biome and diff-check clean. Do not merge until the new exact-head Actions run is fully green.

PR #458, opened 2026-08-18
fix(security): redact runtime failure telemetry

lalalune:
## Security finding

Runtime paths still passed caught exceptions, provider diagnostics, DB/Redis failures, and failure strings directly into console logs, durable audit metadata, autonomous-agent logs, or failure webhooks. Depending on the upstream error, those values can contain tokens, credential-bearing URLs, provider bodies, database details, or other secrets.

## Fix

- route caught failures through the shared bounded redactedThrownDiagnostics classifier across API, proxy, auth, trading, Redis, and agent-trader paths
- replace raw failure text in durable audit records and outgoing failure webhooks with fixed messages plus allowlisted diagnostics
- make unauthenticated auth/provider failures return fixed public errors where raw exception text was reflected
- recursively redact nested autonomous-trader metadata, including webhook failure diagnostics and credential fields
- add a TypeScript-AST invariant rejecting raw caught values/messages/stacks at console, audit, warning, and webhook sinks

## Exact-head evidence

Head: `21208bc9d35ca3bc092385ddae9cfef7a083719e`
Base: `ce75a7d28854c1a974658d9a362e3908dc93be96`

- focused safe-error, AST invariant, logger, auth-audit, and scheduler tests: 42/42, 246 assertions
- dependency-aware API + agent-trader graph: 25/25
- Biome and diff-check: green
- original four-commit series remained patch-identical across the current-develop rebase

Broader patch-equivalent package evidence before the final nested-metadata regression: auth 278 pass / 1 environment-gated skip; agent-trader 54 pass; plugin-trading 163 pass.

Keep unmerged until exact-head hosted CI and required review complete.

### DIFF
--- a/packages/agent-trader/src/__tests__/logger-redaction.test.ts
+++ b/packages/agent-trader/src/__tests__/logger-redaction.test.ts
@@ -4,7 +4,7 @@
  * `items: [{ apiKey: ... }]`) previously logged secrets in the clear.
  */
 import { afterEach, beforeEach, describe, expect, it, spyOn } from "bun:test";
-import { logWebhook } from "../logger";
+import { logError, logSubmission, logWebhook } from "../logger";
 
 let writeSpy: ReturnType<typeof spyOn> | undefined;
 let lines: string[];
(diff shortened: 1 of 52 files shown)


lalalune:
Second-pass regression audit found the AST guard could allow a raw throwable when the same telemetry argument also contained one sanitizer call (for example `{ safe: redactedThrownDiagnostics(err), raw: err }`). Exact head `0fe35698` now treats only the sanitizer call subtree as safe, distinguishes object/property labels from catch-binding references, and adds a mixed safe+raw regression. Exact focused invariant suite: 4/4 pass; Biome and diff-check clean. Keep draft until the newly queued exact-head CI is fully green.

lalalune:
Exact-head status: `21208bc9d35ca3bc092385ddae9cfef7a083719e` now has all 53 hosted checks successful, with no failed or cancelled checks. Independent adversarial review covered nested provider diagnostics through the generic agent-trader logger; the closed metadata sanitizer redacts raw error/stack/token-bearing fields while preserving only fixed classes/codes. Local focused evidence remains 42/42 (246 assertions), dependency build 25/25, Biome/diff clean. Marking ready; required review remains the merge gate.

PR #459, opened 2026-08-18
fix(db): bound Workers transaction transport

lalalune:
## Security corrective after #457

The merged transport primitive was sound in its explicit handle/close and transaction-local RLS semantics, but a post-merge audit found two defense-in-depth gaps before activation:

- production TLS policy consulted process.env, not the authoritative Worker bindings passed with the database URL;
- the request-scoped Neon pool inherited unbounded connection/query/lock/idle transaction defaults.

This corrective validates TLS against the exact Worker binding object and configures a 10s connection-acquisition bound plus 30s client/server query, statement, lock, idle-pool, and idle-transaction limits. PostgreSQL cancellation is scheduled 100ms before the client hard bound. Documentation records the contract.

## Exact-head evidence

Head: `9d7f811e4be3d45fcb20276865eb050f2ceec654`
Base: `ce75a7d28854c1a974658d9a362e3908dc93be96`

- DB deadline, RLS context, and Neon driver: 17/17 tests, 48 assertions
- direct @stwd/db build: green
- Biome and diff-check: green
- one-commit range-diff across the rebase: patch-identical

Keep unmerged until exact-head hosted CI and required review complete.

### DIFF
--- a/packages/db/CLOUDFLARE.md
+++ b/packages/db/CLOUDFLARE.md
@@ -92,7 +92,8 @@ Worker request finishes. Connection acquisition is capped at 10 seconds; query,
 statement, lock, idle-connection, and idle-in-transaction phases are capped at
 30 seconds (with PostgreSQL cancellation scheduled slightly earlier). Production
 TLS policy is evaluated from the Worker bindings passed to the handle, not from
-a Node compatibility shim's `process.env`.
+a Node compatibility shim's `process.env`; because Cloudflare does not provide
+`NODE_ENV` automatically, missing `NODE_ENV` defaults to production enforcement.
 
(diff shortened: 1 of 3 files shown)


lalalune:
Exact-head revalidation after the externally added TLS-default hardening:

- Head: `214aa6be48944543873fb9a11af6bc91979c6e0d`
- Worker transaction/deadline focused tests: 11/11, 29 assertions
- `@stwd/db` TypeScript build: green

The added commit defaults request-owned Worker transport to production TLS when Cloudflare omits `NODE_ENV`; explicit non-production remains supported. No blocker found. Required review and hosted exact-head CI remain merge gates.

PR #464, opened 2026-08-18
fix(security): close runtime telemetry redaction gaps

lalalune:
## Summary

- redact aliased exception data before persistence, logs, audit metadata, webhooks, and public responses
- cover provider reservations, proxy releases, retention, webhooks, price oracle, policy/trading, vault/intents, and Eliza runtime paths
- expand the runtime guard across package and web sources with alias-taint handling
- add hostile behavioral regressions for persisted and public failure boundaries

## Exact-head evidence

Head: `f4101e2b50342e22ab61f375e5cbfd4d5b07970e`
Base: `63e0b8ddd705e43e0415651d2880034a8d38ba83`

- focused redaction/price/webhook/retention suites: 48/48 tests, 241 assertions
- proxy persisted/public failure boundary: 6/6 tests, 48 assertions
- affected dependency build: 25/25 tasks
- Biome on all 19 changed files and diff-check: green

### DIFF
--- a/packages/api/src/__tests__/cumulative-spend-cap-e2e.test.ts
+++ b/packages/api/src/__tests__/cumulative-spend-cap-e2e.test.ts
@@ -808,7 +808,7 @@ describeRedis("#206 cumulativeSpend cap - full-chain E2E (real service + real Re
       .where(eq(providerActionReservationGenerations.intentId, out.intentId));
     expect(crashed.state).toBe("pending");
     expect(crashed.attempts).toBe(1);
-    expect(crashed.lastError).toContain("injected crash");
+    expect(crashed.lastError).toBe("reservation reconciliation failed");
     expect(crashed.nextRetryAt).not.toBeNull();
     expect(crashed.handles).toMatchObject({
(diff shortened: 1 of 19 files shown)


lalalune:
Security re-audit update at exact head `886037bcfda333be5c679472054f1c8a04ebefce` (base `63e0b8ddd705e43e0415651d2880034a8d38ba83`): strengthened the whole-runtime error-leak guard to track caught values written into mutable object/array payloads before telemetry (for example `payload.error = err.message` and `values.push(err.stack)`). Added a regression; guard is now 7/7 green. Exact candidate also passed changed webhooks 29/29, shared oracle 10/10, policy 490/490, retention audit 3/3 (18 environment skips), Eliza Vitest 17 pass/24 intentional skips, 24-package API build, 12-package Eliza build, Biome, and diff-check. The new proxy execution-error regression passed; unrelated older lifecycle assertions currently fail on the merged auth baseline and are left to exact hosted CI. Waiting for the newly triggered exact-head matrix before merge.

PR #460, opened 2026-08-18
feat(db): propagate request-owned Worker database

lalalune:
SEC-169 checkpoint: request-owned transaction-capable Worker database propagation.

## Summary

- add async request-local database propagation so existing getDb() consumers use one explicit Worker-owned handle
- bind both neon-http and neon-websocket fetches plus every scheduled recovery job to independent request database capabilities
- reject unsupported Worker drivers, nested request database replacement, detached use after close, and raw getSql() bypass inside a request context
- await socket cleanup on success/failure without allowing cleanup diagnostics to replace a handler error
- validate authentication-critical Worker configuration before opening the request database on every fetch
- await the best-effort boot audit so no database work survives handle cleanup
- document that this is transport plumbing only; RLS remains disabled until trusted tenant middleware and route propagation land

## Exact-head evidence

Head: `8674485fe2b32734db132ad4076bec1371b02520`
Base: `63e0b8ddd705e43e0415651d2880034a8d38ba83`

- exact Worker boot/runtime suites: 14/14 tests, 32 assertions
- dependency-aware API graph: 24/24 builds
- Biome on all changed TS files and diff-check: green
- three-commit range-diff from the independently reviewed predecessor is patch-identical after the #461 rebase

This PR intentionally does not close SEC-169 and does not enable RLS.

PR #474, opened 2026-08-19
fix(worker): retain request DB for background writes

lalalune:
## Security corrective

Post-merge audit of #460 found that request DB capability was revoked immediately after `app.fetch()` resolved, while existing webhook and best-effort audit paths intentionally start DB-backed work without awaiting it. On Workers this could drop webhook deliveries and token-expiry audit telemetry with `REQUEST_DATABASE_CONTEXT_CLOSED`; the WebSocket driver could also close the pool underneath that work.

This corrective:
- adds explicit request DB background-task registration;
- hands the bounded drain/close promise to Worker `waitUntil`, so responses remain non-blocking while the capability stays valid only for registered work;
- registers all configured-webhook dispatches and best-effort audit writes;
- preserves immediate revocation for unregistered detached work;
- validates cron security bindings before opening any of its three DB handles.

Regression evidence on exact head `be8934eee56d076729eb8ca9a2567452bb5976d2` / base `ce5eabd26950ae13cf7a617b33d2c7061eef9a2b`:
- request DB + Worker lifecycle + boot security: 21/21, 55 assertions;
- direct API typecheck: pass;
- dependency-aware API graph: 24/24 builds (on the immediately preceding current-develop aggregate; the final base-only auth delta is covered by exact API typecheck);
- Biome and diff-check: pass.

Keep draft until exact-head hosted CI is green.

### DIFF
--- a/packages/api/src/__tests__/worker-boot-validation.test.ts
+++ b/packages/api/src/__tests__/worker-boot-validation.test.ts
@@ -90,4 +90,27 @@ describe("workers boot JWT env validation (SEC-134)", () => {
       ),
     ).rejects.toThrow('AGENT_TOKEN_EXPIRY "not-a-duration" is not a valid positive duration');
   });
+
+  it("validates scheduled security bindings before opening any database handle", async () => {
+    snapshotEnv();
+    const worker = await freshWorker();
+    let waitUntilCalls = 0;
+    await expect(
+      worker.scheduled(
+        {},
+        {
+          NODE_ENV: "production",
+          STEWARD_JWT_SECRET: "short",
+          DATABASE_DRIVER: "bogus",
+          DATABASE_URL: "postgresql://credential-canary@worker.invalid/steward",
+        },
+        {
+          waitUntil() {
+            waitUntilCalls += 1;
+          },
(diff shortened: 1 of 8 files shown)


lalalune:
Second-pass audit at exact live head `fcbc37c933e793663d45e9dd39eb8f1bdbc71950`: the follow-up isolation commit is materially stronger. Registered jobs now execute in child AsyncLocalStorage contexts, while the owner request context is revoked immediately when the handler settles; an unregistered detached task therefore cannot piggyback on a registered webhook/audit job lifetime. Exact focused validation: 22/22 tests, 57 assertions; direct API typecheck; Biome and diff-check all green. The dependency-aware 24-package API build was green on the parent with the same production Worker/API surface. Waiting for exact-head hosted CI.

lalalune:
Superseded by the stronger current-base corrective PR #481. #481 closes the captured raw/query capability escape, drains registered work on both handler success and failure, and binds configured webhook fan-out to the request lifetime with broader real-Drizzle/Worker regressions. Keeping one authoritative correction avoids merging the older competing lifecycle design.

PR #475, opened 2026-08-19
fix(provider-x): order staged connects by audit sequence

lalalune:
## Summary
- replace timestamp ordering of staged X connect grants with the tenant audit sequence
- fail closed when a connected account lacks the staged exchange-intent audit anchor
- prove equal lifecycle timestamps and an older account timestamp cannot let an old grant overwrite a newer reconnect

## Why
PostgreSQL timestamps are not a monotonic total order: equal timestamps or clock rollback can defeat the `createdAt`/`updatedAt` comparison introduced in #469. Tenant audit sequence numbers are serialized and monotonic, so they provide the authoritative connect-lineage ordering.

## Validation
- `bun test packages/api/src/__tests__/provider-x-connect.test.ts` — 55/55, 348 assertions
- `bunx biome check packages/api/src/services/provider-x-connect.ts packages/api/src/__tests__/provider-x-connect.test.ts`

Refs #195

### DIFF
--- a/packages/api/src/__tests__/provider-x-connect.test.ts
+++ b/packages/api/src/__tests__/provider-x-connect.test.ts
@@ -899,6 +899,24 @@ describe("connect exchange recovery", () => {
       );
     expect(staleLifecycle.providerAccountId).toBeNull();
 
+    const [newerLifecycle] = await getDb()
+      .select({ createdAt: providerXCredentialLifecycles.createdAt })
+      .from(providerXCredentialLifecycles)
+      .where(
+        and(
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head review BLOCKS `4c71eb7e`: staged intent ancestry is required, but the existing account's authoritative lineage is not. An older staged recovery can overwrite a newer same-X account created through generic `provider.account.create`, because the empty X-specific adoption query is treated as “no newer adoption.” Missing/tampered existing-account adoption anchors have the same effect. Require and compare an authoritative current-account anchor (including generic creation) or fail closed/revoke, with both interleaving regressions. Existing 55/55 tests do not cover this case.

PR #476, opened 2026-08-19
feat(db): bind EVM nonce state to tenant ownership

### LINKED ISSUE #343: gap(security): enforce database RLS for tenant isolation (SEC-169)
## Gap

Security audit finding SEC-169 remains documentation-only: tenant isolation is enforced in application queries, but PostgreSQL row-level security is not implemented.

## Required outcome

- Define tenant session/transaction context that cannot be caller-forged.
- Enable and force RLS on every tenant-scoped table, including partitions and future migrations.
- Add deny-by-default policies for missing tenant context.
- Prove cross-tenant reads, writes, upserts, deletes, joins, background jobs, migrations, and pooled-connection context reset fail closed.
- Cover privileged maintenance roles explicitly; no routine API path may use an RLS-bypass role.
- Add real-Postgres concurrency/pool-reuse tests and an operator rollout/rollback plan.
- Inventory any tables intentionally excluded and justify them.

## Context

PR #333 documents the current application-layer boundary but does not implement RLS. This issue tracks the remaining architecture work and must not be considered closed by documentation alone.

lalalune:
## Summary
- add a tenant-owned `evm_wallet_nonce_owners` claim so one `(wallet_address, chain_id)` namespace cannot allocate independently across tenants
- tenant-scope nonce counters and in-flight rows with composite ownership foreign keys while retaining the global allocator lock
- backfill only unambiguous legacy ownership and fail the migration on missing or cross-tenant ambiguous namespaces
- move the nonce tables from the SEC-169 backfill inventory into direct tenant ownership
- prove simultaneous cross-tenant claims serialize to one owner and wrong-tenant drop/confirm operations cannot mutate the owner’s in-flight nonce

## Security invariants
- two tenants cannot claim the same address on the same chain
- the same address on different chains may be owned independently
- legacy missing or ambiguous ownership fails closed for operator reconciliation
- counter/in-flight reads and writes bind the authenticated tenant throughout allocation, confirmation, and gap recovery
- a non-owner cannot force nonce reuse or delete another tenant’s in-flight state

## Validation (exact head `affd5644a4e3b9f3a2495732bef7a31d710a4251`, base `a680098cc5d78127dd802d9cece198fa92a471b9`)
- EVM nonce migration: 3/3, 6 assertions
- EVM nonce manager: 10/10, 30 assertions
- SEC-169 RLS inventory: 2/2, 81 assertions
- `@stwd/shared`, `@stwd/db`, and `@stwd/vault` TypeScript builds: green
- fresh hosted exact-head CI: pending; keep draft until fully green

## Scope
This is a bounded SEC-169 prerequisite checkpoint and intentionally does **not** close #343. FORCE RLS activation still requires restricted bootstrap roles/functions, request-wide transaction propagation (including Workers/background jobs), policies for every direct/derived/hybrid table, and real-Postgres pool/concurrency proofs.

Refs #343

### DIFF
--- a/docs/security/database-rls-rollout.mdx
+++ b/docs/security/database-rls-rollout.mdx
@@ -85,9 +85,11 @@ a Drizzle table is added without classification.
   reputation cache, and archive chunks) use an `EXISTS` policy through their
   tenant-owned parent. Parent indexes and immutable ownership constraints are
   activation prerequisites.
-- EVM nonce tables cannot safely infer a tenant from wallet address because an
-  address is not tenant-unique. Add and backfill `tenant_id` plus a composite
-  ownership foreign key before enabling their policies.
+- EVM nonce counters are tenant-scoped and reference the globally unique
+  `(wallet_address, chain_id)` claim in `evm_wallet_nonce_owners`. Migration
+  `0108` backfills only namespaces that resolve to exactly one tenant and fails
+  closed on missing or ambiguous legacy ownership; the claim prevents two
+  tenants from allocating the same on-chain nonce independently.
 - `user_push_subscriptions` has intentionally nullable `tenant_id`. Tenant
   subscriptions and global user subscriptions require separate, explicit
   policies; it must not inherit the uniform direct-table policy generator.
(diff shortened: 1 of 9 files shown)


lalalune:
Independent semantic review: ACCEPT as a bounded SEC-169 checkpoint at `5db5bad6` (not as issue closure). The runtime derives the EVM address from the tenant-scoped agent key, retains the global address+chain lock, claims `(wallet_address, chain_id)` exactly once across tenants, and tenant-binds every counter/in-flight operation. Migration 0108 lowercases legacy namespaces, resolves both legacy agent and EVM wallet ownership, and aborts rather than guessing when ownership is missing or ambiguous. Composite FKs plus the global claim prevent duplicate cross-tenant allocation. Exact local evidence remains migration 3/3, nonce manager 8/8, inventory 2/2, DB/Vault builds, Biome/diff green. Keep draft until current-base CI/review; #343 must remain open for transaction propagation and FORCE-RLS activation.

Issue #466, opened 2026-08-18
security: make X disconnect cleanup and recovery complete


lalalune:
## Problem

The merged X disconnect lifecycle preserves a revocation handle, but malformed or misbound credentials can still block local revocation, disconnect can reuse an active refresh lifecycle, and account credential pointers, bound routes, and old secrets can remain usable after disconnect. Recovery also needs bounded leased claims and revision-safe route restoration.

## Acceptance criteria

- Local account revocation is not blocked by loading or validating the upstream credential.
- Disconnect uses a separate durable lifecycle and never destroys active refresh escrow.
- Bound routes are disabled, the account credential pointer is cleared, and the old secret is soft-deleted atomically.
- Reconnect restores only routes disabled by disconnect whose authority revision is unchanged.
- Revocation recovery uses a reclaimable lease, bounded exponential retry, and terminal handling while retaining required escrow.
- Applied migration 0106 remains byte-identical; schema follow-up uses a new migration.
- Behavioral migration, lifecycle, route, vault, and failure-injection tests pass.

PR #478, opened 2026-08-19
fix(provider-x): bind disconnect to deterministic secret lineage

lalalune:
Corrective follow-up to #467 and the lifecycle requirements in #466.

## Security correction

- resolve disconnect ownership from the deterministic `provider-x/<workspace>/<x-user>` active secret lineage under the account lock
- treat `provider_accounts.credential_secret_id` as mutable bookkeeping, never as proof of secret ownership
- prevent a wrong pointer from revoking, deleting, or disabling an unrelated secret and its routes
- discover the deterministic lineage when the pointer is null, then durably stage revocation, disable live routes, soft-delete the lineage, and clear local account authority atomically
- retain the existing bounded leased recovery path and immutable 0106/0107 migration history

## Exact-head evidence

Head: `6c08d289338bfe3dc8459f11a1cd09b980d95dcf`
Base: `ce5eabd26950ae13cf7a617b33d2c7061eef9a2b`

- provider-X plus disconnect migration and vault lifecycle: 64/64 tests, 382 assertions
- adversarial wrong-pointer test proves unrelated secret and live route remain unchanged while the deterministic X grant is revoked
- adversarial null-pointer test proves the hidden deterministic live route is disabled and its X lineage is deleted/revoked
- API dependency build: 24/24 tasks
- changed-file Biome: pass
- `git diff HEAD^ --check`: pass

This PR is intentionally draft pending independent review. It does not alter migrations or recovery bounds.

### DIFF
--- a/packages/api/src/__tests__/provider-x-connect.test.ts
+++ b/packages/api/src/__tests__/provider-x-connect.test.ts
@@ -49,6 +49,7 @@ import {
   refreshXProviderCredential,
   runXCredentialLifecycleSweep,
   X_ADAPTER_KEY,
+  X_DEFAULT_SCOPES,
   XConnectError,
   type XCredentialPayload,
   type XForwardRequest,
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head review BLOCKS `6c08d289`: null-pointer recovery selects only the newest non-deleted deterministic secret, but schema permits multiple live versions. Older live versions and enabled routes remain usable after disconnect. Deterministic name alone is also insufficient ownership proof: a same-name mismatched `xUserId` secret can be revoked/deleted. Lock and classify all live lineage rows, prove binding before destruction, then revoke/disable/delete every owned version or fail closed on ambiguity. Add multi-version/routes and same-name mismatched-binding migrated-DB regressions.

PR #479, opened 2026-08-19
fix(api): advertise PATCH in tenant CORS preflight

lalalune:
## Summary
- include `PATCH` in the tenant CORS allow-methods response
- exercise a real Hono `OPTIONS` request with an allowlisted production origin
- retain and assert fail-closed 403 behavior for an untrusted origin
- use the supported database teardown path and assert credential/header/cache boundaries

## Validation
Exact head `93833990c51f8e30c79e15c8af9eafad68e53b23`, base `63e23a0230face5a214855fe28f7f3b814681ddf`:
- tenant CORS preflight: 2/2, 10 assertions
- `@stwd/api` TypeScript build: green
- Biome changed surfaces and `git diff --check`: green

Closes #473

PR #480, opened 2026-08-19
fix(api): advertise PATCH in CORS preflight

### LINKED ISSUE #473: fix(api): advertise PATCH in tenant CORS preflight
## Problem

The API exposes multiple live `PATCH` routes, but `tenantCors` advertises only `GET, POST, PUT, DELETE, OPTIONS` in `Access-Control-Allow-Methods`. Browsers therefore reject otherwise allowed cross-origin PATCH requests at preflight.

## Acceptance criteria

- Allowed CORS preflights advertise `PATCH`.
- A behavioral OPTIONS request proves browser-visible method parity.
- Existing fail-closed origin behavior remains unchanged.

lalalune:
Closes #473

## Summary
- advertise `PATCH` alongside the API methods already supported by live routes
- keep the explicit test/development wildcard path independent of database availability
- replace CORS implementation-string assertions with browser-visible OPTIONS behavior
- mount security-header behavior directly so this focused suite no longer requires application database bootstrap

## Exact-head validation (`76bdb019`)
- dependency-aware API build: 24/24 tasks
- `middleware-security-hardening.test.ts`: 10/10, 62 assertions
- `packages/api` TypeScript build: pass
- Biome changed files: pass
- `git diff --check`: pass

### DIFF
--- a/packages/api/src/__tests__/middleware-security-hardening.test.ts
+++ b/packages/api/src/__tests__/middleware-security-hardening.test.ts
@@ -1,7 +1,9 @@
-import { describe, expect, it } from "bun:test";
+import { afterEach, describe, expect, it } from "bun:test";
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
-import app from "../app";
+import { Hono } from "hono";
+import { securityHeaders } from "../middleware/security-headers";
+import { tenantCors } from "../middleware/tenant-cors";
 
 const apiRoot = join(import.meta.dir, "..");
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as duplicate of #479. The concurrent #479 is the narrower canonical PATCH fix and proves an allowlisted production preflight against PGLite. I will carry the separately discovered development-wildcard/database-independence behavior as its own issue instead of widening #473.

PR #481, opened 2026-08-19
fix(worker): contain request-owned database work

lalalune:
## Summary

Corrective follow-up to merged PR #460 for the SEC-169 program in #343.

- wrap each request-owned Drizzle handle and derived query capability in a revocable lease membrane, so a handle, method, or query builder captured before request completion cannot reach the pool after release
- track database promises started while the lease is active and drain registered detached tasks on both handler success and failure before closing the Worker pool
- register the intentionally fire-and-forget configured webhook fan-out with the request database lifetime while preserving the existing Bun process-owned behavior
- add adversarial captured-handle, captured-method, captured-query, owner-error, real-Drizzle, and Worker webhook lifecycle regressions

This PR deliberately keeps #343 open; it corrects the request-lifetime checkpoint from #460.

## Exact-head validation

Head: `636c053fc7aa3fe0d6e109fed8f0ed95579001b9`
Base tested: `63e23a0230face5a214855fe28f7f3b814681ddf`

- request database context: 7 passed, 25 assertions
- Worker runtime + configured webhook dispatch: 18 passed, 61 assertions
- API dependency build: 24/24 packages successful
- Biome: 5 changed TypeScript files clean
- `git diff --check`: clean

## Independent repair

Exact head `160ec45b91b57b8940969ad876917224546533e9` additionally isolates registered background work in child capability contexts. The prior shared context let unregistered detached work piggyback while a webhook kept the request alive. Regression now proves the unregistered path is revoked immediately. Exact local: request DB 8/8 (27 assertions), dependency build 24/24, Biome/diff clean.

### DIFF
--- a/packages/api/src/__tests__/worker-runtime-sentinel.test.ts
+++ b/packages/api/src/__tests__/worker-runtime-sentinel.test.ts
@@ -1,7 +1,7 @@
 import { expect, test } from "bun:test";
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
-import { getDb } from "@stwd/db";
+import { getDb, tenants, waitUntilRequestDatabaseTask } from "@stwd/db";
 import {
   hydrateProcessEnv,
   runWorkerGoogleCredentialLifecycleSweep,
@@ -44,7 +44,8 @@ test("Worker request database is exact, request-owned, and always closed", async
       env,
       async () => {
         await Promise.resolve();
-        expect(getDb()).toBe(requestDb);
+        expect(getDb()).not.toBe(requestDb);
+        expect((getDb() as unknown as { marker: string }).marker).toBe("worker-request");
         return "ok";
       },
       { createHandle },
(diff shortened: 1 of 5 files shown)


lalalune:
Superseded by #474. The current #474 head retains nonblocking Worker waitUntil cleanup and explicit audit/webhook registration, isolates registered child contexts, prevents unregistered piggybacking, and now revokes captured DB handles/methods/query builders with real-Drizzle regression coverage.

lalalune:
Independent source review: no new code blocker found at `636c053f`. The revocable membrane checks every captured DB/method/query capability on access/invocation, tracks thenable driver work, drains registered work on both owner success and failure, and only closes the Worker pool after that drain. Configured webhook fan-out is registered with the same request lease; non-Worker Bun behavior stays process-owned. Existing adversarial evidence covers captured handles/methods/query builders, real Drizzle execution, owner failure, and webhook-before-close. Keep draft until exact hosted CI and independent required review complete.

PR #484, opened 2026-08-19
fix(security): repair runtime redaction packaging

lalalune:
## Summary

Corrective follow-up to #464.

- bundle both `@stwd/shared` runtime entrypoints into `@stwd/eliza-plugin` and remove the unpublishable `workspace:*` runtime dependency
- add a tarball-level artifact regression that builds, packs, and inspects the published manifest and bundle
- extend the runtime telemetry guard to Promise `.catch` callbacks, `.then(_, onRejected)` callbacks, direct rejection sinks, aliases, and the Waifu console wrapper
- replace raw rejection logging in the DB migrator, Sol default-policy seed, and Waifu integration example with bounded diagnostics

## Exact-head evidence

Head: `617c950f2ebb9fc576c2b26c15e531cd87108953`
Base: `63e23a0230face5a214855fe28f7f3b814681ddf`

- runtime error-log safety guard: 10/10 tests
- Eliza plugin Vitest: 54 passed, 24 intentional skips across 8 files (includes 2 artifact regressions)
- Sol default-policy seed: 6/6 tests, 18 assertions
- affected dependency build: 10/10 tasks
- `npm pack` artifact `@stwd/eliza-plugin@0.4.4`: clean fresh-consumer install with `--omit=peer`, 4 packages added, 0 vulnerabilities; installed runtime dependencies contain only `@stwd/sdk`
- frozen lockfile install: green
- Biome on all changed source/config files and `git diff --check`: green

This PR is intentionally draft for corrective review and must not be merged until hosted checks and review are complete.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -148,10 +148,9 @@
     },
     "packages/eliza-plugin": {
       "name": "@stwd/eliza-plugin",
-      "version": "0.4.3",
+      "version": "0.4.4",
       "dependencies": {
         "@stwd/sdk": "^0.10.0",
-        "@stwd/shared": "workspace:*",
       },
       "devDependencies": {
         "@elizaos/core": "2.0.0-alpha.77",
@@ -159,6 +158,7 @@
         "@stwd/db": "workspace:*",
         "@stwd/proxy": "workspace:*",
         "@stwd/proxy-client": "workspace:*",
+        "@stwd/shared": "workspace:*",
         "@types/node": "^25.5.0",
         "hono": "^4.12.34",
         "tsup": "^8.0.0",
@@ -493,6 +493,7 @@
       "version": "0.4.0",
       "dependencies": {
         "@stwd/db": "workspace:*",
+        "@stwd/shared": "workspace:*",
         "@stwd/vault": "workspace:*",
         "drizzle-orm": "^0.44.5",
         "viem": "^2.30.0",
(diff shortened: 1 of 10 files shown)


lalalune:
Independent source review: ACCEPT semantics at `617c950f`, pending current-base hosted CI/review. The public Eliza artifact no longer contains any `workspace:*` runtime dependency, both shared entrypoints are bundled, and the package version is bumped. The runtime telemetry guard now covers catch clauses plus Promise rejection callbacks/direct sinks/aliases, while the three concrete raw rejection paths use bounded redacted diagnostics. Lockfile additions match the new runtime imports for DB/seed and the Eliza dev-only bundling dependency. Exact tarball install and focused/build evidence in the PR body are sufficient; keep draft until required checks/review complete.

lalalune:
Corrective review found two packaging blockers before readiness:

1. `packages/eliza-plugin/package.json` still has no `files` allowlist. `npm pack` therefore includes repository source/tests/config rather than only the distributable artifact. Add a narrow publish allowlist (at minimum `dist` and the intended package documentation) and extend the tarball test to assert no `src`, test, or build-config entries are shipped.
2. The manifest moves to `0.4.4`, but `packages/eliza-plugin/CHANGELOG.md` has no `0.4.4` entry. Document the published artifact/redaction correction at that version.

The runtime dependency bundling and Promise-rejection redaction changes otherwise look coherent. Keep draft until these are fixed and exact-head artifact/consumer/CI checks rerun.

PR #490, opened 2026-08-19
fix(auth): preserve consumed OAuth exchanges

### LINKED ISSUE #485: fix(auth): preserve OAuth redemption after lock cleanup failure
## Problem

Both OAuth nonce-redemption paths atomically consume a one-time code and then await deletion of an ancillary 10-second redemption lock in a `finally` block. If the consume succeeds but the subsequent Redis/Postgres lock deletion fails, the cleanup error replaces the successful response with HTTP 500. The code is already gone, so retry cannot recover the minted access/refresh-token handoff.

## Expected behavior

Once atomic consume succeeds, ancillary lock cleanup failure must not change the terminal redemption result. The short-lived lock may expire naturally. Atomic consume failures must continue to fail closed and must never return tokens.

## Acceptance criteria

- Both `/auth/oauth/exchange` and the bound-code branch of `/auth/oauth/:provider/token` return the consumed token payload when only lock deletion fails.
- Cleanup diagnostics are redacted.
- Atomic consume failures still return an error and no token.
- Behavioral tests cover both successful-consume/failed-cleanup paths and consume failure where practical.

lalalune:
## Summary

- treat bounded redemption-lock deletion as advisory after the atomic one-time code consume result
- retain redacted operational diagnostics when lock cleanup is unavailable
- behaviorally cover both OAuth token handoff routes and preserve fail-closed consume errors

## Validation

- `bunx biome check packages/api/src/routes/auth.ts packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts`
- `bun test --timeout 120000 packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts` — 15 pass, 0 fail, 50 expectations
- `bunx turbo run build --filter=@stwd/api` — 24/24 tasks successful

Closes #485

### DIFF
--- a/packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts
+++ b/packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts
@@ -16,8 +16,10 @@ import {
   expect,
   it,
   setDefaultTimeout,
+  spyOn,
 } from "bun:test";
 import { readFileSync } from "node:fs";
+import { ChallengeStore } from "@stwd/auth";
 import { closeDb } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
 import { Hono } from "hono";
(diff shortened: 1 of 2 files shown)


lalalune:
Independent follow-up review removed the new mutable production backend test hook. Failure injection now spies on the real `ChallengeStore` methods only within the behavioral test and restores them in all paths. The consume-failure case now proves preservation by successfully redeeming the same code after the injected outage clears.

Exact refreshed head validation: OAuth suite 15/15 (50 expectations), affected API build 24/24, Biome and diff-check green.

PR #505, opened 2026-08-19
fix(agent-trader): remove untrusted price oracle fetch

lalalune:
## Summary

Remediates CodeQL alert #39 and a confirmed SSRF/resource-exhaustion surface.

The undocumented `PRICE_ORACLE_URL` fallback accepted an arbitrary environment/file-derived URL, followed redirects, parsed an unbounded body, and authenticated neither the endpoint nor its response. Its output was already low-confidence and could not safely authorize a price-driven trade.

- remove the arbitrary external fetch entirely
- retain the hardened liquidity-weighted oracle as the only high-confidence source
- retain on-chain reserve fallback as low-confidence, strategy-gated telemetry
- correct the trust-order documentation and add a source invariant that blocks direct network-fetch reintroduction in the state resolver

## Exact-head evidence

Head `8773caf24018d612b173a3d79bf92ca5e13825a1`, rebased on current develop at validation time:

- full agent-trader suite: 55/55, 135 assertions
- dependency-aware agent-trader graph: 8/8 builds

Keep draft until exact-head hosted CI and a subsequent CodeQL scan are green.

### DIFF
--- a/packages/agent-trader/src/__tests__/state-security.test.ts
+++ b/packages/agent-trader/src/__tests__/state-security.test.ts
@@ -1,0 +1,11 @@
+import { expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+test("agent state never fetches an operator-controlled price-oracle URL", () => {
+  const source = readFileSync(join(import.meta.dir, "../state.ts"), "utf8");
+  expect(source).not.toContain("PRICE_ORACLE_URL");
+  expect(source).not.toContain("PRICE_ORACLE_");
+  expect(source).not.toContain("fetch(url");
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking cleanup before readiness: `state-security.test.ts` is a pure source-text assertion (`readFileSync` + string presence/absence), which does not execute the pricing behavior and directly conflicts with the repository test-quality goal. Keep the production removal, but replace this file with behavioral coverage through the public state/strategy path: set hostile `PRICE_ORACLE_*` env values, make `fetch` fail if called, stub the hardened oracle/reserve inputs, and assert the resulting confidence/price contract without any operator-controlled network call. Then rerun the full agent-trader suite on the exact head.

PR #506, opened 2026-08-19
test(deps): satisfy security floor formatting

lalalune:
## Summary

- sort the dependency security-floor test imports according to Biome
- restore the develop lint/typecheck workflow after #494 merged with a formatting violation

## Validation

- `bun test scripts/__tests__/dependency-security-floors.test.ts` (1/1, 12 assertions)
- `bunx biome check scripts/__tests__/dependency-security-floors.test.ts`
- `git diff --check`


lalalune:
Superseded by merged #510, which applies the same Biome import-order correction on the post-#494 develop tip. No unique functional or test delta remains.

PR #507, opened 2026-08-19
fix(vault): reconcile ambiguous local EVM broadcasts

lalalune:
## Summary
- locally prepare and sign EVM transaction bytes before the mutating RPC boundary
- derive and durably checkpoint the deterministic transaction hash before submission
- disable transport retries for `eth_sendRawTransaction`, submit the exact bytes once, and reconcile a lost response once by hash
- keep allocator-issued nonces reserved for every unresolved, mismatched-hash, and post-broadcast persistence failure
- release a nonce only when preparation or the durable checkpoint fails before any mutating RPC
- preserve the existing outcome-unknown wire contract for API/SDK compatibility

## Security invariant
After signed bytes can have reached an RPC, Steward must never classify the attempt as safely retryable or make its nonce reclaimable unless chain state proves the exact deterministic hash was accepted. A lost response must produce a durable `outcome_unknown` record and require hash reconciliation.

## Exact local validation
Head `c656f34be32dfeaeb480afc31b7934cfcc53ccef` on develop `f14389fb1ff8c1186b183c08ce59fa46c6373ff2`:
- local deterministic broadcast lifecycle + wiring: 7/7, 25 assertions
- `@stwd/vault` and `@stwd/api` builds: green
- broader compatibility suites and fresh hosted exact-head CI are pending; keep this PR draft until fully green

## Failure semantics
- prepare/checkpoint failure: no mutating RPC occurred, so best-effort nonce release is safe
- RPC response loss + hash found: finalize the deterministic accepted hash without reposting
- RPC response loss + hash absent/unavailable: retain nonce and return outcome-unknown
- mismatched RPC hash: retain nonce and return outcome-unknown
- final persistence failure: retain nonce; the pre-broadcast outcome-unknown checkpoint remains authoritative

Security audit follow-up to #476.

### DIFF
--- a/packages/vault/src/__tests__/local-evm-broadcast-wiring.test.ts
+++ b/packages/vault/src/__tests__/local-evm-broadcast-wiring.test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, test } from "bun:test";
+import { readFile } from "node:fs/promises";
+
+describe("local EVM broadcast wiring", () => {
+  test("routes local-key broadcasts through the checkpointed raw-transaction lifecycle", async () => {
+    const source = await readFile(new URL("../vault.ts", import.meta.url), "utf8");
+
+    expect(source).toContain("return executeLocalEvmBroadcast({");
+    expect(source).toContain('status: "outcome_unknown"');
+    expect(source).toContain("publicClient.sendRawTransaction({ serializedTransaction })");
+    expect(source).toContain("transport: http(rpcUrl, { retryCount: 0 })");
+    expect(source).toContain('status: localRecordOptions.status ?? "broadcast"');
+    expect(source).not.toContain("client.sendTransaction({");
+  });
+});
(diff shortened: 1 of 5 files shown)


lalalune:
Exact-head follow-up validation for `c656f34be32dfeaeb480afc31b7934cfcc53ccef`:
- local deterministic broadcast lifecycle/wiring: 7/7, 25 assertions
- AWS deterministic broadcast compatibility: 19/19, 90 assertions
- EVM nonce manager: 6/6, 15 assertions
- external custody outcome-focused: 3/3, 15 assertions
- API vault execution gateway: 27/27, 177 assertions
- `@stwd/vault` and `@stwd/api` builds: green
- Biome/diff: green

The PR remains draft pending fresh hosted exact-head CI.

lalalune:
Two blockers before readiness: (1) `local-evm-broadcast-wiring.test.ts` is a pure source scan and does not prove the real Vault local-custody branch uses the lifecycle. Replace it with a public `Vault.signTransaction` behavioral test that stubs preparation/sign/broadcast/reconcile/persistence and proves one raw submission, durable pre-submit checkpoint, zero transport replay, nonce retention on ambiguity, and accepted finalization. (2) `releaseBeforeBroadcast()` failures are swallowed twice with no bounded diagnostic. Retaining the nonce is safety-correct, but a failed allocator transition can wedge future sends invisibly; emit redacted operational telemetry (without replacing the original pre-submit failure) and test it. The lifecycle unit tests are useful and should remain.

Issue #500, opened 2026-08-19
test(api): replace agent audit source scans with route and rollback behavior


lalalune:
`packages/api/src/__tests__/agent-audit-order.test.ts` and `agent-audit-rollback.test.ts` read `routes/agents.ts` and assert substrings/order. These tests can stay green while real route behavior, transaction rollback, headers, or audit persistence is broken. They also encode historical PR commentary rather than current executable contracts.

Replace them incrementally with behavioral route/DB coverage before deleting any assertion:

- recent-MFA rejection for agent creation, token mint, wallet/signer/quorum privilege changes, and policy mutations, with no DB/audit mutation on rejection;
- authorization/final audit rows queried after successful mutations;
- no-store/Pragma/Expires on secret-bearing token and signer responses;
- injected required-audit failures proving agent, wallet, signer, quorum, and policy mutations roll back or restore exactly;
- agent deletion failure/retry behavior proving revocation, row deletion, restoration, and final audit semantics;
- server-generated agent/policy IDs proven through route calls and cross-tenant/global collision cases rather than source text.

Delete or trim each source assertion only in the same narrow change that adds its behavioral replacement. Run against real Postgres where advisory locks/concurrency/rollback behavior matters.

PR #509, opened 2026-08-19
test(api): replace agent token audit source scans

lalalune:
Refs #500

## Tranche scope
This first behavioral tranche removes only source assertions with executable replacements:

- token mint rejection without recent MFA now proves no authorization/final audit row is written
- successful token mint proves ordered authorization/final audit rows and all three non-cacheable response headers
- the nested policy-rule ID source scan is removed because `agent-policy-rules.test.ts` already calls the route with a colliding caller ID and proves a fresh server-generated ID
- historical PR commentary in the touched signer scan is removed

All uncovered rollback, deletion, wallet, signer, quorum, agent-ID, and remaining MFA source assertions are intentionally retained for later issue #500 tranches. The signer cache scan is also retained; the broader signer behavioral suite is currently not reliable enough on develop to justify deleting it.

## Exact-head validation
Validated at `e579beea5fb0074bced85b21878b923c293821fd` after rebasing onto `f14389fb1ff8c1186b183c08ce59fa46c6373ff2`.

- `bun install --frozen-lockfile`
- `bunx turbo run build --filter=@stwd/api` (24/24 tasks)
- agent audit source suites (8/8)
- agent admin mutation behavior (5/5)
- agent policy-rule behavior (4/4)
- Biome on all changed files

### DIFF
--- a/packages/api/src/__tests__/agent-admin-mutations-mfa.test.ts
+++ b/packages/api/src/__tests__/agent-admin-mutations-mfa.test.ts
@@ -1,6 +1,7 @@
 import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
-import { agents, closeDb, getDb, tenants } from "@stwd/db";
+import { agents, auditEvents, closeDb, getDb, tenants } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { and, asc, eq, inArray } from "drizzle-orm";
 import { Hono } from "hono";
 import type { AppVariables } from "../services/context";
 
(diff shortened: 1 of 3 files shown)


lalalune:
Independent test-integrity audit on live head `d1281ed933f358d3cebfe5690d464254f145fab6` (patch-id `6dc323bb03536e9aaa99376e9f648b8be8065ab8`): no blocker found.

The removed token source scans have genuine route/DB replacements: recent-MFA denial returns 403 with no authorization or issuance audit mutation; successful mint proves ordered authorization/final audit rows and all three non-cache headers. The removed nested-rule scan is already covered by a real route call that ignores a colliding caller ID and returns a fresh generated UUID. Unreplaced signer/rollback scans remain in place, so this tranche does not weaken uncovered invariants.

Validated the identical patch after restack: four owning suites 17/17 (171 assertions), API build, Biome, and diff-check pass. PR remains draft; no Actions were manipulated.

PR #514, opened 2026-08-19
fix(seed): generate demo API credentials

lalalune:
## Summary
- replace the repository-known demo tenant bearer key with a fresh standard Steward API key on every seed run
- store the generated hash and reveal the plaintext only in the interactive seed summary
- add regression coverage for format, uniqueness, verification, and cross-key rejection

## Security context
CodeQL alert #30 points at SHA-256, but SHA-256 is appropriate for lookup of a high-entropy bearer token. The actual weakness was that the input was a repository-known constant, making any seeded tenant immediately authenticatable if the seeder was used in a reachable environment.

## Validation
- `bun test --timeout 30000 packages/seed/src/__tests__/demo-api-key.test.ts`
- `bunx @biomejs/biome check packages/seed/src/index.ts packages/seed/src/demo-api-key.ts packages/seed/src/__tests__/demo-api-key.test.ts`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ __pycache__/
 .env.local
 .env.*.local
 !.env.example
+.steward/
 # Operator-local fleet inventory (SEC-130 — node IPs are never committed)
 scripts/deploy-nodes.local.conf
 
--- a/packages/seed/src/__tests__/demo-api-key.test.ts
+++ b/packages/seed/src/__tests__/demo-api-key.test.ts
@@ -33,6 +33,13 @@ describe("demo API key", () => {
       symlinkSync(path, symlink);
       expect(() => writeDemoCredentials("other", pair.key, symlink)).toThrow();
       expect(readFileSync(path, "utf8")).toContain("STEWARD_TENANT_ID=waifu.fun");
+
+      const linkedParent = join(root, "linked-parent");
+      symlinkSync(dirname(path), linkedParent);
+      expect(() =>
+        writeDemoCredentials("other", pair.key, join(linkedParent, "other.env")),
+      ).toThrow("parent must be a real directory");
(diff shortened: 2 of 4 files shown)


lalalune:
Directionally correct, but please fix two package/operational seams before readiness:

1. `@stwd/seed` imports `../../auth/src/*` without declaring `@stwd/auth`. Use the public `@stwd/auth` export, add the workspace dependency through Bun, regenerate the lockfile, and validate package-context build/test so the seed package is not coupled to another package's source tree/hoisting.
2. The new live bearer key is always printed to stdout. In non-interactive/container seed runs that persists the credential in CI/deployment logs. Make plaintext reveal an explicit operator action (TTY-only or a deliberate `--show-credential`/secure owner-only output contract); default logs should report where/how to retrieve it without exposing it. Test both default non-interactive redaction and explicit reveal. Also document that rerunning the seed rotates the tenant key and invalidates the prior credential.

Keep the fresh high-entropy key generation and exact hash verification tests.

lalalune:
The log-exposure blocker is fixed at exact head `34a3ee51695feeda8ee1e90b8b6e61be782282e8`: the default summary prints only the owner-only file path, and the behavioral test verifies mode `0600` plus symlink rejection. One packaging blocker remains before readiness: both production (`packages/seed/src/demo-api-key.ts`) and the test still import `../../auth/src/*` / `../../../auth/src/*`, while `packages/seed/package.json` still does not declare `@stwd/auth`. Please import `generateApiKey` and `validateApiKey` through public `@stwd/auth`, avoid the private `ApiKeyPair` source import (or publicly export that type), add `"@stwd/auth": "workspace:*"` via Bun so `bun.lock` is regenerated, then rerun the seed package test/Biome on the rebased exact head. Current focused evidence: 2/2 seed tests pass and Biome passes, but they pass under monorepo source layout and therefore do not establish the package boundary.

PR #523, opened 2026-08-19
fix(api): audit agent enrollment token issuance

### LINKED ISSUE #511: fix(api): make agent enrollment token issuance auditable
`packages/api/src/routes/agent-enroll.ts` currently signs and returns a privileged short-lived agent JWT even when the allow audit append fails: both deny and allow paths call `writeAuditEvent(...).catch(() => {})`, with no telemetry. A storage outage can therefore create an unrecorded enrollment credential, while audit failure is completely invisible.

Implement a narrow fail-closed/durable design:

- distinguish authorization from successful issuance in current event names/metadata;
- durably append the required authorization event before token signing/return, or use an audit outbox that cannot lose the event;
- ensure a signing failure cannot leave an audit row falsely claiming a token was issued;
- preserve uniform denial responses, but emit only bounded redacted operational diagnostics if a denial-audit append fails;
- add route-level failure injection proving an allow-audit outage returns no token, a signing failure does not claim issuance, the happy path records the expected event(s), and diagnostics never expose nonce/signature/token/error canaries;
- remove the silent catches rather than replacing them with raw exception logging.

Run the focused enrollment route/core tests and the affected API build on the exact final head.

lalalune:
Closes #511.

## Summary

- durably record successful enrollment authorization before signing any agent JWT
- record successful issuance separately after signing and before returning the token
- fail closed with no token when either required audit append is unavailable
- ensure signing failures leave only an authorization event, never a false issuance event
- preserve uniform verification denial responses while replacing silent audit catches with bounded redacted diagnostics
- add route-level PGLite failure injection for authorization-audit outage, issuance-audit outage, signing failure, denial-audit outage, and the successful authorization/issuance sequence

No implementation-source scans or new production test hooks.

## Validation

Exact head: `2e668e9c818750e0b79bcbc809ae94146ecafb1b`
Base: `1a3b0d9542db43f638cf2c3bff5fca6e789f5b93`

- `agent-enroll-route.test.ts`: 9 pass, 0 fail, 53 assertions
- `agent-enroll-abuse.test.ts`: 3 pass, 0 fail, 10 assertions
- `bun run --cwd packages/api build`: pass
- Biome on both changed files: pass
- `git diff --check origin/develop...HEAD`: pass

### DIFF
--- a/packages/api/src/__tests__/agent-enroll-route.test.ts
+++ b/packages/api/src/__tests__/agent-enroll-route.test.ts
@@ -7,7 +7,7 @@
 
 import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
 import { generateP256KeyPair, signP256, verifyToken } from "@stwd/auth";
-import { agentSigners, agents, getDb, tenants } from "@stwd/db";
+import { agentSigners, agents, auditEvents, eq, getDb, tenants } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
 import { Hono } from "hono";
 import type { AppVariables } from "../services/context";
@@ -18,6 +18,8 @@ const AGENT_ID = `enroll-agent-${Date.now()}`;
 setDefaultTimeout(30000);
 
 let keypair: Awaited<ReturnType<typeof generateP256KeyPair>>;
+let auditTransactionFailure: Error | undefined;
+let successfulAuditTransactionsBeforeFailure = 0;
 
 async function makeApp() {
   const { agentEnrollRoutes } = await import("../routes/agent-enroll");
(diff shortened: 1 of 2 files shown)


lalalune:
Independent security audit on exact head `2e668e9c818750e0b79bcbc809ae94146ecafb1b`: no blocker found.

The route now durably appends authorization before signing, appends issuance only after signing, and returns no token when either required append fails. A signing failure cannot create a false issuance record. Denials remain uniform, and every new operational diagnostic is reduced to fixed-shape `redactedThrownDiagnostics`; nonce, signature, token, and thrown canaries are absent. The existing nonce single-use and unknown-agent behavior remain covered.

Exact-head validation: enrollment route + abuse suites 12/12 (63 assertions), API build, Biome, and diff-check pass. PR remains draft; no Actions were manipulated.

PR #530, opened 2026-08-19
fix(auth): stage email challenges until delivery

### LINKED ISSUE #512: fix(auth): stage email challenges until provider acceptance
`packages/auth/src/email.ts` persists redeemable magic-link and OTP records before provider delivery, then treats invalidation as best-effort (`invalidate().catch(() => {})`) when the provider throws or returns no receipt. If the provider outcome is ambiguous and cleanup storage is unavailable, Steward returns an error but leaves a live credential that may already have been delivered. Retrying can create multiple simultaneously redeemable challenges.

Replace the delete-after-failure pattern with a durable staged lifecycle:

- persist magic-link and OTP challenges in a non-redeemable `delivery_pending` state;
- activate them only after a valid provider acceptance receipt;
- verification/consume must reject staged, failed, malformed, and expired records;
- provider ambiguity or activation failure must leave the challenge durably non-redeemable without depending on best-effort deletion;
- superseding a prior challenge must remain atomic and fail closed;
- add behavioral fault injection for provider throw-after-possible-delivery, missing receipt, cleanup/store outage, activation outage, concurrent redemption during delivery, successful activation, and retry/supersession;
- log only bounded redacted diagnostics, never recipient, code, link token, provider body, or raw exception.

Run the full auth email suite and affected auth/API builds on the exact final head.

lalalune:
Closes #512

## Summary

- persist magic-link, companion-code, and standalone OTP credentials as durable `delivery_pending` records before provider I/O
- validate a bounded provider acceptance receipt, then commit activation; provider rejection or ambiguity leaves every delivered credential non-redeemable without cleanup
- resolve lost activation acknowledgements by reading back the durable active record rather than returning an ambiguous failure
- preserve rolling-deploy compatibility for previously issued magic-link and OTP records
- supersede prior magic-link and OTP challenges before publishing a retry, while keeping link/code redemption atomic through their shared record
- retain bounded redacted diagnostics with no recipient, token, code, provider body, or raw error text

## Behavioral coverage

- provider throw after capturing the deliverable secret, with delete storage unavailable
- missing/malformed acceptance receipt
- durable staging outage before provider invocation
- activation failure before commit for magic-link and OTP
- lost storage acknowledgement after a committed activation
- simultaneous redemption while provider delivery is in flight
- atomic link-vs-code redemption
- concurrent magic-link retry/supersession and OTP retry/supersession
- malformed/failed stored records, TTL expiry, tenant/email binding, and one-time consumption
- API route behavior: 502 + non-redeemable staged credentials on rejection, 200 only after acceptance

## Exact-head validation

Validated commit `7522b578b8a51e7cc52ac3d782dcee4d9120c935` on develop `37d60fa1064b53b92889d1e6a1f36b0431bf975e`:

- `bunx turbo run build --filter=@stwd/api...` — 24/24 package builds pass
- `cd packages/auth && bun test --timeout 120000 src/__tests__` — 295 pass, 1 skipped integration test, 0 fail, 893 expectations
- API email suite (delivery, OTP signup, login-code contract, tenant config, platform config) — 36 pass, 0 fail, 184 expectations
- `bunx biome check` on all three changed files — pass
- `git diff --check origin/develop...HEAD` — pass
- worktree clean; final diff limited to auth email implementation and targeted auth/API behavioral tests

### DIFF
--- a/packages/api/src/__tests__/auth-email-delivery-failclosed.test.ts
+++ b/packages/api/src/__tests__/auth-email-delivery-failclosed.test.ts
@@ -5,7 +5,7 @@
  * unless the provider ACCEPTED the message:
  *   - production with no delivery-capable provider  → 503, no challenge issued
  *   - tenant with partial/unsupported email config  → 503, no challenge issued
- *   - provider rejects the send                     → 502, challenge invalidated
+ *   - provider rejects the send                     → 502, challenge remains staged
  *   - provider returns an acceptance receipt        → 200, challenge live
  */
 import { afterAll, beforeAll, describe, expect, it } from "bun:test";
@@ -135,7 +135,7 @@ describe("fail-closed email delivery routes (production)", () => {
     expect(body.error).toBe("Email delivery is not configured");
   });
 
-  it("returns 502, invalidates the challenge, and keeps logs redacted when the provider rejects", async () => {
+  it("returns 502, keeps the challenge staged, and redacts logs when the provider rejects", async () => {
     // Prime the tenant cache with a delivery-capable EmailAuth (Resend key
     // present), then swap in a rejecting provider — the seam a Resend outage
     // or invalid key hits in production.
@@ -174,7 +174,7 @@ describe("fail-closed email delivery routes (production)", () => {
     expect(body.ok).toBe(false);
     expect(body.error).toBe("Email could not be sent. Please try again later.");
 
-    // The would-be credentials must not redeem (challenge invalidated).
+    // The would-be credentials must not redeem while delivery remains staged.
     const token = text.match(/[?&]token=([a-f0-9]{64})/)?.[1] ?? "";
     const code = text.match(/\b(\d{6})\b/)?.[1] ?? "";
     expect(token).not.toBe("");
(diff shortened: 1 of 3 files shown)


lalalune:
Post-merge audit found a rolling-deploy compatibility blocker in final head `7d47732d37f9fa804bd21b081464af2cb78b618e`.

The new reader accepts legacy records, but the old reader cannot consume records issued by the new writer:

- A newly activated magic-link challenge is serialized with `status:"active"`; the deployed reader only parses `status:"pending"`, and it consumes the alias/challenge before rejecting it.
- A newly activated OTP is serialized as `{status:"active",payload:{...}}`; the deployed reader consumes that value and treats the wrapper JSON as the legacy email payload, then rejects it.

During a rolling deploy, one request routed to an old replica can therefore permanently destroy a provider-accepted, newly delivered credential. The new tests cover new-reader/old-record compatibility but not old-reader/new-record compatibility.

I am preparing a narrow follow-up atop the merged result: retain `delivery_pending` as the pre-acceptance gate, but serialize activated records in the exact legacy wire formats, while continuing to read the short-lived `active` wrapper records already emitted by this merge. Behavioral tests will emulate both deployed readers consuming newly issued magic-link and OTP records.

PR #532, opened 2026-08-19
fix(cli): preserve redacted migration TLS diagnostics

lalalune:
## Summary
- tag the production missing-TLS database rejection with a fixed `DB_TLS_REQUIRED` code
- allow only that bounded code through the existing redacted diagnostic helper
- update the CLI migration regression to distinguish the TLS gate from a network failure without exposing the database URL or raw error text

## Validation
- 34 focused CLI, database-security, and safe-error tests pass (148 assertions)
- six dependency-aware builds pass
- Biome and `git diff --check` pass

Fixes the current `develop` CLI test failure introduced when migration errors were correctly redacted.

### DIFF
--- a/packages/cli/src/__tests__/init.test.ts
+++ b/packages/cli/src/__tests__/init.test.ts
@@ -342,10 +342,12 @@ describe("steward init", () => {
       );
       const [stderr, exitCode] = await Promise.all([new Response(proc.stderr).text(), proc.exited]);
       expect(exitCode).toBe(1);
-      // SEC-087: the gate now demands a server-authenticating TLS mode —
-      // sslmode=require alone no longer satisfies it.
-      expect(stderr).toContain("sslmode=verify-full");
-      expect(stderr).toContain("sslmode=verify-ca");
+      // SEC-087: the child reached the TLS gate rather than attempting a
+      // network connection. The migrator exposes only this fixed diagnostic
+      // code; raw exception text and the connection URL remain redacted.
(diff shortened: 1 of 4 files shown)


lalalune:
Exact merged-head security audit: PASS (no production blocker found).

Reviewed exact restacked head `b1c0cac5dc9e4676584f3254376adaaa67d72232` on parent `3ad606de3a9df6a207ac185bbfc8dd86bd01b17d` (merged as `9bb2a653744c046a2f04f2e25b06bf6e5bc4102d`). Its stable patch ID `914d4650a52c47271a01a623bc94c385efe9d989` exactly matches the originally assigned head `8ecb252c306fb09e2d9eb1b7d68e50d8469221ec`.

Security review confirmed the migration path emits only `redactedThrownDiagnostics`; the new `DB_TLS_REQUIRED` value is a fixed, bounded allowlisted machine code and neither the exception message nor database URL/host/credentials can pass that boundary. TLS validation semantics remain fail-closed and unchanged: authenticated modes are accepted, ambiguous/missing TLS is rejected, and explicit insecure/unverified acknowledgements retain their existing behavior.

Exact-head validation:
- dependency-aware Turbo build: 6/6 packages passed
- focused CLI/DB/shared suites: 43 passed, 0 failed, 169 assertions
- Biome: 5 owning/touched files passed
- `git diff --check`: passed

I also prepared stronger local regression assertions for username/password/scheme non-disclosure and direct pinning of the DB error code, but automation merged and deleted the source branch before a guarded push, so those supplemental assertions were not pushed onto this already-merged PR.

Issue #525, opened 2026-08-19
fix(vault): behaviorally verify local broadcast wiring and report nonce release failures


lalalune:
PR #507 merged the local EVM deterministic broadcast lifecycle, but two review blockers remain on current `develop`:

1. `packages/vault/src/__tests__/local-evm-broadcast-wiring.test.ts` only scans `vault.ts` source strings. It does not prove the public `Vault.signTransaction` path checkpoints before exactly one raw broadcast, reconciles ambiguity, or retains the nonce. Replace it with a behavioral public-Vault test using scoped dependencies/fakes.
2. `executeLocalEvmBroadcast` silently swallows `releaseBeforeBroadcast` failures after preparation or checkpoint failure. Keep the original pre-broadcast error as the thrown error, but emit bounded redacted telemetry for the cleanup failure. Never log provider/RPC URLs or secret-bearing error messages.

Acceptance:
- public Vault behavior replaces the source scan;
- preparation and checkpoint failure tests inject release failure and assert the original error survives;
- telemetry is bounded/redacted and contains no injected secret;
- focused Vault tests/build pass on the exact final head.

PR #535, opened 2026-08-19
fix(vault): report pre-broadcast nonce release failures

lalalune:
Addresses #525.

## Summary
- report failed pre-broadcast EVM nonce releases with a fixed, bounded diagnostic
- preserve the original preparation/checkpoint failure so caller behavior remains stable
- behaviorally inject cleanup failures at both safe-release boundaries and prove secret-bearing diagnostics are redacted

## Audit conclusion
- the two best-effort transitions in `upstream-leases.ts` leave `acknowledging` or `revoking` claims that the stale recovery sweep can reclaim; callers still fail closed
- accepted EVM nonce confirmation remains intentionally non-fatal: the deterministic broadcast is already accepted, `nextNonce` remains monotonic, and only explicitly `dropped` rows can be reclaimed

## Validation
Exact head checks after the final rebase:
- `bun run --cwd packages/shared build`
- `bun test packages/vault/src/__tests__/local-evm-broadcast.test.ts` (7 pass)
- `bunx biome check packages/vault/src/local-evm-broadcast.ts packages/vault/src/__tests__/local-evm-broadcast.test.ts`
- `bun run --cwd packages/vault build`

Broader package validation before the final base refresh:
- `bun run --cwd packages/vault test` (448 pass, 9 skip, 0 fail)

### DIFF
--- a/packages/vault/src/__tests__/local-evm-broadcast.test.ts
+++ b/packages/vault/src/__tests__/local-evm-broadcast.test.ts
@@ -72,6 +72,54 @@ describe("local EVM deterministic broadcast lifecycle", () => {
     expect(subject.events).toEqual(["prepare", "checkpoint", "release"]);
   });
 
+  test("preserves pre-broadcast failures and reports release failures with redacted diagnostics", async () => {
+    const originalConsoleError = console.error;
+    const diagnostics: unknown[][] = [];
+    console.error = (...args: unknown[]) => diagnostics.push(args);
+
+    try {
+      for (const stage of ["prepare", "checkpoint"] as const) {
+        const operationError = new Error(`${stage} failed`);
+        const subject = lifecycle({
+          prepare: async () => {
+            subject.events.push("prepare");
+            if (stage === "prepare") throw operationError;
+            return SERIALIZED;
+          },
+          checkpoint: async () => {
+            subject.events.push("checkpoint");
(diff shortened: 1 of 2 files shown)


lalalune:
Exact merged-byte security audit: PASS.

Reviewed final PR head `849f4f29b9da1e89407c798f259c9d770fcc63dc` (merged as `b70c158a04e169d3c93d50c009ca2935ea21c207`). The final head includes the additional guard that prevents a throwing diagnostic sink from replacing the original pre-broadcast operation failure.

Outcome classification remains fail-closed: prepare/checkpoint failures are the only release points; every failure after the first mutating RPC call remains `external_broadcast_outcome_unknown`; reconciliation performs one read-only lookup without rebroadcast; confirmation/persistence failures never mark the nonce reclaimable. Release failure logging contains only a fixed message plus `redactedThrownDiagnostics`, with no transaction bytes, provider errors, URLs, tenant identifiers, tokens, or secrets.

Independent validation:
- exact final-head local lifecycle suite: 8 passed, 0 failed, 26 assertions
- allocator concurrency/tenant ownership plus lifecycle/wiring: 18 passed, 0 failed, 61 assertions on the restacked production patch
- shared and vault builds passed
- Biome and `git diff --check` passed

No Actions were manipulated.

PR #536, opened 2026-08-19
fix(seed): bind credential promotion to validated inode

lalalune:
## Summary
- open each pending credential once with `O_NOFOLLOW`, validate its inode, ownership, link count, and mode through the descriptor, and keep it open through promotion validation
- verify both the promotion pathname and its opened descriptor identify that same validated inode before atomic rename
- reject non-canonical Steward API keys before the exported staging API writes dotenv content
- add deterministic pathname-swap and dotenv-injection regressions

Minimal race-hardening follow-up to merged #533.

## Tests
- `bun test --timeout 30000 packages/seed/src/__tests__/demo-api-key.test.ts` (8 pass, 34 assertions)
- `bunx @biomejs/biome check packages/seed/src/demo-api-key.ts packages/seed/src/__tests__/demo-api-key.test.ts`
- `bunx turbo run build --filter=@stwd/api...` (24/24)

### DIFF
--- a/packages/seed/src/__tests__/demo-api-key.test.ts
+++ b/packages/seed/src/__tests__/demo-api-key.test.ts
@@ -5,8 +5,10 @@ import {
   readdirSync,
   readFileSync,
   realpathSync,
+  renameSync,
   rmSync,
   symlinkSync,
+  writeFileSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking exact-head review finding: the promotion candidate is verified, its descriptor is closed, and only then renameSync(promotionPath, finalPath) resolves the pathname again. A same-UID cleanup/racing process can replace the random .promote-* entry after the inode comparison but before rename, allowing an unvalidated inode to become canonical. Keep the candidate descriptor open through rename and verify the canonical path still resolves to the validated staged inode before reporting success (and clean only the candidate inode you created on pre-rename failure). Add a deterministic swap at that second boundary, not only after the pending-file open. Also avoid leaving an attacker-linked .promote-* artifact when the existing mismatch check throws. The current optional afterPendingOpen callback is test-only API surface; prefer an immutable filesystem dependency seam or a behavioral child-process coordination harness if you need deterministic interleaving.

PR #547, opened 2026-08-19
fix(trading): bound Polymarket market metadata transport

lalalune:
## Summary

Security corrective following the merged #537 audit:

- refuse redirects on authoritative token-to-market metadata reads
- apply one wall-clock deadline across fetch and the entire streamed body
- bound decoded response bytes and make body/reader cancellation non-blocking and best-effort
- sanitize provider, malformed URL, stream, abort, and cancellation failures
- reject CLOB endpoint overrides containing credentials, query strings, fragments, controls, or oversized input while retaining the documented HTTPS compatible-gateway deployment boundary
- release pre-execution idempotency claims even when metadata-error or condition-mismatch audit writes fail
- add behavioral regressions for redirects, stalled streams, hostile cancellation, URL-secret canaries, authoritative-condition confusion, persisted diagnostic redaction, unsafe endpoint overrides, and audit-failure retries

Arbitrary HTTPS gateway hosts remain allowed intentionally: this is trusted deployment configuration used for compatible CLOB gateways, not tenant/request input. The corrective removes URL-carried secret material and cross-URL ambiguity without breaking that documented integration surface.

## Validation

Exact production patch validation, then guarded restack onto `2f14d881108b914ecaa09fdeb406aea1d8d9b0f8`:

- 12-package dependency build: passed
- venue-polymarket: 63 passed, 0 failed, 135 assertions
- Polymarket order/session route: 33 passed, 0 failed, 160 assertions
- Biome: passed
- `git diff --check`: passed

No Actions were manipulated.

PR #548, opened 2026-08-19
fix(trading): release metadata outage claims before audit

### LINKED ISSUE #544: Release Polymarket idempotency claims before outage audit writes
## Problem

When Polymarket token metadata lookup fails, the order route writes `trade.order.policy-rejected` before releasing its durable idempotency claim. If the audit backend is also unavailable, the audit exception skips release and replaces the intended sanitized metadata response. The same idempotency key then remains `pending` for its long TTL, so a safe retry returns an in-progress conflict instead of re-attempting metadata verification.

## Expected behavior

- Provider metadata failure remains the primary, sanitized `502 Unable to verify Polymarket market binding` response even when audit persistence is unavailable.
- The durable trade idempotency claim is released before or independently of the fallible audit write.
- Retrying the same key re-runs metadata lookup rather than observing a 24-hour pending claim.
- Strict token-to-condition binding remains unchanged.
- Audit-outage diagnostics remain bounded and do not include provider-controlled error text.

## Regression coverage

Inject a hostile audit failure together with a first-attempt metadata failure, then assert the first response envelope, retry response, and exactly two metadata upstream calls.

lalalune:
Closes #544.

## Summary
- release the Polymarket trade idempotency claim before recording a metadata-lookup rejection
- contain audit persistence failures with bounded redacted diagnostics so they cannot replace the intended sanitized metadata response
- prove two hostile metadata attempts with the same key both reach the upstream lookup even while audit persistence fails
- preserve strict token-to-condition verification and all existing execution/idempotency behavior

## Validation
- `bun test --timeout 120000 packages/plugin-trading/src/__tests__/trade-polymarket-order.test.ts` — 31 passed, 139 assertions
- affected `@stwd/plugin-trading` dependency build — 12/12 packages
- Biome and `git diff --check` passed

No production-global test hook was added; the regression uses the existing route context dependency injection.

### DIFF
--- a/packages/plugin-trading/src/__tests__/trade-polymarket-order.test.ts
+++ b/packages/plugin-trading/src/__tests__/trade-polymarket-order.test.ts
@@ -345,6 +345,51 @@ describe("POST /v1/trade/polymarket/order", () => {
     }
   });
 
+  it("releases the metadata-failure claim when the audit backend is unavailable", async () => {
+    const { tenantId, agentId, sessionId } = await seedSession({
+      allowedAssets: [`pm:cond:${COND_ID}`],
+    });
+    stubWallet(true);
+    const { createTradeRoutes } = await import("../routes/trade");
+    const auditOutageRoutes = createTradeRoutes({
+      ...sharedTestContext,
+      writeAuditEvent: async () => {
+        throw new Error("AUDIT_BACKEND_SECRET_CANARY");
+      },
+    });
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by #547, which includes the same claim-before-audit release and sanitized audit-outage handling together with the complete bounded metadata transport hardening. The merged #547 exact head passed 63 venue tests, 33 route tests, the 12-package dependency build, and Biome.

PR #555, opened 2026-08-19
test(api): preserve tenant KDF proof point

lalalune:
## Summary
- restore the pre-KeyStore observer used by tenant request-signature regressions
- assert unknown, malformed, and key-id-less unauthenticated requests never reach constructor-side scrypt work
- preserve the existing deep-imported test-only export and reset the observer after every test

The merged #545 replacement spy only observed `decrypt()`, but `KeyStore` performs an expensive scrypt in its constructor before decrypt is called. This restores the stronger CPU-amplification regression without reverting the comment cleanups.

## Tests
- `bun test --timeout 30000 packages/api/src/__tests__/authorization-signature-tenant-keys.test.ts` (5 pass, 15 assertions)
- `bunx @biomejs/biome check packages/api/src/middleware/authorization-signature.ts packages/api/src/__tests__/authorization-signature-tenant-keys.test.ts`
- `bunx turbo run build --filter=@stwd/api...` (24/24)

### DIFF
--- a/packages/api/src/__tests__/authorization-signature-tenant-keys.test.ts
+++ b/packages/api/src/__tests__/authorization-signature-tenant-keys.test.ts
@@ -1,4 +1,4 @@
-import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, spyOn } from "bun:test";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "bun:test";
 import { closeDb, getDb, tenantRequestSigningKeys, tenants } from "@stwd/db";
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by #556. This branch restores the exact mutable production-global test hook that #545 was intended to remove. #556 preserves constructor/KDF-level regression evidence through immutable middleware dependency injection, with the same focused behavior and build checks, without a test-named export or cross-test module state.

lalalune:
Superseded by #556; do not merge this mutable-global variant.

PR #566, opened 2026-08-19
fix(solana-signer): reject ambiguous bearer credentials

lalalune:
## Security corrective\n\nFollow-up to externally merged #559. Keeps its AgentNet-compatible `Authorization: Bearer` transport while making the credential boundary unambiguous and bounded:\n\n- reject duplicate custom or Authorization fields using raw header occurrences\n- preserve fail-closed custom-header precedence\n- accept the auth scheme case-insensitively but require exact single-space token68 syntax\n- cap configured and presented secrets at 1024 UTF-8 bytes before hashing\n- keep 401 responses generic and prove secret canaries are not reflected\n\nThis branch is based on the merged/restacked #559 bytes and changes only `packages/solana-signer`; it does not revert #558/#560 or unrelated current-develop work. The loopback bridge still emits no CORS allow-origin response, so adding Authorization does not grant browser access.\n\n## Exact validation\n\n- full solana-signer suite: 50 pass, 129 assertions\n- dependency build: 7/7 packages\n- Biome exact changed files: pass\n- `git diff --check`: pass\n\nDraft pending independent exact-head and hosted CI review. No Actions manipulation or merge requested.

### DIFF
--- a/packages/solana-signer/src/__tests__/bridge.test.ts
+++ b/packages/solana-signer/src/__tests__/bridge.test.ts
@@ -1,7 +1,9 @@
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "bun:test";
+import { connect } from "node:net";
 import { Keypair, Transaction } from "@solana/web3.js";
 import {
   BRIDGE_MAX_BODY_BYTES,
+  BRIDGE_MAX_TOKEN_BYTES,
   BRIDGE_MIN_TOKEN_BYTES,
   BRIDGE_TOKEN_ENV,
   BRIDGE_TOKEN_HEADER,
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head review completed for `69e0c9e6c477df3c3e1777598d30a5385e3feabe` after rebasing onto current `develop` (`9239a728`). The duplicate raw-header handling, custom-header fail-closed precedence, token68 parsing, UTF-8 byte bounds, generic 401 behavior, and loopback/no-CORS boundary are sound. Exact validation passed: full `packages/solana-signer/src/__tests__` (50/50, 129 assertions), dependency build (7/7), Biome on both changed files, and `git diff --check`. Hosted checks remain queued because the organization Actions budget is currently configured at $0 with `prevent_further_usage: true`; this is not a code failure.

lalalune:

PR #567, opened 2026-08-19
fix(worker): isolate pre-hydration JWT validation import

### LINKED ISSUE #561: Keep Worker security validation synchronous with binding hydration
## Problem

## Scope

lalalune:
## Security corrective

Post-merge audit of #565 found that its top-level `@stwd/auth` barrel import loads auth stores and JWT module state before the first Worker request hydrates bindings. This preserves #565's synchronous hydration/validation boundary without importing the side-effectful barrel:

- expose a narrow `@stwd/auth/jwt` package entrypoint and import only that leaf from the Worker entry
- remove the JWT module's import-time `AGENT_TOKEN_EXPIRY` capture from token signing; resolve it at call time after Worker hydration
- keep `AGENT_TOKEN_EXPIRY` as the stable documented 30-day default
- add a behavioral token claim regression proving a post-import hydrated expiry controls the signed token

## Exact validation

- Worker boot + auth session hardening: 10 pass, 27 assertions
- isolated Worker boot/runtime sentinel: 2/2 files
- API dependency build: 24/24 packages; direct API TypeScript clean
- Biome exact changed files and `git diff --check`: clean

Closes #561

### DIFF
--- a/packages/api/src/worker.ts
+++ b/packages/api/src/worker.ts
@@ -39,7 +39,10 @@
  *                                   configured Redis is unreachable
  */
 
-import { validateJwtSecretEnv } from "@stwd/auth";
+// Import the JWT leaf, not the @stwd/auth barrel: the barrel loads auth stores
+// that must not snapshot Worker-sensitive environment state before bindings
+// are hydrated by the first request.
+import { validateJwtSecretEnv } from "@stwd/auth/jwt";
 import {
   createDbForRequest,
   createNeonTransactionDbForRequest,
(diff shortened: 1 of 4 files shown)


lalalune:
Closing as superseded by draft #572. #572 includes the same narrow pre-hydration JWT entrypoint, and additionally removes the API context/routes' explicit module-captured expiry fallback, validates the resolved runtime value, and proves both initial and rotated Worker bindings in minted JWT claims. #567's direct signAgentToken fix alone leaves createAgentToken passing the stale context constant explicitly.

PR #568, opened 2026-08-19
feat(db): add all-or-nothing RLS catalog gate

lalalune:
## SEC-169 bounded checkpoint

Adds a read-only structural catalog gate without creating policies, enabling RLS, or wiring startup activation. The gate permits an entirely inactive catalog and structurally staged policies while every RLS flag remains off. Once any protected table or partition is enabled, it requires the complete composed inventory to be `ENABLE` + `FORCE` in the same catalog snapshot.

The catalog traversal is relation-OID based and includes descendants stored outside the root schema. The caller must explicitly name the runtime role; the gate rejects owner privileges available through `USAGE` or assumption, and `SUPERUSER`/`BYPASSRLS` attributes on the runtime or an assumable role. PostgreSQL 16+ uses per-membership `SET`; PostgreSQL 15 conservatively treats every `MEMBER` role as assumable.

The mandatory inventory covers core Drizzle tables plus always-migrated `auth_kv_store`. Plugins add explicit owned contributions; contributions extend but cannot replace core, and unknown, missing, invalid, or duplicate classifications fail closed. Policy counts remain structural evidence only and do not claim semantic correctness.

## Exact local evidence (`44f059d9381461661888bb3bcfb03e893cb8a4cf`)

- catalog + inventory unit: 9 pass, 102 assertions
- real PostgreSQL 16: 1 pass, 8 assertions; covers external-schema partition discovery, partial activation, full active restricted-role success, inherited-only owner rejection (`USAGE=true`, `SET=false`), assumable-only BYPASSRLS rejection (`SET=true`, `USAGE=false`), SUPERUSER rejection, mandatory auth inventory, and plugin inventory
- PostgreSQL 15 compatibility is a dedicated lightweight job in both PR and develop CI; hosted exact-head result pending
- DB dependency build: 5/5 packages, uncached
- DB direct TypeScript build: pass
- Biome exact TypeScript files: pass
- actionlint and CI test inventory: pass
- diff check: pass

## Still blocked / not included

SEC-169 remains open. This does not enable RLS and is not wired to startup. Restricted bootstrap functions, full request/background transaction propagation, generated direct/derived/hybrid policy expressions, all-table activation, and cross-tenant CRUD/join/upsert proofs remain required.

Draft pending independent exact-head and hosted PostgreSQL 15/16 review. Do not merge until those checks complete.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -494,8 +494,8 @@ jobs:
         env:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
 
-      - name: RLS pool isolation proof (real Postgres)
-        run: bun test ./packages/db/src/__tests__/tenant-rls-postgres.test.ts
+      - name: RLS pool and catalog isolation proof (real Postgres)
+        run: bun test ./packages/db/src/__tests__/tenant-rls-postgres.test.ts ./packages/db/src/__tests__/rls-catalog-gate-postgres.test.ts
         env:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -448,6 +448,41 @@ jobs:
         env:
           STEWARD_MASTER_PASSWORD: "ci-test-secret"
 
+  rls-catalog-postgres15:
+    name: RLS Catalog Gate (Postgres 15)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    services:
+      postgres:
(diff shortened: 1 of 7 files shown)


lalalune:
Independent audit found and repaired one active-catalog gap: the application role ownership check covered parent tables but not protected partitions. Exact head 078c3f8a now evaluates ownership across the full activation set and adds a runtime-owned partition regression. Exact local proof: catalog + inventory 6/6, 92 assertions; DB dependency build 5/5; Biome and diff-check green. Keep draft until fresh exact-head hosted CI, especially the real-Postgres lane. #343 remains open because this is still a read-only structural checkpoint.

lalalune:

PR #570, opened 2026-08-19
fix(eliza): keep API URL normalization linear

lalalune:
## Summary
- replace polynomial trailing-slash regex in `SUBMIT_TRADE` URL normalization with a linear scan
- add an adversarial near-miss regression covering a 200,000-slash runtime URL

## Security context
This was found while reconciling open CodeQL alerts #1-17, #30, #32, #33, #35, and #39-41 against current `develop`. The alert-era ReDoS sinks are already stale after #518/#526, but this adjacent runtime URL path had reintroduced the same normalization pattern.

## Validation
- `bunx turbo run build --filter=@stwd/eliza-plugin...` (12/12)
- `bunx vitest run packages/eliza-plugin/src/__tests__/submit-trade.test.ts` (15/15)
- `bunx biome check packages/eliza-plugin/src/actions/submit-trade.ts packages/eliza-plugin/src/__tests__/submit-trade.test.ts`
- `git diff --check HEAD^ HEAD`


lalalune:

PR #571, opened 2026-08-19
fix(auth): close publication concurrency gaps

### LINKED ISSUE #564: fix(auth): close atomic email publication concurrency gaps
Post-merge audit of #550 found two concurrency gaps:

1. `PostgresBackend.publish()` checks guards before unconditional writes under advisory locks that ordinary backend writers do not acquire. A rolling old pod can mutate a guarded target after the check and before the UPSERT, allowing the publication to overwrite a newer generation.
2. If publication commits but its acknowledgement is lost, a newer issuance can replace the publication marker before retry. The original call then reports failure; if the newer provider attempt fails, the original credential remains active despite the failure response.

Required acceptance:
- make PostgreSQL guarded publication conditional at write/commit time against ordinary writers, including absent/expired expectations; rollback the full batch on mismatch
- add a deterministic real-Postgres interleaving with an ordinary writer after the initial guard observation
- confirm an own committed magic-link/OTP publication after ambiguous acknowledgement loss even when a newer reservation has appeared, or revoke it before reporting failure
- add behavioral magic-link and OTP races covering lost ACK + newer reservation + newer provider failure
- preserve #550 rolling formats, redaction, PII boundaries, and Memory/Redis semantics
- exact focused tests, auth build, Biome, diff check, and CI inventory

lalalune:
Closes #564

## Summary
- run PostgreSQL batch publication at `SERIALIZABLE` and retry serialization failures so an ordinary or rolling old-pod write after the guard read cannot be overwritten
- add an immutable, TTL-bounded per-issuance commit receipt so magic-link and OTP sends can confirm their own publication after a lost storage acknowledgement, even when a newer issuance has reserved the target
- keep confirmation bound to both the exact receipt and the still-active target, preserving consumption safety and newer-writer ownership
- add behavioral magic-link/OTP overlap tests and deterministic real-PostgreSQL interleaving tests using an uncooperative legacy-style writer
- cover existing, absent, and expired publication guards plus receipt opacity, ownership, and TTL bounds

## Validation (exact head `3cf3d40cde19028903254864af94a7c2b3dda099`)
- `bun test --timeout 60000 packages/auth/src/__tests__/email-delivery-failclosed.test.ts packages/auth/src/__tests__/email.test.ts packages/auth/src/__tests__/store-backends.test.ts packages/auth/src/__tests__/challenge-store.test.ts` (60 pass, 229 assertions)
- `DATABASE_URL=... bun test --timeout 60000 packages/auth/src/__tests__/email-postgres-publish.integration.test.ts` against isolated PostgreSQL 16 (3 pass, 22 assertions)
- `bun run --cwd packages/auth build`
- `bunx turbo run build --filter=@stwd/auth...` (7/7 packages)
- Biome on all four changed files
- `git diff --check`

A clean `bun install --frozen-lockfile` reached the existing native `pkcs11js` postinstall and failed because the host Python environment lacks the `gyp.generator.xcode` module. `bun install --frozen-lockfile --ignore-scripts` completed without changing the lockfile; hosted CI remains the clean-install authority.

### DIFF
--- a/packages/auth/src/__tests__/email-delivery-failclosed.test.ts
+++ b/packages/auth/src/__tests__/email-delivery-failclosed.test.ts
@@ -31,6 +31,7 @@ class CapturingBackend implements StoreBackend {
   failActiveWritesAfterCommit = false;
   failReadsAfterActiveCommit = false;
   activeTransitionFailuresAfterCommit = 0;
+  afterActivationCommitBeforeLostAck?: () => Promise<void>;
   failDeletes = false;
   replaceGuardBeforeTransition = false;
 
@@ -152,6 +153,7 @@ class CapturingBackend implements StoreBackend {
       entries.some((entry) => this.isActivation(entry.key, entry.value)) &&
       this.activeTransitionFailuresAfterCommit++ === 0
     ) {
+      await this.afterActivationCommitBeforeLostAck?.();
       throw new Error("durable activation response lost");
     }
     return true;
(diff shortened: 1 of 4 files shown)


lalalune:

PR #572, opened 2026-08-19
fix(worker): defer auth config capture until hydration

### LINKED ISSUE #561: Keep Worker security validation synchronous with binding hydration

lalalune:
Closes #561

## Summary
- import the dependency-light JWT subpath so Worker boot validation does not evaluate the full auth barrel before bindings are hydrated
- resolve and validate default agent-token expiry at mint time instead of module initialization
- cover initial and rotated Worker expiry bindings with exact token lifetime assertions

## Security impact
Merged #565 made validation synchronous, but its new static auth-barrel import captured Worker-sensitive configuration before bindings existed. A configured tighter AGENT_TOKEN_EXPIRY could validate successfully while default token minting still used the stale 30-day fallback.

## Validation
- `bunx turbo run build --filter=@stwd/api...` (24/24 packages)
- auth suite: 315 pass, 3 environment-gated skips
- Worker/runtime/enrollment focused suite: 27 pass
- Biome and diff checks pass

No workflow files changed.

### DIFF
--- a/packages/api/src/__tests__/worker-boot-validation.test.ts
+++ b/packages/api/src/__tests__/worker-boot-validation.test.ts
@@ -6,7 +6,9 @@
  */
 
 import { afterEach, describe, expect, it } from "bun:test";
-import worker from "../worker";
+import { signAgentToken } from "@stwd/auth/jwt";
+import { decodeJwt } from "jose";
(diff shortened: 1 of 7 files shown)


lalalune:

PR #574, opened 2026-08-19
test(proxy-client): configure fail-closed dependencies

### LINKED ISSUE #573: test(proxy-client): provide required audit HMAC fixture
The root `bun run verify` contract fails in `@stwd/proxy-client` because `proxy-e2e.test.ts` exercises credential forwarding without configuring the now-required audit-chain HMAC key. Both happy and reflected-secret paths fail closed at the required pre-forward audit with HTTP 503 instead of reaching the forwarding assertions.

Add an explicit production-shaped `STEWARD_AUDIT_HMAC_KEY` fixture, restore the prior environment exactly, and prove the package suite plus root verify behavior. This is test-fixture drift; production correctly fails closed.

lalalune:
Closes #573

## Summary
- configure the production-required audit HMAC in the proxy-client forwarding fixture
- opt the hermetic test explicitly into the development in-memory replay store while preserving production fail-closed defaults
- restore both pre-existing environment values exactly after the suite

## Evidence
- reproduced current root `bun run verify` failure: two proxy-client E2E cases returned `503 Shared proxy idempotency store unavailable`
- `bunx turbo run build --filter=@stwd/proxy-client...` (10/10)
- `bun test --cwd packages/proxy-client` (21/21, 56 assertions)
- `bunx biome check packages/proxy-client/src/__tests__/proxy-e2e.test.ts`
- `git diff --check`

Production continues to require shared Redis and required audit persistence.

### DIFF
--- a/packages/proxy-client/src/__tests__/proxy-e2e.test.ts
+++ b/packages/proxy-client/src/__tests__/proxy-e2e.test.ts
@@ -19,7 +19,7 @@
 
 import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
 import { signAgentToken } from "@stwd/auth";
-import { agents, closeDb, getDb, tenants } from "@stwd/db";
+import { __resetAuditHmacKeyCacheForTests, agents, closeDb, getDb, tenants } from "@stwd/db";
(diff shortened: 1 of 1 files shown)


lalalune:

PR #576, opened 2026-08-19
test(capabilities): isolate proxy enforcement fixture

### LINKED ISSUE #575: test(plugin-capabilities): configure explicit proxy enforcement fixtures
## Problem\n\nThe CI-equivalent root `bun run verify` fails in `packages/plugin-capabilities/src/__tests__/invoke-e2e.test.ts`: five allowed proxy requests receive 429 instead of their upstream 200/201 responses. The integration fixture dispatches through the real proxy but neither opts into the explicit development replay-store boundary nor installs an allow-only proxy rate-limit checker. Production correctly fails closed when shared enforcement storage is unavailable.\n\n## Acceptance criteria\n\n- Preserve production fail-closed replay/rate-limit defaults.\n- In this hermetic integration fixture only, explicitly set and restore `STEWARD_PROXY_DEV_MODE=true`.\n- Install an explicit allow-only proxy rate-limit dependency and restore all proxy handler hooks in teardown.\n- Keep the real proxy auth, route matching, credential injection, scrubbing, policy and call-count behavior under test.\n- The isolated owning suite and affected package build pass on the exact head.\n\nDiscovered while continuing the repository-wide quality audit and root verifier.

lalalune:
Closes #575.\n\n## Summary\n- explicitly opt the hermetic capability-to-proxy integration fixture into the development replay-store boundary\n- install an allow-only proxy rate-limit checker so policy call-count tests remain the authority under test\n- restore the original dev-mode environment and every mutable proxy-handler dependency in teardown\n\nProduction replay and rate-limit enforcement remain fail closed.\n\n## Exact-head validation\n- `bun test --isolate packages/plugin-capabilities/src/__tests__/invoke-e2e.test.ts` — 16 pass, 72 assertions\n- `bun run build --filter=@stwd/plugin-capabilities...` — 12/12 packages\n- Biome and `git diff --check` — pass

### DIFF
--- a/packages/plugin-capabilities/src/__tests__/invoke-e2e.test.ts
+++ b/packages/plugin-capabilities/src/__tests__/invoke-e2e.test.ts
@@ -72,6 +72,7 @@ interface ForwardedCapture {
 }
 let lastForwarded: ForwardedCapture | null = null;
 let proxyApp: Hono | null = null;
+let proxyRateLimitChecks = 0;
 const realFetch = globalThis.fetch;
 
 // the policy set the injected getPolicySet returns for the current test.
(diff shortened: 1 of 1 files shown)


lalalune:
Independent audit repaired remaining shared-worker environment leakage: teardown now restores every binding mutated by the fixture, not only STEWARD_PROXY_DEV_MODE. Handler hooks and global fetch were already reset correctly. Exact head 3fd72d02: isolated invoke E2E 16/16 (72 assertions), dependency build 12/12, Biome and diff-check green. Keep draft until fresh exact-head CI.

lalalune:
Final exact head adafb263 also snapshots the environment inside beforeAll (not at module-evaluation time), then clears the snapshot after the unconditional nested teardown. The external 9c67 hardening is preserved: hook reset and DB close cannot prevent env restoration, and rate-limit seam invocation is asserted. Exact isolated suite: 16/16, 76 assertions; prior dependency build 12/12; Biome/diff green. Fresh exact-head CI is required.

lalalune:

PR #578, opened 2026-08-19
fix(worker): apply rotated security bindings at request time

lalalune:
## Summary
- resolve request-expiry and authorization-signature requirements, HMAC secrets, and freshness limits per request
- revoke global-wallet execution opt-ins and insecure webhook HTTP immediately when Worker bindings rotate
- refresh Worker global rate limits and the auth default tenant from current bindings while preserving Bun startup constants
- add same-instance rotation regressions for every corrected gate

## Security impact
A long-lived Worker isolate could retain a permissive or obsolete value captured when the app/module was first composed. Removing an unsafe-wallet, insecure-webhook, request-signature, or request-expiry binding would not take effect until isolate recycle; a rotated HMAC secret could also leave the old secret valid. These gates now read the currently hydrated binding synchronously on each request.

## Validation
- request-expiry, authorization-signature, webhook URL: 41/41
- global-wallet route suite: 13/13
- Worker global-rate/default-tenant suite: 5/5
- `bun run --cwd packages/api build`
- Biome on all 13 changed files
- `git diff --check`

No workflow files changed.

### DIFF
--- a/packages/api/src/__tests__/authorization-signature-middleware.test.ts
+++ b/packages/api/src/__tests__/authorization-signature-middleware.test.ts
@@ -46,7 +46,7 @@ function makeDefaultApp() {
   return app;
 }
 
-async function signedHeaders(path = "/vault/agent-1/sign", body = BODY) {
+async function signedHeaders(path = "/vault/agent-1/sign", body = BODY, secret = SECRET) {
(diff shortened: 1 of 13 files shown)


lalalune:

PR #581, opened 2026-08-19
test(api): isolate agent-client proxy fixture

### LINKED ISSUE #580: test(api): isolate agent-client proxy enforcement fixture
## Problem\n\nThe root verifier reaches the real proxy through `agent-client-e2e.test.ts`, but that hermetic fixture neither opts into the development replay-store boundary nor installs an explicit allow-only rate checker. Production correctly returns 429 without shared enforcement storage, so the expected broker invoke 201 fails. The fixture also deletes ambient env values instead of restoring them and does not reset mutable proxy handler dependencies.\n\n## Acceptance criteria\n\n- Preserve production fail-closed enforcement.\n- Explicitly opt only this fixture into proxy dev mode and allow-only rate checking.\n- Snapshot at setup and restore every environment binding changed by the fixture.\n- Reset all mutable proxy handler dependencies during teardown.\n- Keep the real AgentClient, enrollment, manifest, broker invoke, proxy auth, route matching, secret injection, and scrubbing behavior.\n- Exact focused test, API dependency build, Biome, and diff check pass.

lalalune:
Closes #580.\n\n## Summary\n- explicitly opts this hermetic real-AgentClient integration fixture into the development replay-store boundary\n- installs an allow-only rate checker while preserving policy/call-count behavior above the proxy\n- configures the audit HMAC required by the real proxy path\n- snapshots and restores every changed environment binding\n- resets every mutable proxy handler dependency during teardown\n\nProduction enforcement remains fail closed.\n\n## Exact-head validation\n- agent-client E2E: 1/1 pass, 12 assertions\n- affected API dependency build: 24/24 packages\n- Biome and diff-check: pass

### DIFF
--- a/packages/api/src/__tests__/agent-client-e2e.test.ts
+++ b/packages/api/src/__tests__/agent-client-e2e.test.ts
@@ -23,8 +23,16 @@
 
 import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
 import { fileURLToPath } from "node:url";
-import { generateP256KeyPair } from "@stwd/auth";
-import { agentSigners, agents, closeDb, getDb, runPluginMigrations, tenants } from "@stwd/db";
+import { generateP256KeyPair, signAgentToken } from "@stwd/auth";
+import {
+  __resetAuditHmacKeyCacheForTests,
+  agentSigners,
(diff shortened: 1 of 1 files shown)


lalalune:

PR #582, opened 2026-08-19
test(proxy): isolate approval auth fixtures

lalalune:
## Summary
- explicitly opt the two approval lifecycle fixtures into proxy dev mode for their intentionally unsigned polling requests
- assert in each suite that the same unsigned request is admitted only in the fixture and still fails with 401 when `NODE_ENV=production`
- snapshot and restore every mutated environment value, reset the memoized audit HMAC key, and restore proxy handler hooks

## Security boundary
Production remains fail-closed: `NODE_ENV=production` overrides `STEWARD_PROXY_DEV_MODE=true`, which both suites now prove behaviorally. Request-signature tests are unchanged; these suites continue to focus approval expiry, atomicity, and single-use execution.

## Validation
- solo expiry/atomicity: 5/5
- solo lifecycle: 7/7
- isolated package runner filters: 2/2 files
- `bun run --cwd packages/proxy build`
- Biome on both files
- `git diff --check`

No production or workflow files changed.


lalalune:

PR #584, opened 2026-08-19
test(capabilities): isolate session broker proxy fixture

### LINKED ISSUE #583: test(capabilities): isolate session-broker proxy rate fixture
## Problem\n\nThe root `bun run verify` fails all three forwarding assertions in `session-broker-e2e.test.ts` with HTTP 429. The in-process proxy fixture reaches the real fail-closed Redis rate limiter without provisioning Redis or an explicit test checker. The suite also deletes environment values rather than restoring them and leaves proxy handler hooks installed.\n\n## Acceptance criteria\n\n- Opt the fixture into explicit proxy dev mode and install an allow-only request-scoped test rate checker.\n- Prove each forwarded request actually crosses that checker.\n- Restore every mutated environment value and reset proxy handler hooks.\n- Keep production proxy rate enforcement unchanged.\n- Pass the owning session-broker suite, affected build graph, Biome, and root verifier integration.

lalalune:
Closes #583.

## Summary
- explicitly opts the session-broker fixture into proxy dev mode
- installs an allow-only proxy rate checker and proves forwarded requests cross it
- proves unsigned requests remain rejected in production despite fixture dev mode
- asserts the signed proxy leg carries the Steward signature and timestamp
- restores every mutated environment value and resets all proxy handler hooks

## Security boundary
Production enforcement is unchanged. The checker is installed only through the existing test dependency seam, and the suite continues to exercise signed proxy requests.

## Exact-head validation
- isolated session-broker E2E: 6/6 pass
- dependency-aware build: 12/12 tasks
- Biome: pass
- diff-check: pass

Keep draft until exact hosted checks are green.

### DIFF
--- a/packages/plugin-capabilities/src/__tests__/session-broker-e2e.test.ts
+++ b/packages/plugin-capabilities/src/__tests__/session-broker-e2e.test.ts
@@ -53,11 +53,25 @@ const PROXY_URL = "https://proxy.broker-e2e.test";
 const BROKER_HOST = "broker.cap-e2e.test";
 const CAP_NAME = "broker.session.render";
 const MIGRATIONS_FOLDER = fileURLToPath(new URL("../../drizzle", import.meta.url));
+const TEST_ENV_KEYS = [
+  "STEWARD_PGLITE_MEMORY",
+  "STEWARD_MASTER_PASSWORD",
+  "STEWARD_JWT_SECRET",
+  "STEWARD_PROXY_REQUIRE_REQUEST_SIGNATURE",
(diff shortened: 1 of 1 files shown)


lalalune:
Exact head 75cf21cf7b71fa3d93bd32d22b06b73f62e3076e was independently audited and is now contained unchanged in canonical security aggregate #585. Root and package-context owning suites pass 6/6 with 32 assertions, the dependency build is green, and Biome/diff-check are clean. Closing this duplicate so protected validation occurs once on the consolidated exact head.

PR #587, opened 2026-08-19
fix(security): consolidate audited fail-closed hardening

lalalune:
Closes #583.
Closes #557.
Closes #561.
Closes #564.
Closes #573.
Closes #575.
Closes #577.
Closes #580.

## Security consolidation

Consolidates the exact reviewed final heads of #566, #568, #569, #570, #571, #572, #574, #576, #578, #579, #581, #582, and #584 so protected-branch validation and review apply to one candidate instead of being invalidated by sequential merges.

### Included hardening

- reject ambiguous or duplicate signer bearer credentials with bounded canonical parsing
- add an all-or-nothing PostgreSQL RLS catalog gate with dynamic inventory, partition, role, and PostgreSQL 15/16 evidence
- make auth email publication atomic against stale and ordinary writers with bounded receipts
- bind Worker request and scheduled security configuration to immutable AsyncLocalStorage snapshots; per-event code no longer mutates isolate-global `process.env`
- resolve OAuth, email, passkey, phone, platform, request-signing KDF, local vault, and provider lifecycle secrets from the current immutable request snapshot
- require the explicit runtime KDF salt throughout those request-scoped custody paths while preserving byte-identical existing ciphertext derivation
- rotate process caches with random-salt HKDF identities over every behavior-affecting setting; no plaintext secret metadata or persistent cross-process fingerprint is retained
- require freshness for every production sensitive mutation
- require authorization signatures for production machine mutations; only an exact public auth bootstrap list plus verified user access JWTs on enumerated auth, `/user/**`, `/v1/users/**`, `/tenants/**`, and `/dashboard/**` browser surfaces are exempt, while machine auth routes, the same user JWT on `/vault/**` and other control-plane paths, and all agent JWTs remain fail-closed
- sign canonical CLI JSON and raw machine mutations with the distinct API request root; unsigned clients still send freshness
- keep API and governed-proxy signing roots cryptographically separate
- fail closed for request-scoped KMS, PKCS11, and external custody until request-owned provider clients and credentials exist; request-scoped local custody remains isolated
- add freshness headers in the TypeScript SDK auth/client and server-side dashboard auth proxy without exposing a signing root to browsers
- provision a distinct API machine-request signing root in canonical Compose, CLI init/doctor, provisioner, Worker, and deployment documentation
- replace webhook, transaction receipt, CORS, wallet-signing, and key-export source assertions with behavioral coverage, including a provisioned real-PostgreSQL wallet lock race
[...]

PR #592, opened 2026-08-19
fix(auth): close publication concurrency gaps

### LINKED ISSUE #564: fix(auth): close atomic email publication concurrency gaps

lalalune:
Closes #564.

## Summary
- run PostgreSQL guarded batch publication at `SERIALIZABLE` with bounded serialization retries, closing races with ordinary and rolling old-pod writers
- persist an opaque TTL-bounded per-issuance commit receipt so a magic-link or OTP publication can confirm its own commit after acknowledgement loss
- bind confirmation to both the exact receipt and still-active target; superseded credentials are revoked/fail closed
- preserve one absolute credential expiry through retries and reject the entire publication once any value deadline has elapsed
- enforce the preflight before any delete/write with `Date.now()` in Memory, Redis server `TIME`, and PostgreSQL `clock_timestamp()` after advisory locks
- treat Memory entries and guards as expired at the exact millisecond deadline (`>=`), matching Redis/PostgreSQL authority
- allow Memory `setIfNotExists` to replace an entry exactly at expiry, matching Redis/PG absence semantics
- re-check PostgreSQL time after all writes and roll back the transaction if row-lock contention crossed the deadline
- preserve the previously valid OTP or magic link when a delayed replacement misses its commit deadline
- run the real Redis deadline proof in both hosted integration workflows that provide a Redis service and `REDIS_URL`
- prove an expired Redis batch cannot delete the previously valid credential
- retain the shared empty-CORS-allowlist behavioral preflight correction

This clean survivor contains only the reviewed #564 publication work and excludes the rejected runtime/cache portions of closed #587.

## Exact validation
Exact head: `384b1ae8ab20f374ae0fd59ea32363636c4b3fa5`
Exact tree: `869cd4fc05c565d544c05aeb82dd8c03d8a10de4`
Exact base: `9239a728b9ca28175e00e2825f69d85f3e02856e`

- focused Memory boundary suite: 18 pass, 60 assertions
- full auth suite: 326 pass, 12 integration skips, 1,027 assertions
- real PostgreSQL publication suite: 7 pass, 35 assertions
- real Redis publication suite: 2 pass, 6 assertions
- auth dependency build: 27/27 tasks
- Biome on all changed TypeScript files: clean
- `git diff --check`: clean

Keep draft until exact hosted CI and independent review are green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -504,6 +504,9 @@ jobs:
         env:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
 
+      - name: Email challenge publication deadline proof (real Redis)
+        run: bun test ./packages/auth/src/__tests__/email-redis-publish.integration.test.ts
+
       - name: Vault custody transition race proof (real Postgres)
         run: bun test ./packages/vault/src/__tests__/custody-transition-race.real-pg.test.ts
         env:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -594,6 +594,9 @@ jobs:
         env:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
 
+      - name: Email challenge publication deadline proof (real Redis)
+        run: bun test ./packages/auth/src/__tests__/email-redis-publish.integration.test.ts
+
       - name: Vault custody transition race proof (real Postgres)
         run: bun test ./packages/vault/src/__tests__/custody-transition-race.real-pg.test.ts
         env:
--- a/packages/api/src/__tests__/middleware-security-hardening.test.ts
+++ b/packages/api/src/__tests__/middleware-security-hardening.test.ts
@@ -56,7 +56,6 @@ describe("middleware security hardening", () => {
     // an unset NODE_ENV must fail closed like production.
     expect(tenantCorsSource).toContain("function devWildcardAllowed()");
     expect(tenantCorsSource).toContain('env === "development" || env === "test"');
-    expect(tenantCorsSource).toContain("origins.length === 0");
     expect(tenantCorsSource).toContain("return c.newResponse(null, 403)");
     expect(tenantCorsSource).toContain("TENANT_ID_RE.test(tenantId)");
     expect(tenantCorsSource).toContain("MAX_CORS_CACHE_ENTRIES");
--- a/packages/api/src/__tests__/tenant-cors-preflight.test.ts
+++ b/packages/api/src/__tests__/tenant-cors-preflight.test.ts
@@ -27,6 +27,15 @@ describe("tenant CORS preflight", () => {
         tenantId: "cors-patch-tenant",
         allowedOrigins: [ALLOWED_ORIGIN],
       });
+    await getDb().insert(tenants).values({
+      id: "cors-empty-tenant",
+      name: "CORS empty tenant",
+      apiKeyHash: "cors-empty-tenant-hash",
+    });
+    await getDb().insert(tenantConfigs).values({
+      tenantId: "cors-empty-tenant",
+      allowedOrigins: [],
+    });
   });
 
   afterAll(async () => {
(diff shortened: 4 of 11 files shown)


lalalune:
BLOCKER on exact head `4de9533e95de7b25f6223c17f7b9802fd6cf4633`: the new absolute-expiry contract diverges across backends and currently reports a known-unusable credential as successfully issued.

- Redis correctly computes remaining TTL from atomic server `TIME` and returns false before any write when the deadline has passed.
- `MemoryBackend.publish()` writes the already-expired value and returns true.
- `PostgresBackend.publish()` likewise inserts/updates the past `expires_at` and returns true; the next read filters/deletes it.
- The new delayed-success tests explicitly encode this: an 80ms publish delay against a 40ms challenge makes `sendOtp`/`sendMagicLink` resolve successfully, then immediate verification is false. The real-PG test expects `publish=true` for a row already expired at commit.

That is not fail-closed issuance: the API tells the caller delivery succeeded even though it already knows the delivered credential cannot work. Make Memory atomically reject the whole batch before writes when any non-null deadline is expired. Make PostgreSQL perform the equivalent database-time check inside the serialized transaction before applying any entry (and return false with guards/credentials untouched). Keep Redis behavior as the parity target. Reverse the delayed-success EmailAuth tests to require `EmailDeliveryError`, and add no-partial-write assertions for all backends. Then rerun exact auth/Redis/real-PG/build/CI and update the stale body/head evidence.

lalalune:
Correctness repair pushed as exact head `9c96a30f1b191ba0846db7c6a809852072c08425` from unchanged remote `4de9533e95de7b25f6223c17f7b9802fd6cf4633`.

Memory and PostgreSQL previously returned publication success after a delayed batch crossed its absolute expiry, allowing EmailAuth to report send success even though immediate verification was false; Redis already rejected the same state. All backends now fail the atomic batch without writes after any non-null value deadline. PostgreSQL checks `clock_timestamp()` after acquiring its locks, avoiding transaction-start-time staleness; Redis validates server-time TTL before even the idempotent all-desired path. OTP and magic-link behavioral tests now require activation failure after expiry.

Exact validation: focused auth 64 pass / 249 assertions; live local PostgreSQL suite 4/4 / 26; live Redis suite 2/2 / 5; auth dependency build 7/7; Biome and diff-check clean. PR remains draft pending hosted exact-head checks and review.

lalalune:
Pushed fac50710cf4075a2c78cfd24afedc60647ddc03d to close the remaining PostgreSQL deadline race. A transaction now re-checks clock_timestamp() after its writes and throws a private sentinel so the full batch rolls back if row-lock contention crossed the absolute expiry. The deterministic real-PG test holds an unmodified credential row lock, proves the publisher is blocked after the guard phase, releases it after expiry, and verifies false plus both original values restored. Exact validation: focused auth 64/64 (249 assertions); real PG 5/5 (29); real Redis 2/2 (5); auth dependency build 7/7; Biome and diff check clean. Draft remains for hosted CI and independent review.

PR #589, opened 2026-08-19
test(auth): make OIDC JWKS transport hardening behavioral

lalalune:
Closes #588

## Summary
- construct the pinned OIDC JWKS byte transport from immutable HTTPS, lookup, and deadline dependencies
- replace source-substring assertions with hostile behavioral coverage for exact status/body propagation, redirects, declared and streamed response bounds, deadlines, aborts, synchronous construction failures, request/response failures, late events, and cleanup
- destroy request/socket state immediately for redirects and declared oversize responses; ignore late stream events after terminal rejection
- discard hostile bytes for HTTP 204/205 while retaining the 256 KiB transport bound
- replace the stale empty-CORS-allowlist source assertion with real PGLite preflight denial behavior
- bind the JWKS cache policy to an immutable request-local Worker environment snapshot, including overlapping request isolation and missing-binding behavior
- attach concurrent boot rejection assertions before awaiting either request so neither rejection can surface as unhandled

## Exact-head validation
Exact head: `f6e7b114f15f888198652f5307a09c16faf67fc0`
Exact base: `9239a728b9ca28175e00e2825f69d85f3e02856e`

- Worker boot/runtime: 22/22, 59 assertions
- OIDC transport + shared runtime environment: 18/18, 78 assertions
- tenant CORS preflight: 3/3
- middleware security hardening (isolated): 8/8, 68 assertions
- `@stwd/api...` dependency build: 24/24 packages
- Biome on all 7 affected runtime files: pass
- `git diff --check`: pass
- mutation proof: removing the real `worker.fetch` runtime-environment wrapper makes the nested Worker test fail with `first snapshot: undefined` instead of the request-local `60000`

The prior PostgreSQL failure was the stale CORS source assertion replaced directly by this head; there is no remaining dependency on closed PR #587. Keep this PR draft until every required exact-head hosted check completes green and an independent write-access reviewer approves it.

PR #594, opened 2026-08-19
fix(worker): defer auth config capture until hydration

### LINKED ISSUE #561: Keep Worker security validation synchronous with binding hydration

lalalune:
Closes #561.

Depends on #589 at exact head `f6e7b114f15f888198652f5307a09c16faf67fc0`.

## Summary
- preserve the dependency-light JWT leaf import so Worker boot validation does not evaluate the auth barrel before bindings are hydrated
- construct immutable per-invocation runtime and JWT authorities from Worker bindings before any yield, preserving the outer general runtime snapshot and inner prevalidated JWT authority
- default Worker JWT authority to a conservative production posture when `NODE_ENV` is omitted
- resolve identity-token issuer and discovery base through one request-scoped canonical URL resolver
- require an explicit canonical HTTPS `STEWARD_IDENTITY_JWT_ISSUER` or `APP_URL` in production; never fall back to `steward` or a request Host
- return sanitized 503 responses when identity token/discovery configuration is unavailable
- declare production validation in the default, staging, and production Wrangler environments
- provide an ignored, explicit development override template so local `wrangler dev` remains operable
- document per-environment secrets and identity bindings, plus the exact locked-Wrangler bundle evidence
- cover hostile overlapping Worker authorities through real identity-token, root/tenant JWKS, and root/tenant discovery routes under distinct RSA/EC configurations
- prove canonical missing/short production JWT roots fail before database selection
- retain the shared tenant-CORS explicit-empty-allowlist behavioral regression unchanged

This stack contains only #594's unique work above #589. It excludes the rejected request-config/cache rewrite from #587.

## Exact validation
- `bun install --frozen-lockfile`: pass
- identity token/JWKS/discovery suites: 8 pass, 0 fail, 64 assertions
- Worker boot/runtime suites: 26 pass, 0 fail, 78 assertions
- full auth package: 324 pass, 3 skip, 0 fail, 1027 assertions
- CORS middleware/preflight suites: 11 pass, 0 fail, 81 assertions
- API dependency build: 24/24 packages
- Wrangler 4.105.0 dry run: default, staging, and production all pass with `NODE_ENV=production`; 7046.63 KiB raw / 1894.87 KiB gzip
- Biome on all 11 changed TypeScript/JSON files and `git diff --check`: green

Exact head: `a07a31a73e476991b53473d1ea42d59359f63109`
Exact tree: `ddc3c5718108e52b93668aecdfda6d9b564e0b43`
Exact dependency base: `f6e7b114f15f888198652f5307a09c16faf67fc0`

Keep draft until #589 is merged, exact hosted CI is green, and independent review approves the final tree.


lalalune:
Blocking exact `e7fe64d9`: `getAgentTokenExpiry()` still defaults to `process.env.AGENT_TOKEN_EXPIRY`. The new test mutates bindings sequentially, so it proves import-time capture is gone but does not prove request isolation. Overlapping Worker requests can hydrate different globals and an agent-token mint after an await can read the other request expiry. PR #589 now adds the request-local runtime environment seam and also changes `worker.ts`; merge/restack in that order, resolve the Worker conflict, make the default use `runtimeEnvironmentValue("AGENT_TOKEN_EXPIRY")`, and add a barrier-based overlap test with distinct expiries plus a missing-binding/global-poison case. Keep draft until the rebased exact head is green/reviewed.

lalalune:
BLOCKER on exact head `0a5cdb902bfb57b5e142fed5ee81a9e2befecfbb`: canonical Wrangler production posture never sets `NODE_ENV=production`, while `workerJwtEnvironment()` intentionally reads only the invocation binding. With the checked-in production binding shape and `{ STEWARD_JWT_SECRET: 'short', DATABASE_DRIVER: 'bogus' }`, the Worker only warns about the weak JWT secret and proceeds to `WORKER_DATABASE_DRIVER_UNSUPPORTED`; it does not fail boot on the weak signing root.

Please add an explicit production posture to default/staging/production Wrangler bindings as appropriate (or make Worker production validation independent of ambient NODE_ENV), and add a behavioral boot test derived from the canonical binding shape proving missing/short JWT authority fails before database selection. Keep local-development semantics explicit. Also document the optional identity-JWT bindings and `AGENT_TOKEN_EXPIRY` in the Wrangler/operator comments if they remain supported.

This PR also cannot land before #589. Exact merge-tree against repaired #589 `f6e7b114f15f888198652f5307a09c16faf67fc0` has two content conflicts: `packages/api/src/worker.ts` and `packages/api/src/__tests__/worker-boot-validation.test.ts`. Land #589 first, then rebase #594 while preserving both the outer general runtime snapshot and inner prevalidated JWT authority snapshot, plus both hostile overlap/missing-key test sets. Rerun the combined exact build/auth/Worker/CORS/hosted matrix after that restack.

lalalune:
Restacked onto #589 exact `f6e7b114f15f888198652f5307a09c16faf67fc0` and force-updated with lease from prior live head `50831e51a9524448fb56b74fb85f39e12655807d`. Final candidate is `a07a31a73e476991b53473d1ea42d59359f63109`; remains draft pending #589, hosted CI, and independent review.

PR #595, opened 2026-08-19
test(api): replace agent audit source scans

### LINKED ISSUE #500: test(api): replace agent audit source scans with route and rollback behavior

lalalune:
Closes #500

## Summary

- replace agent-audit source scans with executable route and database coverage
- serialize tenant-admin and scoped-platform agent deletion on the tenant agent row, revalidate inside the audited transaction, require one deleted row, and append the required completion audit atomically
- acquire that deletion fence before token revocation; tenant and platform token mint paths take the same row lock, preventing an iat-newer token from escaping after the revocation cutoff
- refuse deletion while an upstream lease is nonterminal or still retains token material; preserve secret-free terminal lease rows and append-only lifecycle evidence after agent and workspace authority are removed
- install PostgreSQL parent-row writer fences and direct-delete guards for leases, pending proxy work, enabled secret routes, active capability grants, provider-action bindings, and provider-action intents, including rolling old-pod deletion safety
- refuse unresolved transaction/provider/proxy execution instead of destroying outcome-unknown evidence; detach and retain resolved binding-backed and legacy/recovery intent-only provider evidence
- serialize both agent and platform tenant deletion against late lease/provider/route/capability/intent writers; writer-first deletion fails cleanly before revocation and delete-first late authority fails after parent removal
- preserve migration 0109/0110 journal adjacency and plugin capability lifecycle isolation
- align the enrollment audit-injection fixture, PostgreSQL operator cleanup, and platform token-helper assertion with the added locking transaction and retained-evidence lifecycle

## Exact-head validation

Exact head: `365b412064eb11bf7f1b8e3ddea7387d95d739e1`

Exact tree: `3c30465f30de986d77437a932fcd0d2178d8021c`

Exact base: `9239a728b9ca28175e00e2825f69d85f3e02856e`

- fresh PostgreSQL 16 deletion/evidence/writer-race suite: 19/19 tests, 101 assertions
- real-Postgres operator-transfer concurrency: 4/4 tests, 17 assertions
- focused PGLite enrollment/deletion/platform aggregate: 40 pass, 9 real-Postgres-only skips, 368 assertions
- focused migration suites: 4 pass, 1 environment-gated real-Postgres skip
- `@stwd/api...` uncached dependency build: 24/24 packages
- Biome on the changed TypeScript files: clean
- `git diff --check`: clean

The PR remains draft pending fresh hosted exact-head CI and independent review.


lalalune:
Exact-head deletion parity repair pushed as `61e89481ff74411ea4d747f1c75815a150026723` from unchanged remote head `aec2b76aa087ada49b2de9c2348b89a912851557`.

The behavioral suite now proves an owner session without recent MFA cannot delete an agent, append an audit, or create revocation state. The completion-audit failure fixture now seeds both a transaction and linked approval, proves both are restored after the failed delete, and proves the successful retry removes both. No production code or migration changed.

Exact validation: five agent suites 38/38 (284 assertions); migration/status 4/4 (7 assertions); CORS 3/3 (13 assertions); API dependency build 24/24; Biome and diff-check pass. PR remains draft pending exact-head hosted PostgreSQL CI and independent approval.

lalalune:
Independent exact-head audit completed for b01ed4294f174f844925ecb052dda5c0ecc39482. No new source blocker found. Every injected audit trigger now has a per-process random identifier, is scoped to the exact fixture tenant plus bounded action set, and is dropped with its function in finally; the five affected suites pass together (38/38, 286 assertions), confirming isolation. I also reproduced the workflow contract on a fresh local PostgreSQL database: the production journal migrator applied 90/90 entries through 0109, then the reconciliation suite passed 3/3 (6 assertions), including the `NOT NULL DEFAULT false` column contract and migration-count parity. Both ci.yml and pr.yml place this proof immediately after the production migrator in their PostgreSQL integration jobs. API dependency build 24/24, Biome/diff, and merge-tree are clean. Body evidence was refreshed; draft remains pending exact hosted PostgreSQL CI and review.

lalalune:
Exact-head follow-up on `87f36d15` reproduced a shared-PostgreSQL isolation failure under the PR workflow's `TEST_JOBS=2`: all five suites inserted the globally unique `tenants.api_key_hash` value `hash`, and the admin suite compared whole shared agent/key tables. Concurrent activity failed 3/5 suites (`tenants_api_key_hash_unique_idx` plus global snapshot drift).

Fixed in `470caa6a` atop the atomic-delete repair `7ff70d7e`: tenant fixtures now use unique hashes and rollback assertions are scoped to the suite tenant/agent graph. Exact combined-head evidence: real-Postgres `TEST_JOBS=2` runner 5/5; the same five suites 38/38 (286 assertions); 0109/migration status 5/5; affected API dependency build 24/24; Biome and diff-check clean; zero residual suite tenants, audit rows/heads, test triggers/functions, or temporary 0109 databases. Hostile concurrent unrelated-agent churn that failed 3 tests before the fix passes 10/10 after it. Keeping the PR draft pending hosted CI and independent review.

lalalune:
BLOCK at exact head `ef1d5505139fdb409c8160f7dcc10a36acc415a0`: both agent DELETE paths now falsely finalize upstream credential revocation. They update every bound `upstream_credential_leases` row to `status=revoked`, scrub `tokenCiphertext`/IV/tag/salt/hash, and append `lease.revoked`, but never exercise the sealed token through `revokeUpstreamLeasesForAuthority()` / the issuer revoker. A GitHub installation token (or future provider credential) can therefore remain valid upstream after the agent is deleted, while the only exact revocation handle has been irreversibly erased and recovery can no longer invalidate it. This contradicts `PROVIDER-MODES.md`, which requires provider-proven invalidation and preserves ambiguous outcomes as `needs_attention`. Dropping the agent FK in migration 0110 removes the previous fail-closed restriction and makes the false finalization reachable.

Do not merge this head. Safest bounded repair is to retain evidence but refuse agent deletion while any bound lease is nonterminal, instructing the caller to revoke through the existing exact-token authority flow; under the locked deletion transaction recheck that every surviving lease is provider-proven terminal/scrubbed before removing the agent. If automatic deletion is required, orchestrate the existing exact-token revoker before the delete and fail closed on ambiguous/provider/audit outcomes—never scrub or report `revoked` locally first. Add tenant-admin and platform behavioral cases for successful provider revocation, ambiguous failure preserving sealed `needs_attention`, and a hostile concurrent lease issuance/revocation race. The duplicated ~100-line deletion implementations should also become one shared service so their security lifecycle cannot drift.

lalalune:
Exact live head `573d8168` fixes the original false local revocation by invoking the exact-token issuer, but remains blocked by a two-phase authority race. The first transaction only disables/revokes rows that already exist; it does not durably mark the agent as deleting. During the external provider call the 0110 triggers still see an existing parent and permit new secret routes, pending/approved proxy requests, and leases; capability grants have no agent FK/fence. The final transaction rechecks only recoverable leases and `executing` proxy work, so a newly enabled route, active grant, or pending/approved request can survive the agent deletion as orphan authority (these tables have no agent FK). Add a durable deletion marker enforced by every authority writer, or use a single-transaction fail-closed deletion that refuses nonterminal leases and relies on the parent row lock/fence. Also remove the exported mutable `__setAgentDeleteIssuerFactoryForTests`; use immutable injected dependencies. Add a deterministic post-tombstone/pre-final creation race for grants, routes, pending requests, and leases before treating this head as safe.

lalalune:
Exact-head corrective published at `05514494b761a51fd463c0fc7be0c96d45a72717` on base `9239a728b9ca28175e00e2825f69d85f3e02856e`.

This closes the residual capability-authority lifecycle gap: optional-plugin grants are revoked inside the same audited agent-row-lock transaction, routes are disabled, and immutable plugin migration `0002_agent_grant_lifecycle.sql` serializes insert/agent-rebinding writers against deletion with a `FOR KEY SHARE` parent fence. Historical active orphans are fail-closed (route disabled before grant revocation). All SQL relation resolution is fixed to `public.*`, and the trigger has `SET search_path = pg_catalog, public`.

Exact local evidence: plugin migration PGLite 3/3 (11 assertions); deletion lifecycle PGLite 5/5 plus 2 real-PG-only skips (25 assertions); API and plugin direct TypeScript builds pass; Biome/diff-check clean. The earlier independent fresh-PG run on the same functional patch passed deletion 7/7 (36 assertions), including an existing grant and a deterministic writer blocked behind deletion then failing `23503`; hosted exact-head PG remains the acceptance gate. Keep draft pending full exact CI and independent review.

lalalune:
Metadata correction: the exact published head is `0551449461a074ad917b3ef40658b53facac3202` (same `05514494` short prefix and tree/evidence); the earlier comment expanded that short prefix incorrectly. The PR body is corrected.

lalalune:
Hosted PG diagnosis and exact repair: the owning deletion suite passed, but shared teardown left retained lease evidence, causing two later unrelated files to fail their global workspace cleanup with FK `23503`. Exact head `a7e7d88c0a3cf4fdeac894263abbbaa538635d07` now explicitly deletes fixture lease events/leases and optional capability-plugin tenant rows before the core tenant cascade. PGLite owning files pass together 18/18 plus 2 PG-only skips (124 assertions); API TypeScript, Biome, and diff-check pass. Hosted exact PG has restarted and remains required.

lalalune:
Final teardown scope at exact `e7969ecdb5ee77d5ae74365696c11c78edfe9eaf` also removes append-only capability invocation fixtures when the optional plugin tables exist; this prevents test tenant evidence from leaking across the shared real-PG runner. TypeScript, Biome, and diff-check remain green. Hosted exact-head matrix restarted.

lalalune:
Exact-head authority-fence repair: `496ac87d1c9253d07c46031841e6e528da91e89a` (parent `e7969ecdb5ee77d5ae74365696c11c78edfe9eaf`).

Fixed the remaining deletion race seams: the core trigger now uses a fixed `pg_catalog, public` search path and qualified `public.agents`; retained lease token/status, secret-route enabled state, and capability grant status/route bindings all reacquire the agent key-share fence on authority-bearing updates. A deterministic PostgreSQL test holds deletion, proves all three reactivation writers block, commits deletion, then proves all fail with `23503` and retained rows remain terminal/disabled/revoked.

Exact local evidence: fresh PostgreSQL 16 migration through 0110; migration proof 3/3; agent deletion lifecycle 8/8; combined 11/11, 60 assertions; plugin migration 3/3, 12 assertions; API/DB/plugin TypeScript builds; Biome; diff-check. Keep draft until the exact hosted PostgreSQL/full matrix and independent review are green.

lalalune:
Exact lint repair pushed at `5044d86071b2199d78a1d19a4ba0a8aeab0f519c` (parent `496ac87d1c9253d07c46031841e6e528da91e89a`). Root Biome now checks all 1,323 files clean; diff-check is clean. The prior provider-aws and SDK jobs failed in dependency installation before tests, while the lint job exposed the three formatter drifts corrected here. Hosted exact-head CI has restarted.

lalalune:
Exact-head parity/test repair: `da03932adede3345a1b346453abb4cf09bb33514` (parent `629d9d726b3357ca883a996ef97abed1105e6676`, tree `d0f36d4617e05f91c3696582f83adbba2e91d9b2`). Core 0110’s optional-plugin fallback now matches plugin migration 0002 and fences `secret_route_id` changes as an authority-bearing active-grant update. The real migration proof asserts the full trigger shape. It also now eagerly executes the hostile PostgreSQL reactivation query before passing a native Promise to Bun’s rejection matcher; the previous lazy postgres.js PendingQuery could wait until the 120s test timeout without sending SQL. Exact fresh PG16 journal-at-0108 upgrade proof passes 1/1, 14 assertions in 8.6s; DB dependency build 5/5, Biome, and diff-check pass. Hosted exact-head CI has restarted; keep draft.

lalalune:
Final exact amended head: `c9d72fafe9de1e9595b3fce1ae051f418518bbbd`, tree `4b6d3ad991bff74d776ca5102877eb7fa517eb8e`. In addition to the capability-route parity and eager hostile-query execution described above, all new core trigger targets are explicitly `public`-qualified, matching the fixed function search paths. The prior exact real-PG upgrade proof covered the same trigger definitions; this amendment only removes migration-time relation-resolution ambiguity. Hosted checks restarted.

lalalune:
Exact-head blocker at 2694711f8a271bf70faf299f6a224e677e79a4c8: platform tenant deletion does not inspect, revoke, or delete optional capability plugin rows. Migration 0110 now rejects cascading agent deletion while public.capability_grants contains an active grant, but DELETE /platform/tenants/:id only preflights leases/provider bindings before deleting secret routes and the tenant. Because capability_grants has no tenant, agent, or route FK, active grants make token revocation occur before the DB transaction later aborts with SQLSTATE 55000/500; terminal rows can also orphan after a successful tenant deletion. Add a pre-side-effect 409 or an atomic plugin lifecycle/cleanup inside the deletion transaction, define retained evidence semantics, and cover the real PostgreSQL platform route with active and terminal plugin data.

lalalune:
Second exact-269 blocker: tenant advisory locking creates an untested lock-order inversion in agent deletion. Authority-writer triggers take tenant advisory lock then agent FOR KEY SHARE. deleteAgentAuthority currently takes agent FOR UPDATE first, and when terminalizing an existing pending or approved pending_proxy_request its UPDATE trigger then tries the tenant advisory lock. A concurrent writer can hold tenant advisory while waiting on the agent, as deletion holds agent while waiting on advisory, producing a PostgreSQL deadlock/aborted deletion. Acquire steward_lock_tenant_deletion(tenantId) at the start of deleteAgentAuthority before agent FOR UPDATE, preserving the audit -> tenant advisory -> agent order, and add a deterministic real-PG pending-proxy-row/concurrent-writer test.

lalalune:
Exact hosted PostgreSQL failure at head 88462186c4da3d76793017c6a9db05c4784ef542 has three owning failures (269/272 API files passed): (1) agent-enroll-route expected the injected issuance-audit failure but instead reached token signing failed, so the fixture no longer establishes the intended post-sign audit boundary; (2) operator-transfer-postgres-concurrency cleanup deletes a tenant while its transaction remains unresolved and is correctly rejected by steward_guard_agent_delete with SQLSTATE 55000; cleanup must resolve/delete its own transaction evidence through the supported lifecycle; (3) platform-security-hardening has a brittle source-order assertion whose target is now absent. Repair these tests behaviorally and rerun the exact PG lane; a retry alone will not clear the deterministic failures.

lalalune:
Returning this PR to draft despite the green hosted matrix: exact head still has two source-confirmed lifecycle blockers under review—platform tenant deletion does not retire/preflight optional capability grants, and agent deletion can invert tenant-advisory/agent-row lock order through pending proxy cleanup. Green checks do not cover those interleavings. Keep draft until both are repaired with deterministic real-Postgres route/race evidence and exact-head CI reruns.

PR #593, opened 2026-08-19
docs: describe current repository invariants

lalalune:
Closes #557.

## Summary
- replace change-history narration in current tests, scripts, and operator docs with present invariant descriptions
- remove obsolete review scratch artifacts and rename retained mutation scripts by current responsibility
- replace brittle provider-trust source scans with rendered behavior and mutation proofs
- preserve legitimate temporal domain language, immutable migration history, and all production runtime strings/metadata byte-for-byte
- state governed dispatch honestly as a single-winner claim with durable terminal or outcome_unknown replay

The approval-loader runtime correction is intentionally split into draft #598; this PR remains wording and evidence only.

## Exact validation
- all runtime/test/docs sources outside the two package changelogs are byte-identical to the independently accepted pre-regression tree at `e890d51f43cd2908d67ca8a3e4fbaeb71508ebd6`; the only additive delta records the already-present SDK/React invariant-only changes in their Unreleased changelogs
- transpiled changed production TypeScript with comments removed: byte-identical to exact base
- focused CLI/provider/security-surface suites: 26/26, 20,415 assertions
- root lint: 1,321 files green on the prior semantic head
- root typecheck/build: 36/36 packages uncached plus web TypeScript on the prior semantic head
- provider trust UX: 9/9, 25 assertions
- provider-authority mutations: 10/10 killed
- focused claims/contract/adversarial suites: 11/11, 44 assertions
- exact changed-file Biome and diff-check green
- public-package metadata contract: 5/5, 16 assertions; exact base/head validation green

Landing order: #589, then #591 preserving its stronger CORS behavior, then #593, then #598 (which is based on this PR).

Exact head: `fb5bc4ee151738db51a4919fc7c3c74ca413f0c3`
Exact base: `9239a728b9ca28175e00e2825f69d85f3e02856e`

Keep draft until exact hosted CI and independent review are green.

PR #591, opened 2026-08-19
test(security): isolate fail-closed proxy fixtures

lalalune:
Closes #573.
Closes #575.
Closes #577.
Closes #580.
Closes #583.

## Summary
- provision the production-required audit HMAC and explicit development replay boundary only inside hermetic forwarding fixtures
- install load-bearing allow-only rate-check dependencies for capability, session-broker, AgentClient, and proxy-client paths
- prove production unsigned/rate enforcement remains fail closed while real routing, injection, scrubbing, policy, and call-count boundaries stay exercised
- replace the obsolete tenant middleware source assertion with byte-identical denial behavior and real empty-allowlist CORS behavior
- snapshot and restore every security-relevant environment binding, then run each capability test file in its own bounded package-owned process so mutable fixtures cannot cross files
- discover every TypeScript test convention recognized by CI inventory and cover discovery, timeouts, signals, descendant cleanup, stream failures, and bounded diagnostics behaviorally

## Exact-head validation
Exact head: `29a9ac4443e565b531b0c0f58870f222c39a644f`
Exact base: `9239a728b9ca28175e00e2825f69d85f3e02856e`

- isolated-runner adversarial self-tests: 8/8, 40 assertions, three consecutive full passes
- complete plugin-capabilities package runner: 15/15 test files passed
- affected API package-owned isolated runner: 3/3 files passed
- auth middleware: 3/3; proxy-client E2E: 3/3
- plugin-capabilities dependency build graph: 12/12
- CI test inventory: all 47 test-bearing targets covered
- Biome and diff checks: green

All 63 substantive exact-head hosted checks are green (Mintlify skipped). The sole remaining merge gate is independent write-access approval.

PR #598, opened 2026-08-19
fix(web): stabilize approval queue loading

### LINKED ISSUE #599: fix(web): reload approval queue on authenticated tenant rotation
The approval queue loader must be stable across unrelated renders without retaining the previous tenant epoch.

Acceptance criteria:
- load exactly once when an authenticated tenant becomes ready
- reload exactly once when activeTenantId changes
- install the rotated tenant token before descendant queue effects issue requests
- discard late results and action completions from the prior tenant
- preserve retry behavior and use the current Steward client
- behaviorally cover tenant A to tenant B rotation in the rendered dashboard

Draft PR #598 implements this boundary with layout-phase token installation, tenant-keyed queue state, and generation fencing. Its browser regression asserts exactly one GET per tenant epoch and a delayed tenant-A response that cannot replace tenant B.

lalalune:
Closes #599

## Summary
- bind approval list and decision requests to an immutable request-local SDK client created only when the access-token session tenant exactly matches the active tenant
- keep approval review routes outside the optional wallet-provider wrapper, so rejected, delayed, or never-settling wallet chunks cannot block or remount non-wallet authorization UI
- key rendered state to the active tenant and user session, clear prior-session rows/actions immediately, and generation-fence delayed list and approve/reject completions
- install rotated compatibility-client JWTs in the ancestor layout phase as defense in depth and enable the existing inline tenant switcher
- preserve the behavioral empty-tenant-CORS-allowlist regression and the full #593 trust/redaction/non-enumeration assertion union

## Stack order
1. Merge wording-only #593 after its exact-head checks and review complete.
2. Restack #598 patch-identically onto the resulting `develop`.
3. Prove the unique patch equivalent and rerun the exact restacked #598 head before merge.

## Exact validation
- exact head: `b6eb55517f4785f40872ddac377f86198d6e30fe`
- exact tree: `2e22355b10bca886b5bc2eb20bec4d1d1e7c5fc3`
- exact base: `fb5bc4ee151738db51a4919fc7c3c74ca413f0c3` (#593)
- provider trust Playwright Chromium file: 11/11, including the rendered never-settling wallet-chunk boundary, #593 trust/redaction/non-enumeration coverage, tenant credential rotation, later remount refresh, delayed-response suppression, same-tenant user rotation, and stale action fencing
- provider runtime unit suite: 7/7, including rejected, never-settling, and late-resolving wallet-chunk fallback
- clean-worktree `@stwd/web...` dependency graph and production Next build: 8/8
- web TypeScript: passed
- shared behavioral tenant CORS correction is retained byte-for-byte from the audited stack

Keep draft pending exact hosted CI, independent review, and the ordered #593 integration above.

### DIFF
--- a/packages/api/src/__tests__/middleware-security-hardening.test.ts
+++ b/packages/api/src/__tests__/middleware-security-hardening.test.ts
@@ -56,7 +56,6 @@ describe("middleware security hardening", () => {
     // an unset NODE_ENV must fail closed like production.
     expect(tenantCorsSource).toContain("function devWildcardAllowed()");
     expect(tenantCorsSource).toContain('env === "development" || env === "test"');
-    expect(tenantCorsSource).toContain("origins.length === 0");
     expect(tenantCorsSource).toContain("return c.newResponse(null, 403)");
     expect(tenantCorsSource).toContain("TENANT_ID_RE.test(tenantId)");
     expect(tenantCorsSource).toContain("MAX_CORS_CACHE_ENTRIES");
--- a/packages/api/src/__tests__/middleware-security-hardening.test.ts
+++ b/packages/api/src/__tests__/middleware-security-hardening.test.ts
@@ -56,6 +56,7 @@ describe("middleware security hardening", () => {
     // an unset NODE_ENV must fail closed like production.
     expect(tenantCorsSource).toContain("function devWildcardAllowed()");
     expect(tenantCorsSource).toContain('env === "development" || env === "test"');
+    expect(tenantCorsSource).toContain("origins.length === 0");
     expect(tenantCorsSource).toContain("return c.newResponse(null, 403)");
     expect(tenantCorsSource).toContain("TENANT_ID_RE.test(tenantId)");
     expect(tenantCorsSource).toContain("MAX_CORS_CACHE_ENTRIES");
--- a/packages/api/src/__tests__/tenant-cors-preflight.test.ts
+++ b/packages/api/src/__tests__/tenant-cors-preflight.test.ts
@@ -27,15 +27,6 @@ describe("tenant CORS preflight", () => {
         tenantId: "cors-patch-tenant",
         allowedOrigins: [ALLOWED_ORIGIN],
       });
-    await getDb().insert(tenants).values({
-      id: "cors-empty-tenant",
-      name: "CORS empty tenant",
-      apiKeyHash: "cors-empty-tenant-hash",
-    });
-    await getDb().insert(tenantConfigs).values({
-      tenantId: "cors-empty-tenant",
-      allowedOrigins: [],
-    });
   });
 
   afterAll(async () => {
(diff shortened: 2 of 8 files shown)


lalalune:
Source review at exact head `7253385d26c4a42931d2851d7050fd323eb1a95b`: the production fix is directionally correct. `loadPending` was recreated after every state update while the effect depended on it, so a successful/failed load could trigger another render and another request indefinitely. The stable callback removes that loop, retains explicit retry behavior, and the rendered request-count test covers the actual boundary.

Before readiness, please add exact head/base to the PR evidence and wait for the hosted web/build/accessibility matrix plus independent approval. Ordering matters: merge-tree against #593 exact `41b18bb1b5ded86beef84f21d21af7eb43ac246c` conflicts in `web/e2e/provider-trust-ux-a11y.spec.ts`. Land #598 before #593, then restack #593 and preserve both its provider-trust cases and this approval-queue request-count regression. Do not resolve by dropping either test set.

lalalune:
Independent exact-head audit found and repaired two test/stack blockers on the live `4789f142` head. The branch had reintroduced #591-owned API fixtures, creating false overlap and conflicts, and the approval browser test seeded a second independently generated JWT before asserting the request header. `ce0b51ec215455263bb0496e742216c511209985` restores those fixtures byte-for-byte to develop and asserts against the token actually installed in session storage. No production source changed in this repair.

Exact repaired-head evidence: provider unit 4/4 (10 assertions); production Next build passed; provider trust Playwright 6/6 against the exact standalone production build; changed-file Biome and `git diff --check` passed. Sequential merge simulation confirms #591 then #598 is clean/disjoint. Current #593 must be restacked last and deliberately resolve its three conflicts (`agent-client-e2e.test.ts`, `tenant-cors-preflight.test.ts`, `provider-trust-ux-a11y.spec.ts`). Draft remains pending hosted CI and review.

lalalune:
Exact stacked repair published at `031882e8940934222fa98e7041b10ac182c95e98` on current #593 head `fb5bc4ee151738db51a4919fc7c3c74ca413f0c3`.

The approval queue is now fenced by stable session identity `[tenantId,userId]`, not tenant alone or raw access token. A reachable near-expiry refresh that replaces user A with user B in the same tenant clears A state, reloads exactly once for B, and fences delayed loads/actions/toasts from A. The auth compatibility shim exposes the authoritative session user ID. Existing wallet-runtime timeout/redacted failure/late-resolution protections remain.

Identical functional-tree evidence before the changelog-only base rebase: production Next build PASS; provider-trust Chromium 10/10 including delayed same-tenant A→B; focused hostile case 1/1; Biome/diff clean. Exact rebased head Biome/diff clean. A local post-rebase unit retry could not resolve `react/jsx-dev-runtime`/`punycode` from this isolated worktree’s incomplete dependency links, so the restarted clean-checkout hosted matrix remains required. Keep draft pending exact hosted CI, #593 integration, and independent review.

lalalune:
Live follow-up `08c78c915d0d2da32a4fe8d17a9133e9bb9998cf` is accepted: its only functional-tree delta from the reviewed `031882e8` repair is a test-only TypeScript narrowing that stores delayed Playwright routes in stable `{ current }` cells. Assertions and production code are unchanged; it prevents optional-chaining from narrowing a callback-assigned variable to `never`. Hosted exact-head checks are restarted and remain required.

PR #590, opened 2026-08-19
fix(db): close request membrane reflection escape

lalalune:
Refs #488

## Summary

- guard every request-owned Drizzle capability returned through properties, descriptors, methods, callable results, transaction callbacks, and Promise/thenable fulfillment or rejection
- selectively membrane callable and client-like Promise results while leaving ordinary query rows, result objects, and driver errors usable after request completion
- classify `unsubscribe` handles used by postgres.js subscriptions and PGLite live queries, closing the remaining Promise-fulfillment escape
- deny mutating/reflection traps (`set`, `defineProperty`, `deleteProperty`, prototype mutation, extension changes) and re-check `has`/extensibility against the live request lease
- replace the stale CORS source assertion with behavioral empty-allowlist and PATCH preflight coverage

This is a corrective follow-up to merged #488. Native Promise fulfillment, driver thenable callbacks, and subscription handles can otherwise return a raw request-owned capability after the request closes. Regression coverage exercises a Neon-like Promise-resolved pool client, fulfillment and rejection paths, real PGLite `listen()` and live-query subscription handles, real PGLite transactions, and inert query/error values.

## Exact-head validation

Head: `abec18a997e0fdd99f2f4459f7ffe2cb7dabb31a`

Base: `9239a728b9ca28175e00e2825f69d85f3e02856e`

- `bun test --timeout 120000 packages/db/src/__tests__/request-database-context.test.ts packages/api/src/__tests__/worker-runtime-sentinel.test.ts packages/api/src/__tests__/middleware-security-hardening.test.ts packages/api/src/__tests__/tenant-cors-preflight.test.ts` — 44 passed, 246 assertions
- real PGLite transaction, `listen()` cleanup callable, and live-query `unsubscribe`/`refresh` handles are revoked after request completion
- ordinary synthetic rows, real PGLite query values, and native driver rejection errors remain usable after cleanup
- `bun run --cwd packages/db build`
- `turbo build --filter=@stwd/api... --force` — 24/24 packages, uncached
- `biome check packages/db/src/client.ts packages/db/src/__tests__/request-database-context.test.ts`
- `git diff HEAD^ --check`

Keep draft until the fresh exact-head hosted matrix and required independent review complete.

PR #601, opened 2026-08-19
feat(db): establish tenant RLS activation boundary

lalalune:
Refs #343

Stacked draft on PR #595. This remains a gated SEC-169 activation branch and must not close the issue or activate production RLS before the prerequisite stack lands and the final aggregate is rerun.

Current live stack (2026-08-19):

1. PR #590 exact `abec18a997e0fdd99f2f4459f7ffe2cb7dabb31a`
2. PR #595 exact `e7969ecdb5ee77d5ae74365696c11c78edfe9eaf` (still advancing under correction)
3. PR #601 exact `8f71c12765235dd1c54adb13c72cf4b2bde84a6d`, still on prior exact #595 base `5f1c5d2984008fa326a8506c711cfbbc06aa3e2f`

Do not final-restack #601 until #595 stabilizes. The #590 membrane delta is not duplicated here.

Implemented and repaired:

- contiguous migration `0111` after #595 migration `0110`, with 71 relations and 73 tenant policies
- named reachable Hono tenant propagation for tenant, user, proxy, platform, and background paths
- immutable tenant and optional user identity in ALS; nested mismatch rejection
- transaction-local `steward.tenant_id` and `steward.user_id` binding, verification, contamination cleanup, and pool-reset proof
- fixed-shape pre-context auth, OIDC, SAML, config, refresh, system/platform tenant, retention, platform, proxy, and background bootstrap functions
- atomic cross-tenant refresh rotation requiring the exact source tenant/user GUCs and target membership
- truthful reserved `platform` context for global identity and audit operations
- cross-tenant global membership projection plus owner-safe deactivation/deletion and all-tenant refresh-token revocation
- global deactivated-user and auth-KV retention through the restricted app role
- explicit checked-in 71-relation/73-policy activation inventory used by activation and rollback
- transactional role bootstrap with distinct app, migration-maintenance, and NOLOGIN definer roles
- conservative Worker posture: `neon-http` is accepted only in explicit development/test

Fresh validation for exact #601 head before the pending final restack:

- focused RLS/context/migration/platform/SAML suites: 33 pass, 329 assertions
- real PostgreSQL disposable Steward schema and exact independent `NOBYPASSRLS` app-role child process: 1 pass, 16 assertions
- app-role E2E covers platform global identity, tenant API-key, user JWT, SAML pre-auth, lifecycle revocation and platform audit, per-tenant job enumeration, and global auth-KV retention
- operator lifecycle covers migrations through `0111`, exact 71/73 inventory, 71/71 ENABLE+FORCE, rollback 0/0, reactivation 71/71, sole-owner fail-closed behavior, and migrator idempotency
- API and DB builds pass
- touched-source Biome and `git diff --check` pass

Still blocked before production activation/full #343 acceptance:

- PR #590 and PR #595 must land first
- after #595 stabilizes, #601 must be restacked onto its final exact head and the full aggregate validation rerun
- production activation remains an explicit operator action after those gates

Keep this PR draft and #343 open.


lalalune:
Exact-head review at `5dd001782a9d8058d5dd23a4b1eeb0b8cd2a7438`: keep this draft/incomplete. One additional activation blocker is currently missing from the body/runbook. Migration 0110 gives `user_push_subscriptions` global rows (`tenant_id IS NULL`) access only when `user_id = steward_rls.user_id()`, but `withTenantRlsTransaction()` binds only `steward.tenant_id`; no production path sets `steward.user_id`. The shipping `/user/me/push-subscriptions` GET/POST/PATCH routes run inside the verified session tenant transaction. After activation, global subscriptions are therefore invisible and global insert/update fails RLS; the existing GET also changes from all subscriptions for the user to only the current tenant. Current PGLite route tests cannot expose that because RLS remains disabled.

Before activation, extend the trusted verified-user context with the authoritative user id, set and verify both transaction-local GUCs on the same connection, explicitly define whether the endpoint returns global+current-tenant or all-membership subscriptions, and add an app-role real-Postgres route test covering global create/list/update plus hostile cross-user/cross-tenant access and pool reset. Keep #343 open.

The branch is also currently `DIRTY` against its moving #595 base and must be restacked after #595 lands; it must then absorb the accepted Worker/runtime-env lineage rather than retaining its older per-request `process.env` hydration tree.

lalalune:
Additional independent exact audit at `0795e4522f3ef240dee094977bff6d220c23bc6c`; keep draft and #343 open. Beyond the global push `user_id` gap already noted:

- Credential-lease request/recovery deadlines escape the tenant transaction. `runInternalJobForEachTenant` binds one tenant transaction, but `bindLeaseDeadline` replaces `input.db` with `withDatabaseDeadline`'s fresh handle; its audited transaction also uses that handle. The fresh connection has no tenant GUC, so app-role RLS sees zero/fails writes. On Workers, `neon-websocket` falls through the postgres-js/TCP deadline path, bypassing the request-owned socket. Deadline work must remain on the same bound transaction or mint/verify a new trusted tenant transaction on a supported request-owned transport.
- No production startup gate verifies `current_user` is the expected non-owner, NOSUPERUSER/NOBYPASS app role. An accidental migration-role DATABASE_URL receives the all-row maintenance policy and silently bypasses isolation. Make the catalog/role gate load-bearing before serving.
- `rls-activate.sql` enables only relations that already have a `steward_%` policy, then checks only physical `tenant_id` columns. A missing indirect policy (transactions, wallets, encrypted keys, policies, reputation cache, archive chunks) is invisible. Activation must compare exact OID-qualified catalog/policy/partition shape against the complete checked-in inventory and abort on any missing/extra relation or policy.
- The stacked Worker tree still predates #594's conservative production environment/runtime authority. Absence of NODE_ENV can permit a misbound neon-http deployment rather than failing closed. Restack onto the accepted #589/#594 lineage before final validation.

Minor stale current-posture text: 0111 migration line 365 and the rollout doc still call the policy installation migration 0110. Exact DB-focused audit passed 20/20 (72 assertions); the blockers are uncovered integration/activation boundaries, not failures of the current policy-shape unit tests.

lalalune:
Exact live re-audit at `e80797b9f1422e4185149ddfc15bd8506fe8c145`: keep draft and keep #343 open. The user-push `steward.user_id` context blocker is fixed, and the named 71-relation/73-policy inventory is a useful improvement. Remaining blockers: (1) credential-lease deadline/recovery still replaces the tenant-bound DB with a fresh connection whose audited transaction never sets/verifies tenant RLS GUCs; Worker neon-websocket also falls into a fresh postgres-js TCP path; (2) production startup still has no load-bearing current_user/SUPERUSER/BYPASSRLS/owner check, so a migrator/owner DATABASE_URL silently gets maintenance all-row authority; (3) inventory checks only counts, not exact policy name/command/roles/USING/WITH CHECK or partition tree, so same-count policy replacement/drift passes; (4) this stack still lacks #594’s immutable Worker runtime/JWT authority context and retains isolate-global hydration/capture; (5) 0111’s final schema comment still says policies were installed by 0110. New SECURITY DEFINER grants also make the shared app role able to call global destructive retention/platform functions directly; document that as an explicit compromised-app non-boundary or split maintenance/platform execute roles before claiming containment.

lalalune:
Exact-head security disposition for `8f71c12765235dd1c54adb13c72cf4b2bde84a6d`: **keep draft / do not activate**. Two release blockers remain beyond the known stacked-base drift.

1. Bootstrap grants the shared app role `EXECUTE` on all SECURITY DEFINER functions. The new destructive platform functions authorize through caller-settable session context (`steward.tenant_id = platform`), so a tenant SQL-capable app session can set that GUC and invoke the bypass-RLS owner to enumerate membership or deactivate/delete arbitrary users. This needs a non-forgeable database authority boundary (separate narrowly granted role/capability), not a caller-controlled GUC.
2. Platform deactivation/deletion now commits database mutation before external token revocation and required audit. Revoker or audit failure leaves the mutation committed; acceptance must restore fail-closed atomicity or an explicit durable saga whose incomplete state blocks use and is recoverable/auditable.

The branch also remains stacked on stale #595 (`5f1c5d2984008fa326a8506c711cfbbc06aa3e2f`) and must absorb the final #595 authority fences plus #594 Worker runtime lineage before aggregate review. #343 should remain open.

lalalune:
Additional exact `8f71c127` activation blockers to preserve in the redesign:
- credential-lease deadline handling replaces the tenant-bound DB with a fresh handle and opens its audited transaction without setting/verifying the tenant RLS GUC; Worker neon-websocket also falls into a fresh postgres-js TCP path instead of the request socket;
- production startup still lacks a current-role superuser/BYPASSRLS/table-owner/maintenance-role rejection gate;
- policy inventory checks counts, not exact policy name/command/roles/USING/WITH CHECK shape or partition descendants, so a same-count replacement can pass;
- this stack does not include #594’s immutable Worker runtime/JWT authority contexts.

Activation should remain explicitly incomplete until these are behaviorally proven with restricted app/platform/migration roles on real PostgreSQL.

lalalune:
Post-merge exact combined-head audit at `origin/develop` `90895ea25dd1cae4952f4409a04fcce09bbbc0df` confirms the previously reported activation blockers remain in the merged #594 + #601 lineage. Tracked with bounded repairs in #654.

Disposition: **do not activate RLS; keep #343 open**.

Confirmed on the exact merged source:

1. Credential-lease deadline binding still replaces the tenant-bound DB with a fresh deadline handle. That handle does not carry the tenant/user GUC transaction; with the merged #594 Worker request-owned `neon-websocket` runtime it also escapes into the postgres-js deadline branch rather than retaining the request socket.
2. Bun readiness and Worker startup still do not reject an unexpected/table-owner/SUPERUSER/BYPASSRLS/maintenance connection role or enforce the documented exact catalog gate. The activation inventory compares only relation policy counts, not exact names/commands/roles/expressions/partition closure, and its final closure check sees only relations with a physical `tenant_id` column.
3. Bootstrap still grants the shared app role blanket EXECUTE on BYPASSRLS-owned SECURITY DEFINER functions. Destructive platform functions authorize only through caller-settable `steward.tenant_id = platform`, so the app role can forge that context and directly deactivate/delete arbitrary users or revoke their refresh tokens.
4. Platform user deactivation/deletion still commits the database mutation before external token revocation and terminal audit. A revoker or final-audit failure leaves partial committed state. Tenant deletion retains the analogous post-commit external-revocation gap after #652.

#594 does resolve the old isolate-global Worker runtime/JWT authority concern; it does **not** resolve the database handle/GUC escape above.

Audit was read-only. No usable local `DATABASE_URL` or ready Docker daemon was available for a new restricted-role PostgreSQL run, so #654 requires reachable real-Postgres proofs against the exact repaired candidate before activation.

PR #603, opened 2026-08-20
fix(vault): preserve transaction ownership atomically

### LINKED ISSUE #602: Make transaction ID ownership atomic across concurrent recorders
`Vault.recordSignedTransaction` currently reads the existing owner and then performs an `ON CONFLICT DO UPDATE` that can rewrite `agent_id`. Two different agents concurrently recording the same previously absent transaction ID can both pass the read; the conflict loser then overwrites ownership.

Acceptance:
- enforce same-agent ownership in the atomic conflict statement and never rewrite the stored owner;
- reject a conflicting different-agent write when the conditional update returns no row;
- replace the source-order assertion with reachable persistence behavior;
- add a deterministic real-PostgreSQL two-writer race proving exactly one owner wins and the loser is rejected;
- wire the real-PostgreSQL proof into both PR and develop integration workflows.

lalalune:
Closes #602

## Summary
- replace the read-then-upsert transaction ownership check with one conditional conflict update
- never rewrite the stored agent owner; reject a conflict when the existing row belongs to another agent
- replace the source-order assertion with PGLite persistence behavior
- add and CI-wire a deterministic PostgreSQL two-writer race for an initially absent shared transaction ID

## Exact evidence
- head: `4ef0478e1b925159e1081337f4d72c5b51a4135f`
- fresh PostgreSQL 16 migration + PGLite/real-PG focused tests: 2/2, 10 assertions
- affected dependency build: 6/6
- Biome, actionlint, diff-check: pass

Keep draft until exact hosted PostgreSQL/full CI and independent review are green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -505,7 +505,7 @@ jobs:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
 
       - name: Vault custody transition race proof (real Postgres)
-        run: bun test ./packages/vault/src/__tests__/custody-transition-race.real-pg.test.ts
+        run: bun test ./packages/vault/src/__tests__/custody-transition-race.real-pg.test.ts ./packages/vault/src/__tests__/transaction-id-race.real-pg.test.ts
         env:
           DATABASE_URL: "postgres://steward:steward_ci@localhost:5432/steward_test"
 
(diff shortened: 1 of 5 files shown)


lalalune:
Hosted PostgreSQL note for exact `4ef0478e1b925159e1081337f4d72c5b51a4135f`: the new transaction ownership lane itself passed together with custody races (5/5, 41 assertions). The later aggregate API runner failed only the pre-existing `middleware-security-hardening.test.ts` source assertion (`origins.length === 0`), 272/273 files; #603 does not touch CORS/API middleware. Keep draft and restack after the owning behavioral CORS repair lands, then rerun the full exact matrix rather than retrying this stale base.

PR #608, opened 2026-08-20
fix(proxy): support signed broker GET requests

lalalune:
Refs #157

## Summary

- generate an idempotency key before signing every signed request, including GET/HEAD
- strictly validate the capability invoke envelope and bound query selectors to 32 string entries, 128-character keys, 2,048-character values, and an 8,192-character encoded query
- merge effective query selectors into policy-visible arguments and reject any `args`/`query` overlap mismatch, preventing one authorized selector from dispatching another resource
- reject `?` and `#` in authoritative secret-route/capability create and update validation, and fail closed at invoke time for legacy or externally-written unsafe capability rows
- add reachable broker proofs for signed GET replay plus hostile configured-query and fragment cases; unsafe legacy paths record a deny without reaching proxy auth/rate limiting or broker forwarding
- synchronize the proxy-client README and broker runbook with signed safe-request behavior, path-only capability specs, and actual session/MFA setup

## Validation

- vault route validator: 54/54
- capability package isolated runner: all 15 test files green; the long upstream-leases file exceeded the default 180s aggregate wall after 43 passing cases, then passed independently with a 600s wall
- proxy-client: 23/23, 72 assertions
- affected dependency build: 12/12 packages

## Stack and remaining acceptance

Stacked directly on #591 exact head `29a9ac4443e565b531b0c0f58870f222c39a644f`; its fixture environment restoration, rate-limit/audit-HMAC assertions, hook resets, and production fail-closed coverage are preserved alongside the signed GET tests.

This PR does not claim the live phase-1 demo. Issue #157 still needs a publicly reachable broker and proxy deployment, scoped token/capability/grant/policy setup, and live audit evidence.

Base: `29a9ac4443e565b531b0c0f58870f222c39a644f`
Head: `2f3daa9949f8177fb0bd479e88d700f19dd87359`

### DIFF
--- a/docs/guides/session-broker-integration.mdx
+++ b/docs/guides/session-broker-integration.mdx
@@ -230,7 +230,11 @@ The envelope is `{ args?, body?, query? }`:
 - `args` drive policy constraints (`argEquals` / `argMatches`) — they are **not**
   forwarded to the broker.
 - `body` is the JSON request body forwarded to the broker (for `POST`/`PUT`/`PATCH`).
-- `query` is appended to the forwarded URL as a query string.
+- `query` is a bounded string map appended to the forwarded URL. Its exact
+  selectors are also merged into the policy `args`; if a selector is repeated in
+  `args`, both values must match. Query maps are limited to 32 entries, 128
+  characters per key, 2,048 characters per value, and 8,192 encoded characters.
+  Unknown envelope fields and non-string selector values are rejected with 400.
 
 On an allowed call the proxy returns the broker's status and body **verbatim** (the
 injected credential already scrubbed). A `403` means denied by policy or missing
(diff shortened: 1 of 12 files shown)


lalalune:
Exact-head review of b888d9b5b9ba5bca91edd4386690b32091127232: keep draft / request changes. (1) Restack after #591; preserve its stronger environment restoration, rate-limit and audit-HMAC assertions, hook resets, and production fail-closed coverage. (2) The new GET query surface is agent-controlled but unbounded and not bound into capability intent: invoke.ts coerces arbitrary JSON through URLSearchParams while policy evaluates args, not forwarded query. An approved args.account=primary can dispatch a different query-selected resource. Validate and bound keys/string values and bind effective selectors into authorization; add type, unknown, overlong, and mismatch tests. (3) Update packages/proxy-client/README.md, which still says idempotency keys are only generated for mutating verbs. (4) Add a real replay test: assert the generated idempotency header, repeat the identical signed GET, require 409, and prove no second forward. Focused exact evidence is otherwise green: proxy-client 23/23, invoke 16/16, broker 5/5, signed-safe guard, Biome, and 12-package build.

lalalune:
Exact remediation pushed at 702567cc43e1c3a14210f089e7d3e86190acc43c. Query selectors are now bounded, type-safe, and policy-visible; args/query overlap mismatch fails with 400. The broker E2E proves a generated idempotency key, exact replay 409, and no second upstream forward. README/runbook are synchronized. #591 restack remains a required later gate; draft retained for exact hosted CI and independent review.

Issue #230, opened 2026-07-16
Gate D: execute the real GitHub sandbox run of the provider-authority matrix (§3)


0xSolace:
## Context
PR6 (#226) shipped the fake-transport E2E proof (Gate D CI half) but the **real GitHub sandbox run is not yet executed**. The docs/manifest correctly use PRE-real-proof claim wording (U8); this issue tracks earning the post-real-proof wording.

## What exists
- `scripts/provider-authority-sandbox.mjs` — fails closed on missing out-of-band creds (verified).
- `deploy/enterprise-reference/provider-github-sandbox.env.example` — placeholder-only, U7-compliant.
- Same governed code path as the fake, DEFAULT forwarder (`forwardWithVettedDns`), real terminal I/O.

## Acceptance criteria
1. Provision a **throwaway GitHub org + repo** with a dedicated GitHub App installation scoped to `issues:read` + `pull_requests:write` only.
2. Supply `GITHUB_APP_ID`/`GITHUB_APP_INSTALLATION_ID`/`GITHUB_APP_PRIVATE_KEY` + `STEWARD_*` out-of-band (CI secret / operator env), never committed.
3. Run `scripts/provider-authority-sandbox.mjs` reproducing **M02-M09/M14-M18** against real GitHub.
4. Prove the consequential write (`github.pr.comment.create`): on scripted timeout -> `outcome_unknown`, **no auto-repost**, reconciliation finds/definitively-misses the comment via the steward marker (§3.4). If GitHub cannot support safe bounded reconciliation -> record the **Gate D STOP** honestly.
5. Rate-limit (429/403 + Retry-After) classified as bounded `failed`, no blind retry (§3.5).
6. Record `artifacts/provider-authority/sandbox/` (5 artifacts + `provider-reconciliation.json`), secrets scrubbed.
7. Offline `verify-evidence-bundle.mjs --expected-key-fingerprint <fp>` PASSES clean / FAILS tampered.
8. Only AFTER the recorded artifact exists: update docs to post-real-proof wording; the prohibited-claim scanner must still pass.

## Size: M. Labels: authority, gate-d, security.
Anchor: PR6 spec §3, MASTER-PLAN.md:926-930 (Gate D go/stop).

PR #609, opened 2026-08-20
fix(sandbox): require live vault KDF salt

lalalune:
Refs #230

## Summary

- require `STEWARD_KDF_SALT` during provider-authority sandbox preflight so the isolated dispatch worker uses the same vault root derivation as the live deployment
- share one strict decoder between the production `KeyStore` and sandbox preflight, rejecting short, odd-length, non-hex, whitespace-padded, or partially decoded salts
- document the variable name only in the placeholder env template, with the existing value supplied out-of-band
- keep acceptance claims honest: real sandbox artifacts and a naturally observed GitHub 403/429 with `Retry-After` remain external and Gate D stays STOP without them

## Validation

- `bun test scripts/__tests__/provider-authority-sandbox.test.ts packages/vault/src/__tests__/kdf-domain-separation.test.ts` (26 pass, 82 assertions)
- affected dependency build: 9/9 packages at the preceding strict-decoder tree; the final commit only removes a logically redundant post-validation decoder parity branch
- Biome on the changed source and tests
- `git diff --check`

Node's hex decoder silently truncates odd nibbles and trailing non-hex input. The shared validator checks the complete even-length string before decoding. Valid generated deployment salts remain unchanged, and the now-provably redundant decoded-length recheck was removed rather than retained as defensive duplication.

No provider calls were made, no credentials were created or used, and no real-proof artifact is claimed.

Base: `9239a728b9ca28175e00e2825f69d85f3e02856e`
Head: `766900afadc55e260040e5fd18acbbad9b82159d`

### DIFF
--- a/deploy/enterprise-reference/provider-github-sandbox.env.example
+++ b/deploy/enterprise-reference/provider-github-sandbox.env.example
@@ -64,6 +64,11 @@ STEWARD_SANDBOX_SECRET_ID=
 # dispatch endpoint and no caller-forgeable dispatch context.
 DATABASE_URL=
 STEWARD_MASTER_PASSWORD=
+# The exact per-deployment salt used by the live vault. It must be valid hex
+# encoding at least 16 bytes (generate a new deployment salt with
+# `openssl rand -hex 32`). Supply the existing deployment value out-of-band;
+# using a different value cannot decrypt the pre-provisioned sandbox secret.
+STEWARD_KDF_SALT=
 STEWARD_EXECUTION_AUTH_SECRET=
 STEWARD_AUDIT_HMAC_KEY=
 STEWARD_AUDIT_SIGNING_KEY=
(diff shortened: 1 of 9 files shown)


lalalune:
Exact-head follow-up 91cf41ca2ecb0e31fc1e8bc2826859e9a46c1355 closes a configuration fail-open edge: Node Buffer.from(value, hex) silently truncated odd nibbles and trailing non-hex input, so the prior valid-hex claim was not enforced. Production KeyStore and sandbox preflight now require an even-length all-hex value before decoding. Exact focused tests are 25/25 with 78 assertions; affected dependency build 9/9; Biome and diff-check pass. Draft retained for restarted hosted CI and independent review.

lalalune:
Final exact head 766900afadc55e260040e5fd18acbbad9b82159d removes the redundant post-regex decoded-length branch from the shared salt decoder. Full-string, even-length, minimum-length validation makes that branch unreachable. Exact focused result: 26/26 tests, 82 assertions; Biome and diff-check clean. Draft retained for restarted hosted checks and independent review.

PR #620, opened 2026-08-20
fix(api): persist Solana recovery anchor before broadcast

### LINKED ISSUE #615: fix(api): persist Solana recovery anchor before external broadcast
Both Solana sign routes broadcast through Vault.signSolanaTransaction before inserting any durable transaction recovery row. A post-broadcast insert or audit failure returns a completed result but leaves no durable bookkeeping; the existing test only scans source ordering. Stage an approved or pending recovery anchor before external I/O, update it after broadcast/signing, preserve duplicate-retry-safe success semantics, and add reachable failure-injection tests proving the row survives post-broadcast bookkeeping or audit failure.

lalalune:
Closes #615

## Summary
- persist a canonical database-bound Idempotency-Key/request binding for every broadcast Solana auth mode
- use a token-owned execution lease so concurrent/retried requests cannot sign, submit, or account twice; expired takeover is CAS-fenced and stale callbacks cannot mutate the winner
- checkpoint the exact legacy/v0 signature and original recent blockhash before network submission
- distinguish proven preflight rejection from ambiguous submission, and provide authenticated evidence-bound Solana reconciliation with atomic audited transitions
- preserve durable recovery and successful/ambiguous responses through post-broadcast bookkeeping failures

## Exact evidence
- head: `51936a802297ffe2def031084ab70c4d7e8a3d16`
- parent/base PR head: `3b9e406ddd6e089bb41ace3a2e1ca9de111d7efd`
- mounted API recovery/idempotency/reconciliation: 6/6, 70 assertions
- idempotency middleware: 18/18
- Vault Solana behavior: 10/10
- full workspace typecheck: 36-package graph plus web PASS
- API dependency build: 24/24 PASS
- Biome changed files and `git diff --check`: PASS

## Merge posture
Keep draft pending replacement exact-head hosted CI, independent approval, and restack after #603 if that transaction-owner change lands first.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -556,6 +556,7 @@
         "@solana/web3.js": "^1",
         "@stwd/db": "workspace:*",
         "@stwd/shared": "workspace:*",
+        "bs58": "^6.0.0",
         "drizzle-orm": "^0.45.2",
         "viem": "^2.30.0",
       },
--- a/packages/api/src/__tests__/solana-recovery-anchor.test.ts
+++ b/packages/api/src/__tests__/solana-recovery-anchor.test.ts
@@ -1,0 +1,251 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "bun:test";
+import {
+  __resetAuditHmacKeyCacheForTests,
+  agents,
+  closeDb,
+  getDb,
+  policies,
+  tenants,
+  transactions,
+} from "@stwd/db";
+import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { ExternalBroadcastOutcomeUnknownError } from "@stwd/vault";
+import { eq, sql } from "drizzle-orm";
+import { Hono } from "hono";
+import type { AppVariables } from "../services/context";
+
+const TENANT_ID = `solana-recovery-tenant-${Date.now()}`;
+const AGENT_ID = `solana-recovery-agent-${Date.now()}`;
+const FROM = "7v54NWdBtkjuAFJrLGsS2SXnuk8nKam81mZJeeYxVFi9";
+const RECIPIENT = "6TcyBfPdBt1kjsvDZLzmBFnuMaLWiTaAt4RjUr9VA5YD";
+const SIGNATURE =
+  "4oL4p7QvN3UH7V5wMGZgW5PuzEk4A9LXLHk9RxAoKjDKuLbQBsfXN8kEvKfj5K1oEJa8wFF6RVp2h7pP9w2f51ZV";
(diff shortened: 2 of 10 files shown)


lalalune:
Exact-head blocker at `3b9e406ddd6e089bb41ace3a2e1ca9de111d7efd`: the durable broadcast checkpoint is not actually replay-bound for every allowed caller.

`requireBroadcastActionIdempotency` checks only that `Idempotency-Key` is present. The global idempotency middleware intentionally skips session-JWT mutations without recent MFA, while `requireSignerPermission` permits an admin session without recent MFA when tenant `vaultSigning` MFA is disabled. That reachable broadcast request therefore has no reservation/replay fence. Retrying a lost response with the same required key creates a fresh random recovery `txId`, signs/sends again, and duplicates DB/Redis spend accounting even if Solana itself deduplicates identical transaction bytes.

The new tests mount `vaultRoutes` directly, omit the global idempotency layer, and never replay the same key, so they cannot prove the PR body's non-retryable contract. Keep draft. Bind the durable Solana recovery row/result to the supplied idempotency key at the route/database boundary (or make the global reservation mandatory for this broadcast surface), then add a mounted full-app no-recent-MFA replay test proving the same key returns the same result/202, one recovery row, one send, and one spend record.

lalalune:
A second exact-head release blocker: the new Solana `outcome_unknown` state has no reachable reconciliation path.

The existing receipt poller skips any non-hex `txHash` and resolves only EVM RPCs, so a base58 Solana signature is permanently skipped. The admin lifecycle route cannot promote this row either: its broadcast transition accepts only signed/broadcast state, and confirmation accepts only broadcast state. Consequently every Solana send/preflight/confirmation rejection becomes a permanent `outcome_unknown` row and remains conservatively counted forever—even when preflight definitively rejected before submission. The returned 202 instructs callers to reconcile before retrying, but Steward provides no way to do that.

Add a Solana signature-status reconciliation path bound to the exact durable base58 signature, with guarded audited transitions for confirmed/finalized, explicitly absent/expired/blockhash-invalid, and still ambiguous outcomes. Prove the poller/reconciler handles each state and that only a definitive non-submission outcome makes retry safe. Until then the PR's recovery contract is incomplete; keep draft.

lalalune:
Exact remediation pushed as `51936a802297ffe2def031084ab70c4d7e8a3d16`. It closes both review blockers and the follow-on concurrency race: live attempts return processing without executing, expired takeover is token/CAS owned, every state transition is token-bound, and the strengthened barrier test runs stale A while replacement B is paused then proves B alone checkpoints/submits/finalizes. Hosted exact-head CI and independent review remain required.

PR #621, opened 2026-08-20
fix(web): surface agent account aggregation failures

### LINKED ISSUE #612: fix(web): surface agent account loading failures without duplicate balance fetches
The dashboard agent page catches getAgentAccount failures into null and renders empty/off state, then always performs a legacy balance request even after the richer account request succeeds. Add an explicit unavailable/error state, avoid the duplicate request when aggregate data is present, and add browser coverage for request count and failure behavior.

lalalune:
Closes #612

## Summary
- surface account-aggregation failures explicitly instead of rendering false empty/off state
- use aggregate native balance without a duplicate legacy request; invoke legacy balance only as bounded fallback after aggregate failure
- fence every async agent-detail load by generation and effect lifetime so delayed route A cannot mutate route B or finish B loading
- add browser request-count/error coverage and a mounted A-to-B delayed-response ownership regression

## Exact evidence
- head: `c37b3436f64b058ad3f4fab58f09f628e3a3362e`
- parent: `7629cf937ee9114ebeeee202c78b94a2702ab337`
- mounted route-transition test: 1/1, 7 assertions
- production Next build: PASS on the production delta before the test-only harness replacement
- web TypeScript and root prerequisite build graph 36/36: PASS
- Biome and `git diff --check`: PASS

## Merge posture
Keep draft until replacement exact-head hosted CI and independent approval. The previous Postgres failure is the inherited CORS source assertion repaired by #591; restack after that owning PR lands.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -640,6 +640,7 @@
         "@types/react-dom": "^19.0.0",
         "autoprefixer": "^10.4.0",
         "bs58": "^6.0.0",
+        "happy-dom": "^20.11.6",
         "postcss": "^8.4.0",
         "tailwindcss": "^3.4.0",
         "typescript": "^5.8.0",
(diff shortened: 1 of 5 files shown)


lalalune:
Corrected exact-head blocker at `7629cf937ee9114ebeeee202c78b94a2702ab337`: `loadAgent` has no request-generation or cleanup fence. A client-side route change from agent A to B can let A's slower account/fallback request overwrite B's state, and A's stale `finally` can clear B's loading indicator. Current E2E covers static success/failure only. Keep draft; add a delayed A→B navigation test and generation/abort ownership across all state writes.

lalalune:
Exact stale-load repair pushed as `c37b3436f64b058ad3f4fab58f09f628e3a3362e`. All post-await writes and final loading state are generation-owned. The regression now uses a mounted component with mocked `useParams` A→B and controlled promises—no private React internals—and proves delayed A cannot render or release B loading. Hosted exact-head CI and independent review remain required.

PR #626, opened 2026-08-20
fix(auth): reject malformed platform key scope configuration

### LINKED ISSUE #624: fix(auth): reject malformed platform key scope configuration
## Problem

On exact `origin/develop` `9239a728b9ca28175e00e2825f69d85f3e02856e`, `packages/auth/src/platform.ts` catches every `STEWARD_PLATFORM_KEY_SCOPES` JSON/schema error and returns an empty map. A valid platform key is therefore authenticated but every scoped request is reported as an ordinary 403. This masks a deployment configuration failure as missing authorization and can leave the platform control plane unexpectedly unavailable.

The same parser silently ignores non-array values and filters non-string scope entries, so partially malformed authorization configuration can be accepted without an operator-visible failure.

## Expected

When `STEWARD_PLATFORM_KEY_SCOPES` is present, require a JSON object whose values are arrays of strings. Reject malformed JSON or schema with a redacted configuration error; do not silently substitute an empty scope map. Preserve the current empty-map behavior only when the variable is absent/blank.

Add direct behavioral tests for malformed JSON, non-object JSON, non-array values, and mixed non-string arrays.

lalalune:
Closes #624

## Summary
- reject malformed `STEWARD_PLATFORM_KEY_SCOPES` JSON instead of silently replacing it with an empty scope map
- validate the complete key-to-string-array shape without filtering malformed entries
- exercise the mounted authentication boundary and prove the route is not reached under invalid configuration

## Exact evidence
- head: `fcc540dc6aabca28086bb3fe246f1e974c60c563`
- platform auth tests: 9/9 pass
- affected auth build: 4/4 tasks pass
- Biome and `git diff --check`: pass

## Merge posture
Draft pending exact hosted CI, restack after the root CORS repair in #589 if required, and independent approval.

### DIFF
--- a/packages/auth/src/__tests__/platform.test.ts
+++ b/packages/auth/src/__tests__/platform.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, describe, expect, it } from "bun:test";
-import { getPlatformKeyScopes, isValidPlatformKey } from "../platform";
+import { Hono } from "hono";
+import { getPlatformKeyScopes, isValidPlatformKey, platformAuthMiddleware } from "../platform";
 
 const ORIGINAL_PLATFORM_KEY = process.env.STEWARD_PLATFORM_KEY;
 const ORIGINAL_PLATFORM_KEYS = process.env.STEWARD_PLATFORM_KEYS;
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by #627, which preserves the same fail-closed parser, adds absent/blank compatibility coverage, and returns a deliberate redacted 500 from the mounted middleware. Closing this duplicate carrier without merge.

PR #632, opened 2026-08-20
fix(web): decouple webhook list and delivery refreshes

### LINKED ISSUE #622: fix(web): decouple webhook list and delivery refresh requests
## Problem

`web/src/app/dashboard/webhooks/page.tsx` defines its initial `loadData` callback with both `selectedId` and the filter-dependent `loadDeliveries` callback. The mount effect calls `loadData`, which selects the first webhook; that state change recreates `loadData` and invokes the effect again. The initial page therefore requests both `/webhooks` and the selected delivery list twice.

The same coupling causes selecting an endpoint to load its deliveries directly and then again through the effect, while every delivery-filter change also refetches the unchanged webhook configuration list. These are reachable browser requests in the existing webhooks dashboard E2E flow.

## Acceptance criteria

- Load webhook configuration independently from delivery selection/filter state.
- Fetch deliveries once when selection or delivery filters change.
- Do not directly duplicate the effect-driven delivery refresh on selection or deletion.
- Add behavioral browser assertions covering request counts.
- Keep user-visible request failures surfaced.

lalalune:
Closes #622

## Summary
- decouple webhook configuration loading from endpoint and delivery-filter changes
- fetch deliveries once per selection/filter epoch and fence stale responses
- hide stale delivery rows and summary counters synchronously until the active query owns the response
- preserve valid selection when unrelated endpoints are refreshed or deleted
- add deterministic delayed-response browser coverage for endpoint and filter transitions

## Exact evidence
- head: `0c348722620b32a48b65fa96ce4338d241b8d496`
- parent: `f6a37f792921d088f7d1509ab1757812c16c8959`
- web TypeScript: PASS
- Biome changed files: PASS
- `git diff --check`: PASS
- focused browser assertions were not completed locally because the host reached full disk and reused dependency links had unrelated Solana export skew; hosted exact-head browser/CI remains required

## Merge posture
Keep draft until the replacement exact-head matrix is green and an independent approval lands.

### DIFF
--- a/web/e2e/webhooks-page.spec.ts
+++ b/web/e2e/webhooks-page.spec.ts
@@ -78,6 +78,8 @@ test.describe("Dashboard webhook delivery history", () => {
       },
     ];
     let lastDeliveryQuery = "";
+    let webhookListRequestCount = 0;
+    const deliveryRequestCounts = new Map<string, number>();
     let createPayload: { url: string; events: string[]; description?: string } | null = null;
     let deletedEndpointId = "";
 
@@ -106,6 +108,7 @@ test.describe("Dashboard webhook delivery history", () => {
         await route.fulfill({ json: { ok: true, data: created } });
         return;
       }
+      if (route.request().method() === "GET") webhookListRequestCount += 1;
       await route.fulfill({
         json: {
           ok: true,
@@ -142,6 +145,9 @@ test.describe("Dashboard webhook delivery history", () => {
         });
         return;
       }
+      if (route.request().method() === "GET") {
+        deliveryRequestCounts.set("webhook-1", (deliveryRequestCounts.get("webhook-1") ?? 0) + 1);
+      }
       lastDeliveryQuery = params.toString();
       const status = params.get("status");
       const eventType = params.get("eventType");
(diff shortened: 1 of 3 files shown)


lalalune:
Corrected exact-head UI correctness blocker at `f4ff51a4fbaaf5ab1fc67520ad367ba00ce5d677`: the delivery effect fences stale network completion, but it retains the previous `deliveries` array during a new endpoint/filter request. Summary counters render from those old rows while the history shows a loading skeleton, temporarily attributing the prior endpoint/filter's delivery counts to the new selection. Clear rows on query ownership change or hide/reset counters while loading, and add a delayed-response endpoint/filter transition test. Required Postgres is also red from the inherited stale CORS source assertion, so this must restack after #589.

lalalune:
Exact-head repair pushed as `0c348722620b32a48b65fa96ce4338d241b8d496`. The active endpoint+filter owner key now prevents prior rows and summary counters from rendering during a new query. Deterministic delayed-response coverage asserts counters reset before the new response and remain bound to the active query. Local web TypeScript, Biome, and diff-check pass; exact hosted browser/CI and independent review remain required.

PR #638, opened 2026-08-20
test(api): centralize strict MFA age checks

### LINKED ISSUE #637: test(api): replace MFA timestamp source scans with shared behavior
## Problem

`mfa-future-timestamp-regression.test.ts` only checks that one matching age expression exists somewhere in each source file. A local MFA gate can regress or stop using the safe calculation while borrowing an unrelated expression later in the same file.

## Acceptance criteria

- Route every covered local recent-MFA predicate through one reachable shared validator.
- Exercise the validator at exact current, stale-age, and no-future boundaries, plus hostile far-future and malformed inputs.
- Preserve the mounted audit-gate regression for a hostile future timestamp.
- Remove only the weak per-file source scan.
- Keep focused tests, API build/type validation, and Biome checks green.

lalalune:
Closes #637

## Summary
- replace a weak per-file source scan with one strict runtime MFA timestamp validator
- route all 17 covered privileged MFA predicates through the shared no-future boundary
- behaviorally cover exact-now, exact max age, stale, future, hostile future, malformed/non-finite timestamps, and invalid windows

## Exact evidence
- head: `6ba1cf50cb392b55ce7bb0472f8172e144aee8c2`
- focused behavior: 3/3 tests, 14 assertions
- affected API dependency build: 24/24 tasks successful
- Biome on 19 changed files and `git diff --check`: pass
- frozen dependency hydration used `--ignore-scripts`; no native module was required for these checks

## Merge posture
Draft pending the current API stack, exact conflict-aware restack across affected route files, full hosted CI, and independent approval. The dirty shared checkout contains unrelated work and was not modified.

### DIFF
--- a/packages/api/src/__tests__/mfa-future-timestamp-regression.test.ts
+++ b/packages/api/src/__tests__/mfa-future-timestamp-regression.test.ts
@@ -14,28 +14,27 @@ describe("future-dated MFA timestamps", () => {
     expect(hasRecentSessionMfa(context as never)).toBe(false);
   });
 
-  it("every local recent-MFA helper has a non-negative age bound", async () => {
-    const files = [
-      "middleware/idempotency.ts",
-      "routes/agents.ts",
-      "routes/approvals.ts",
-      "routes/condition-sets.ts",
-      "routes/dashboard.ts",
-      "routes/erc8004.ts",
-      "routes/global-wallet.ts",
-      "routes/intents.ts",
-      "routes/policies-standalone.ts",
-      "routes/secrets.ts",
-      "routes/session-signers.ts",
-      "routes/tenant-config.ts",
-      "routes/tenants.ts",
-      "routes/user.ts",
-      "routes/vault.ts",
-      "routes/webhooks.ts",
-    ];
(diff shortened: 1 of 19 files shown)


lalalune:
Exact-head blocker at `6ba1cf50cb392b55ce7bb0472f8172e144aee8c2`:

`packages/api/src/routes/provider-approvals.ts:89-95` remains a separate privileged recent-MFA predicate. It checks that `sessionMfaVerifiedAt` is finite and rejects only when `Date.now() - mfa > 300_000`; a timestamp one day (or arbitrarily far) in the future therefore passes the route-level MFA gate because the computed age is negative. The new shared validator correctly rejects future timestamps, but this sensitive approval-detail route does not use it, so the centralization leaves a reachable duplicate with the exact unsafe boundary the PR is meant to eliminate.

Please route this predicate through `isRecentMfaTimestamp` (preserving the route's current error contract as appropriate) and add a mounted request regression proving a human session with a hostile future MFA timestamp gets 403 before approval detail is returned. Re-run the focused behavior suite and affected API validation at the amended exact head.

PR #639, opened 2026-08-20
fix(api): make tenant config audits atomic

### LINKED ISSUE #628: fix(api): make tenant-config mutation audits atomic under concurrency
## Problem

On exact `origin/develop` `9239a728b9ca28175e00e2825f69d85f3e02856e`, audited tenant-config mutations use a non-atomic snapshot / mutate / final-audit / whole-row restore sequence.

For example, `POST /tenant-config/:id/redirect-urls` snapshots the row, persists the changed URL list, writes the required completion audit, and on audit failure calls `restoreTenantConfigRow`. That restore starts a new transaction, deletes the current tenant-config row, and reinserts the old snapshot. The same pattern is used by OIDC providers, auth-abuse config, redirect URLs, app origins, test accounts, and the general config route. Related child-table restore helpers use the same delete-and-reinsert pattern.

## Reachable lost-update interleaving

Starting from config `C0` for one tenant:

1. Request A snapshots `C0` and persists change `A`.
2. A blocks in its final required audit.
3. Request B snapshots `A`, persists change `B`, completes its final audit, and returns success.
4. A audit fails and restores `C0` by deleting/reinserting the whole row.

The final state is `C0`: B returned success and has a completion audit claiming `B`, but B was erased by A rollback. Because the restore replaces the entire row, concurrent mutations to unrelated tenant-config columns are also vulnerable.

## Test gap

`packages/api/src/__tests__/tenant-config-audit-rollback.test.ts` only reads source text and asserts that snapshot/restore tokens occur in route bodies. It neither invokes a mounted route nor exercises a concurrent audit failure, so it currently blesses the unsafe rollback mechanism.

## Expected repair

- Perform each state mutation and its required completion audit in one database transaction on the same transaction client (the repository already exposes `appendAuditEventWithinTx`).
- Use the tenant audit lock/order consistently inside that transaction; do not repair failures by deleting and reinserting a stale snapshot.
- Cover at least one mounted tenant-config route with two real concurrent database connections: one request must fail its completion audit while the other commits, and the successful request state/audit must remain.
- Apply the same atomic boundary to the other tenant-config and child-table mutation helpers before removing the source-scan rollback assertions.

lalalune:
Closes #628

## Summary
- move tenant-config mutations and their required completion audits into one tenant-audited transaction
- re-read/recompute read-modify-write state under the same transaction and remove whole-row compensating restores that could overwrite concurrent success
- invalidate CORS cache only after commit
- repair policy-template application metadata to use the transaction-local inserted rows instead of a temporal-dead-zone outer binding
- add deterministic real-Postgres atomicity coverage with production correlation middleware and complete fixture cleanup

## Exact evidence
- head: `a17a751f9e91fd3dfc479a1ad8684a6e4cd8766e`
- parent: `f6d29027734211f7ee19d4fb489e635eae4c1647`
- real PostgreSQL atomicity: 1/1, 4 assertions; repeated cleanly with zero leaked tenant/user/audit/head rows
- mounted tenant-config API: 23/23, 81 assertions, including successful template apply and final audit
- related focused suites: 19/19, 203 assertions
- API dependency build: 24/24 PASS
- Biome changed files and `git diff --check`: PASS

## Merge posture
Keep draft pending replacement exact-head hosted CI and independent approval.


lalalune:
Exact-head test blockers at `f6d29027734211f7ee19d4fb489e635eae4c1647`:

1. The new real-PostgreSQL regression fails deterministically (`0/1`, about 9.9s) with `mounted route did not reach the blocked audit trigger`. Its bare Hono app mounts only `tenantConfigRoutes`; it sends `X-Request-Id` but never mounts request-ID middleware or sets `c.requestId`. Production audit writes therefore carry `request_id = null`, while the trigger waits only for the generated `failRequestId`, so the intended failure/interleaving is never reached. Mount the real request-ID middleware (preferred) or explicitly install the authoritative context in the fixture, then prove the actual concurrent rollback behavior.

2. Cleanup is incomplete for the shared PG lane. Deleting the tenant does not delete the independently seeded user, and audit events/chain heads intentionally lack a tenant FK/cascade. The test leaks its unique user and signed audit rows/head. Explicitly delete audit checkpoints/events/heads and the user in safe order, or run in a disposable isolated database.

The production transaction refactor otherwise looks structurally correct: all 25 required completion audits share their mutation transaction, stale whole-row restore helpers are removed, RMW paths re-read inside the transaction, and CORS invalidation occurs post-commit. Keep draft and rerun exact real-PG evidence after fixing the fixture.

lalalune:
Concrete post-failure DB evidence: tenant cleanup reached 0 rows, but the fixture left 1 user, 2 audit events, and 1 audit-chain head; both audit events had `request_id IS NULL`. The production middleware to mount is `correlation` from `packages/api/src/middleware/correlation.ts`, which populates `requestId` from `X-Request-Id`. Local artifacts were removed after inspection.

lalalune:
Exact repair pushed as `a17a751f9e91fd3dfc479a1ad8684a6e4cd8766e`. The PG fixture now mounts production correlation middleware so its trigger is reachable and cleans every exact user/audit/head artifact. The audit also found and fixed a real TDZ failure in successful policy-template apply, now covered through the mounted API with completion-audit persistence. Hosted exact-head CI and independent review remain required.

Issue #653, opened 2026-08-20
fix(api): safely terminalize signed-only transactions before agent deletion


lalalune:
## Problem

Merged #652 closed #649 after fixing tenant/capability deletion serialization, but it did not address the separate signed-only transaction lifecycle dead end. `deleteAgentAuthority()` blocks every `transactions.status = signed` row. Signed-only (`broadcast:false`) is a normal terminal API result, yet no mounted production route can safely retire that evidence, so any agent that ever returns a signed-only artifact becomes permanently undeletable. The current lifecycle test bypasses production by deleting the row directly through `getDb()`.

The generic transaction lifecycle failure events are not a safe replacement: they can relabel a signed artifact as failed without proving that the externally held broadcastable artifact is no longer usable, and their audit/mutation sequence is not the required atomic retirement contract.

## Acceptance

- Add an explicit, mounted, tenant-admin + recent-MFA signed-artifact retirement lifecycle; do not overload generic provider/execution failure.
- Preserve immutable structured evidence without storing or logging signed bytes.
- Lock/CAS the exact signed transaction and append the required audit in one tenant-audited transaction; audit failure must roll back.
- Fail closed unless chain-specific retirement is authoritative: e.g. Solana original blockhash/last-valid-height is durably recorded and confirmed expired; EVM requires authoritative nonce consumption/replacement or a separately documented high-risk abandonment policy. RPC/provider uncertainty must remain non-terminal.
- Concurrent broadcast/replace versus retirement must have one winner; deletion must remain blocked until the retirement commit and then succeed through the mounted API path.
- Reject generic `transaction.failed` / provider-error relabeling from `signed` unless equivalent authoritative evidence is supplied.
- Add reachable regressions for sign-only -> unexpired/unknown rejection -> authoritative retirement -> deletion, audit rollback, and concurrent lifecycle races. Remove the direct-DB deletion bypass as acceptance evidence.

Related: #649, #652.

lalalune:
[...]

PR #675, opened 2026-08-20
fix(api): support safe personal-tenant account deletion

NubsCarson:
## Scope
Adds the Steward lifecycle required for Google Play account deletion: personal-owner deactivation, ordered personal-tenant removal, identity deletion, and fail-closed preservation of shared sole-owner tenants.

## Safety
- shared sole-owner deactivation/deletion remains rejected
- personal tenant must be deleted before its user identity
- lifecycle invariants are enforced in SECURITY DEFINER functions and route error translation

## Validation
- platform hardening and identity graph: 42 tests pass
- migration status: 2 tests pass
- real PostgreSQL 18/RLS lifecycle: 1 test, 19 assertions pass through migration 0112
- API TypeScript build passes

Companion Eliza branch: elizaOS/eliza codex/android-account-deletion-staging-20260820.


lalalune:
Converted back to draft for exact review. This head has 52 checks pending and no independent approval, and migration 0113 collides with the already validated local #653 signed-artifact-retirement migration reservation. Do not merge until the lifecycle design and real PostgreSQL proof are independently audited; whichever change lands second must receive a new immutable migration number and regenerate the journal.

lalalune:
REQUEST CHANGES / keep draft. Exact source audit found the migration assumption "Personal tenants are single-owner by construction" is not enforced: the existing platform POST /tenants/:id/members route accepts any isValidTenantId, including personal-<userId>, and can add member/admin/owner rows to a personal tenant. 0113 then unconditionally skips the sole-active-owner guard for the matching personal ID, and DELETE /tenants/:id cascades every membership. Account deletion can therefore deactivate the named owner and delete a personal tenant that still contains another user, revoking/removing that user as collateral. Fix fail-closed at both durable boundaries: the deactivation exemption must require exactly the canonical owner membership and no other memberships; personal-tenant deletion must validate the same locked shape before destructive cleanup. Also reject future platform membership/invitation mutations against personal tenant IDs, while retaining the DB check for legacy/inconsistent rows. Add mounted and real-PG hostile tests with an extra member and extra owner proving 409/no mutation. Separately, this PR reserves 0113 while local validated #653 also uses 0113; whichever lands second must be renumbered through a new immutable migration and journal entry. Do not merge on the current 9-success/52-pending head.

lalalune:
Temporarily closing the current head because the confirmed extra-member/extra-owner personal-tenant deletion bug remains in source even though hosted checks are nearly green. A corrected local successor is in progress with mounted and real-PG hostile coverage. Do not reopen/merge b5407599; preserve its branch for comparison only.

lalalune:
Sequencing behind #680: the current base still has the 15-minute Postgres integration ceiling. After #680 merges, rebase this lifecycle change onto the exact tip and rerun the full protected matrix.

Issue #657, opened 2026-08-20
blocker(release): restore protected develop and main gates


lalalune:
## Confirmed live state

At 2026-08-20 08:23 UTC, both `GET /repos/Steward-Fi/steward/branches/develop/protection` and the corresponding `main` endpoint return HTTP 404 `Branch not protected`; the repository rules query for `main` returns an empty list.

PR #586 consequently merged `develop` to `main` while its exact PR matrix still contained proxy, plugin-trading, and CodeQL failures and without a recorded independent approval. The merged main SHA is `3161eb7b4d60c60c9a9177bff7219e9c3cd94ea9`.

## Required repair

- protect both `develop` and `main` with strict/up-to-date required status checks
- require at least one independent approval, dismiss stale approvals, require conversation resolution, and enforce for administrators
- pin the complete current required-check inventory to the expected GitHub App where supported
- prohibit force pushes and branch deletion
- verify a hostile draft/red/no-review PR cannot merge, then record the exact ruleset/protection response
- do not use the unprotected interval as acceptance evidence for the corrective release

This is an operational protection change; obtain repository-owner confirmation before applying if the expected required-context inventory or bypass policy is ambiguous.

lalalune:
Live consequence confirmed: main push CI run 32347880207 completed `failure` (`packages/proxy`, `packages/plugin-trading`) after unprotected PR #586 merged. The pending Cloudflare deployment run 32347880094 for exact main SHA `3161eb7b4d60c60c9a9177bff7219e9c3cd94ea9` was cancelled to prevent promotion of the red tree. Staging had already deployed the byte-identical develop tree twice because its workflow gated only on Docker; #659/#660 now own that separate workflow repair.

lalalune:
Protection repair completed and live-verified on both `develop` and `main`.

Exact current settings on each branch:
- 62 strict/up-to-date app-pinned required checks: 61 GitHub Actions (`app_id=15368`) plus the CodeQL aggregate (`app_id=57789`)
- one approval required
- stale approvals dismissed
- approval required after the most recent push
- conversation resolution required
- administrators included
- force pushes disabled
- branch deletion disabled
- no bypass restrictions configured

The protection API returns the same settings for both branches. Current hostile cases are blocked: draft, red/pending, and review-required PRs report `BLOCKED`/`REVIEW_REQUIRED` rather than merge eligibility. This closes the operational protection gap; the prior unprotected merge interval remains invalid as acceptance evidence.

[...]

PR #684, opened 2026-08-20
fix(api): make Solana recovery ownership atomic

### LINKED ISSUE #651: fix(api): anchor every Solana broadcast before external I/O
## Problem

Merged #620 (`8040b175`) persists a durable recovery anchor for the two `/vault/:agentId/sign-solana` paths, but two other production Solana broadcast families remain outside that state machine:

- `POST /vault/:agentId/actions/transfer`: SPL transfers call `Vault.signSolanaTransaction` without `onBroadcastPrepared`; native SOL transfers call `Vault.signTransaction -> solana.signSolanaTransaction`, which has no checkpoint callback at all. The route allocates a random action ID and its catch does not preserve `ExternalBroadcastOutcomeUnknownError` as recoverable Solana evidence.
- `POST /vault/:agentId/approve/:txId`: Solana approval execution calls `Vault.signSolanaTransaction` without `onBroadcastPrepared`. Its catch may persist the hash after an exception, but a process crash after RPC acceptance and before that update leaves no durable pre-I/O anchor.

A lost send response or crash can therefore still leave an accepted transaction mislabeled or permit later re-execution. The merged PR body’s “every broadcast Solana auth mode” claim is not true.

## Acceptance

- Route every native/SPL/parsed/blind/approved Solana broadcast through one durable request binding and execution-owner state machine before external I/O.
- Make the lower-level native SOL helper produce the deterministic signature and original recent blockhash before `sendRawTransaction`, invoke a required checkpoint, disable transport retries, and classify proven simulation rejection separately from ambiguous send/confirmation.
- Bind broadcast transfer IDs to tenant + agent + route + hashed Idempotency-Key + canonical request fingerprint; same key/different request must 409 and expired lease takeover must CAS-fence stale callbacks.
- Preserve each route’s action/approval metadata and replay response; reconciliation must accept the exact anchored action type and signature without resending.
- Add mounted concurrent, lost-response, crash-after-checkpoint, stale-owner, and post-bookkeeping-failure regressions for native transfer, SPL transfer, and approved Solana execution.
- Run exact PostgreSQL/PGLite/API/Vault suites, full affected builds, Biome, and fresh hosted CI before develop-to-main promotion.

### LINKED ISSUE #673: security: make Solana recovery ownership and accounting atomic
## Gap

Merged PR #663 closes the pre-I/O checkpoint gap, but exact merged-head review found remaining cross-transaction races:

- post-checkpoint success upserts and route bookkeeping can regress a concurrently reconciled confirmed row to broadcast;
- a lost owner can return replay success before spend aggregation, sponsorship, audit, and webhook effects are durably claimed, allowing cumulative policy undercount;
- approval transaction release and approval-queue reopening are separate writes, so a new owner can broadcast before a stale request resets the queue;
- definite preflight handlers do not surface a lost token/signature CAS and can return stale failure instead of the winner state.

## Acceptance

- Every post-checkpoint success mutation is token, signature, and expected-state CAS fenced; CAS loss returns exact winner replay and never regresses terminal state.
- Irreversible spend/accounting effects are durably idempotent and executed exactly once by either reconciliation or replay.
- Transaction reset, approval queue reopen, and required preflight audit are one atomic tenant-audited transition with exact queue and owner predicates.
- Non-approval definite preflight CAS loss returns winner replay.
- Mounted interleaving tests cover native sign, native and SPL transfer, native and serialized approval, post-finalize reconciliation, stale-owner accounting, and queue-reset takeover.

This is a corrective follow-up to merged #663; issue #651 closure must not imply these races are resolved.

lalalune:
## Summary

- fence every native, SPL, parsed, blind, transfer, and approved Solana broadcast with an exact execution-token/signature/status CAS so stale owners cannot regress a reconciled result or perform post-ownership side effects
- reset an approved transaction and its approval queue atomically, and replay the current winner when definite-preflight cleanup loses ownership
- make irreversible recovery effects durable and retryable: spend and aggregation counters use stable event identities and the original signed time, sponsorship is upserted, webhook delivery creation is deterministic, and required audit completion is transactionally marked
- keep recovery effects pending when configured Redis accounting is unavailable or throws, preventing a confirmed broadcast from being acknowledged with silently incomplete cumulative accounting
- preserve request-local native-price fallback configuration while using strict per-chain native decimals, including Solana lamports

## Validation

- exact head `dc283fd50b730bd9588b450b932db1fbfee83bd0` on develop `46d01b88f5ec85db1a3c02367182ae69da35aae6`
- isolated owning suites: 73 pass / 448 assertions across wallet actions, approval gates, recovery anchors, webhook dispatch, Solana policy derivation, Redis enforcement, Vault offline broadcast, and transaction CAS hardening
- disposable real Redis: 24 pass / 58 assertions across spend and aggregation trackers, including duplicate recovery events and delayed authoritative bucket placement
- `bunx turbo run build --filter=@stwd/api...`: 24/24 tasks passed
- Biome clean on all 15 changed files
- `git diff --check origin/develop...HEAD` clean

Closes #651
Closes #673

### DIFF
--- a/packages/api/src/__tests__/solana-recovery-anchor.test.ts
+++ b/packages/api/src/__tests__/solana-recovery-anchor.test.ts
@@ -2,6 +2,7 @@ import { afterAll, beforeAll, beforeEach, describe, expect, it, setDefaultTimeou
 import {
   __resetAuditHmacKeyCacheForTests,
   agents,
+  auditEvents,
   closeDb,
   getDb,
   policies,
@@ -10,8 +11,11 @@ import {
   transactions,
 } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
-import { ExternalBroadcastOutcomeUnknownError } from "@stwd/vault";
-import { eq, sql } from "drizzle-orm";
+import {
+  ExternalBroadcastOutcomeUnknownError,
+  SolanaBroadcastNotSubmittedError,
+} from "@stwd/vault";
+import { and, eq, sql } from "drizzle-orm";
 import { Hono } from "hono";
 import { idempotencyMiddleware, MemoryIdempotencyStore } from "../middleware/idempotency";
 import type { AppVariables } from "../services/context";
@@ -111,6 +115,7 @@ describe("Solana durable recovery anchors", () => {
       "solana-recovery-anchor-audit-hmac-key-with-more-than-32-bytes";
     __resetAuditHmacKeyCacheForTests();
     delete process.env.STEWARD_ALLOW_UNSAFE_SOLANA_BLIND_SIGNING;
+    delete process.env.REDIS_URL;
     await getDb().delete(transactions).where(eq(transactions.agentId, AGENT_ID));
   });
 
@@ -120,6 +125,7 @@ describe("Solana durable recovery anchors", () => {
     delete process.env.STEWARD_AUDIT_HMAC_KEY;
     delete process.env.STEWARD_ALLOW_DEV_SECRETS;
     delete process.env.STEWARD_ALLOW_UNSAFE_SOLANA_BLIND_SIGNING;
+    delete process.env.REDIS_URL;
     __resetAuditHmacKeyCacheForTests();
   });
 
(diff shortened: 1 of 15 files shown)


lalalune:
Temporarily closing and removing the remote carrier because repository protection has been externally deleted three times and other pending/no-review PRs are being auto-merged. Exact `dc283fd50` also remains blocked on semantic intent binding, successor takeover, SPL-vs-native accounting, and retryable reconciliation. The exact head and active local corrective work are preserved; reopen on a fresh protected carrier only after #657 is durably resolved.

PR #697, opened 2026-08-20
fix(auth): restore callback navigation fallback

### LINKED ISSUE #695: fix(auth): restore HTML callback fallback for navigation without Accept
## Problem

## Acceptance

## Local proof available

lalalune:
Closes #695

## Root cause

- A missing `Accept` header was normalized to zero HTML quality before the browser-navigation fallback was consulted, so `Sec-Fetch-Mode: navigate` incorrectly received JSON.
- The post-merge OAuth token-flow 500 was fixture-only: wallet creation completed, then session signing correctly failed closed because the rebased test still set legacy `JWT_SECRET` instead of mandatory `STEWARD_JWT_SECRET`. Required audit signing also needed an explicit test key.

## Changes

- consult the navigation signal when `Accept` is absent, while keeping non-navigation clients on JSON
- preserve explicit media quality, wildcard specificity, and `q=0` behavior from #691
- make the token-encryption suite standalone by installing PGLite before dynamically importing the mounted auth route and configuring current required signing keys
- mount successful browser-navigation and explicit-JSON OAuth callbacks through provider exchange, user creation, encrypted provider-token storage, personal wallet provisioning, session signing, and token-free redirect fragments

## Validation

- callback browser/OIDC suite: 17 passed, 112 assertions
- token encryption/provisioning suite standalone: 4 passed, 24 assertions
- OAuth client + nonce binding + OIDC code-flow: 65 passed, 247 assertions
- API dependency build: 24/24 packages passed
- Biome on all changed files passed
- `git diff --check` passed

Draft only. Do not merge until exact-head hosted checks and independent review are green.

### DIFF
--- a/packages/api/src/__tests__/auth-oauth-callback-browser.test.ts
+++ b/packages/api/src/__tests__/auth-oauth-callback-browser.test.ts
@@ -1,6 +1,7 @@
 import { afterAll, afterEach, beforeAll, describe, expect, it, mock } from "bun:test";
-import { closeDb } from "@stwd/db";
+import { accounts, closeDb, getDb, users } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { eq } from "drizzle-orm";
 
 const REDIRECT_URI = "https://app.example.test/callback?from=oauth";
 const ORIGINAL_FETCH = globalThis.fetch;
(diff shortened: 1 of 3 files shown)


lalalune:
Temporarily closing and removing the remote carrier because repository protection has been deleted three times within minutes and pending/no-review PRs are being auto-merged. The exact head `b6cfd2275` and local repair worktree are preserved. Reopen on a fresh protected carrier only after #657 is durably resolved and exact hosted checks can complete.

PR #705, opened 2026-08-20
fix(auth): restore navigation callback fallback

### LINKED ISSUE #695: fix(auth): restore HTML callback fallback for navigation without Accept

lalalune:
Closes #695\n\n## Summary\n- honor browser navigation fallback when Accept is absent\n- preserve explicit JSON and q=0 negotiation\n- keep callback errors non-reflective and token-free\n\n## Exact-head validation\n- callback browser suite: 17/17, 112 expectations\n- API dependency build: 24/24\n- git diff --check clean\n\nDraft pending protected CI and independent approval.


lalalune:
Closing/removing this externally republished carrier under the active #657 publication freeze. Exact `a5da7ec27` is preserved locally; no accepted review/hosted evidence exists yet.

PR #773, opened 2026-08-20
fix(ci): restore honest raw-signer and Docker gates

lalalune:
Post-merge corrective for #772.

Summary:
- classify the reachable approved native-SOL raw signer as a non-EVM branch instead of falsely calling it invariant-guarded
- tie each guarded inventory entry to its specific following raw signer and nearby runtime guard
- run the globally required Docker workflow for every pull request so app-pinned build-and-push can always report

Validation:
- git diff --check passes
- actionlint passes for .github/workflows/docker.yml
- exact raw signer count remains four
- local Bun tests were intentionally not run because the shared volume was below the safe disk floor; this draft must remain blocked until exact-head hosted checks, including build-and-push, pass and an independent reviewer approves

Operational note:
- develop classic protection was restored before this PR was opened: strict 62 app-pinned checks, one approval, last-push approval, stale dismissal, conversation resolution, admin enforcement, no force-push/deletion
- the deleted additive ruleset remains a separate provenance incident; historical rule-suite evidence shows classic protection currently enforces the same effective categories plus conversation resolution

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,5 +1,5 @@
 # Build and push Docker image to GitHub Container Registry (GHCR)
-# Triggers on pushes to develop + main branches and version tags (v*)
+# Builds every pull request and publishes pushes to develop/main or version tags (v*).
 #   develop -> :develop image (auto-deployed to STAGING by deploy-staging.yml)
 #   main    -> :main image (promoted to PRODUCTION manually via deploy-railway.yml)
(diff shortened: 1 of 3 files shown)


lalalune:
Closed unmerged for #657 containment after branch protections disappeared again and #776 merged with zero reviews. Exact head 4cdccc0c4829476013fee3cdac42de1f981124d6 is preserved locally; no source disposition is discarded.

PR #781, opened 2026-08-20
fix(trading): fail closed without durable rate limits

### LINKED ISSUE #775: fix(plugin-trading): make production rate limits cross-replica and restart-safe
## Problem

## Required change

## Acceptance evidence

lalalune:
Closes #775

## Summary
- require Redis-backed order and operator-transfer limiting in multi-instance production
- permit process-local fallback only outside production or with an exact single-instance acknowledgement
- bound the fallback and reclaim expired keys without resetting live budgets
- return explicit 503 before parsing/signing when durable enforcement is unavailable
- resolve posture from the immutable request environment

## Exact local evidence
- head: `cccedd08a908e34c73d10dfcb33f1f5bb0a2cae8`
- focused helper + mounted route tests: 8/8, 24 expectations on the patch-equivalent pre-rebase tree
- affected Biome and exact rebased diff-check passed
- full package typecheck was not claimed locally because concurrently rebuilt dependency outputs were inconsistent; hosted exact-head build is required

Kept draft pending fresh exact hosted checks and independent review.


lalalune:
Closed unmerged for #657 containment after branch protections disappeared again and #776 merged with zero reviews. Exact head 4382e6e17e5946436650087771c0fddb24ad0ab6 is preserved locally; no source disposition is discarded.

PR #790, opened 2026-08-21
fix(eliza-plugin): fail closed when unconfigured

### LINKED ISSUE #728: test(api): remove redundant account-alias source scan
`packages/api/src/__tests__/account-aggregation-aliases.test.ts` only greps route source for alias strings and implementation details. The real composed-app `privy-completeness.integration.test.ts` already calls `/user/me/account`, both user aggregation aliases, the agent account route, and the agent aggregation alias, then compares their reachable response shapes. Delete the redundant structural test and retain the behavioral coverage as the authority. Validation should run the owning behavioral suite and API build; do not replace it with another source scan.

### LINKED ISSUE #760: cleanup(eliza-plugin): remove inert fallbackLocal configuration
## Problem

`StewardPluginConfig.fallbackLocal` and `STEWARD_FALLBACK_LOCAL` imply a local signing fallback, but `StewardService` explicitly documents that the value only gates one informational log line. No key, signer, or alternate implementation exists. The option defaults true and produces “Falling back to local signing” even though every public operation still fails `assertConnected()`.

This is dead configuration and misleading production/operator behavior.

## Acceptance

- Remove `fallbackLocal` from public config/types/env/docs/examples and remove the misleading log, or implement a separately reviewed real custody-safe fallback (removal is preferred).
- Ensure connection failure reports the actual unavailable state and later operations fail with actionable, non-secret diagnostics.
- Update integration/config tests to prove no advertised local signing path exists.
- Run Eliza plugin tests/build, package/API compatibility checks, Biome, and diff check.

### LINKED ISSUE #763: fix(eliza-plugin): fail closed on missing config and disable implicit auto-registration
## Problem

`StewardService.resolveConfig()` is typed nullable and `initialize()` has a “No configuration found, plugin disabled” branch, but the resolver never returns null: it silently defaults `apiUrl` to `http://localhost:7860`, `agentId` to `default`, and `autoRegister` to true.

This contradicts current docs (`STEWARD_API_URL` required), makes the disabled branch unreachable, and lets a partially configured production character probe localhost and auto-create a default agent on a 404. The archived design also recorded auto-register default false.

## Acceptance

- Require an explicit API URL plus an intentional authentication/tenant/agent posture before enabling the plugin; otherwise return null/disabled with a precise diagnostic.
- Default auto-registration to false and require explicit opt-in with sufficient scoped authority.
- Remove the inert local-signing fallback tracked in #760 in the same stack or compose cleanly with it.
- Prove missing/partial config causes zero network calls and no registration; explicit localhost development config remains available; production URL/auth errors fail closed.
- Keep docs, types, env examples, and changelog current.
- Run full Eliza plugin tests/build, integration tests, Biome, and diff check.

lalalune:
## Summary

- require a normalized, nonblank Steward API URL, authentication credential, and agent identity before any network request
- default agent auto-registration off and accept only a real boolean opt-in
- require tenant-scoped API-key authority for registration and use a fresh API-key-only client so bearer preference cannot shadow it
- remove the inert fallbackLocal option, stale local-signing message, and redundant account-alias source scan
- document the implemented STEWARD_JWT contract

## Validation at exact head `3431a6485ebef484786a3b394bb72588c24bdf96`

- successor focused configuration/authority/header suite: 18/18 passed
- registration POST proves X-Steward-Key and X-Steward-Tenant are present and Authorization is absent; normal GET retains bearer preference
- package typecheck, Biome, and `git diff --check`: clean
- predecessor hosted exact-head matrix: 62 success, 1 skipped; the complete protected matrix is rerunning on this successor head

Independent audit found no blocker. This PR does not alter repository protections or bypass gates.

Closes #728
Closes #760
Closes #763


lalalune:
Closing this stale remote head unmerged under the durable-protection gate in #657. Independent exact audit found that character `autoRegister: "false"` was truthy, registration reused the bearer-preferred SDK client instead of the required tenant API key, and blank config values could trigger network I/O. The complete local repair is preserved as `3431a6485ebef484786a3b394bb72588c24bdf96` (parent is this PR head): exact-boolean opt-in, normalized nonblank config, API-key-only tenant registration client and header assertions, plus current docs; focused tests pass 18/18 and typecheck/Biome/diff are clean. No repaired bytes are being published until #657 is durably resolved. Keep #728/#760/#763 open for the protected successor and fresh exact-head hosted/review gates.

Issue #788, opened 2026-08-21
fix(staging): diagnose and bound API boot before Railway health


lalalune:
## Confirmed live failure

Exact develop `201c1869d40ffca8a207a32a24eecc7eb6f24cd6` has terminal-green CI, Docker, and CodeQL. Railway deployments `1148adcd-f137-4ebf-bfc5-51e22ad4189c` and `66606aa6-9372-47d8-b6a3-ba40129e70a7` both reported SUCCESS, but every public `/health` probe returned HTTP 000. A live probe resolves the Railway domain, completes TLS 1.3 and HTTP/2, sends the request, then receives no response headers or bytes. The merged container bind fix is therefore necessary but insufficient.

## Source boundary

`packages/api/src/index.ts` does not call `Bun.serve` until after compose, core/plugin migrations, the RLS assertion, Redis/auth-store probes, schedulers, and custody initialization. The core migrator takes a blocking advisory lock and uses PostgreSQL without bounded connection/lock/statement/overall boot deadlines. Railway service target-port, domain attachment, and dashboard start-command overrides are not verified in repository source.

## Required correction and evidence

- On post-SUCCESS health exhaustion, collect bounded Railway deployment logs with comprehensive secret redaction; use a configurable bounded readiness window rather than three opaque probes.
- Add ordered, redacted startup phase markers and explicit deadlines so a stalled migration/Redis/auth/custody phase fails rather than leaving a live-but-unlistening container indefinitely.
- If logs show `Steward API running on 0.0.0.0:<port>`, verify the Railway service instance start command, target port, and domain attachment against that exact port.
- If logs stop in migrations, prove bounded advisory-lock/connect/statement/overall behavior and capture the blocking PostgreSQL PID/lock class without leaking connection details.
- Redeploy one immutable exact SHA and record terminal CI/Docker plus an HTTP 200 public `/health` and `/ready` receipt.

A validated local-only diagnostics commit exists, but no carrier should be published while #657 protection durability remains unresolved.

lalalune:
Exact live root cause update for develop `201c1869d40ffca8a207a32a24eecc7eb6f24cd6`:

- Railway runtime logs first failed on missing `STEWARD_APP_DATABASE_ROLE`; after setting the non-secret expected role to `steward_app`, startup reached and failed the RLS deployment safety assertion.
[...]

Issue #787, opened 2026-08-21
fix(api): fail adapter daily spend caps closed without durable state


lalalune:
## Problem

The Redis spend helper returns allowed=true and spent=0 when Redis was never configured. Swap, earn, and bridge adapter routes rely on it for STEWARD_ADAPTER_DAILY_CAP_USD, so supported Redis-optional production silently disables cumulative daily enforcement.

## Required change

Fail closed in production when a configured adapter daily cap cannot use durable state, or implement a durable PostgreSQL fallback. Preserve permissive behavior only for explicit development/test posture. Resolve posture through the immutable request environment.

## Acceptance evidence

Direct production wrapper plus mounted swap/earn/bridge denial with Redis absent; configured-down denial; explicit development posture; cross-replica/restart accumulation if a PostgreSQL fallback is chosen; API/Redis builds and exact hosted checks.

lalalune:
Reopened after exact post-merge audit of PR #794 final head eebfc2229 / develop 4b2daac407b894089fffaae7c70d745685f902b0. Final PR #794 bytes still reserve caller-supplied estimatedUsd across swap, earn, bridge, and Spark money paths. Binding the estimate into a digest does not make it authoritative, so high notional with a tiny estimate bypasses caps. Keep open for a current-develop corrective with hostile production tests, independent review, and terminal hosted gates.

lalalune:
Exact merged-head audit of PR #860 at 7ae3f03e / develop merge a521e7ae: P1 BLOCK. The route values adapter-produced amountIn using caller-controlled quoteInput.fromToken.address in swap/build, and similarly uses parsed quote token identity for internal bridge valuation. The mock artifact does not bind source-token identity, so an attacker can pair a valuable amount/artifact with a cheap token address and understate per-operation/daily USD spend. Final delivery had zero reviews and queued/skipped checks. Require a protected corrective that binds token identity into a trusted artifact or server-held signed quote, with hostile tampered/cheap-token regression tests.

lalalune:
[...]

Issue #786, opened 2026-08-21
fix(api): make Bun global rate limiting production durable


lalalune:
## Problem

The Bun entry creates one process-local InMemoryRateLimiter for every non-health request. Production can run multiple replicas, so each replica grants a full 100/minute bucket and restarts reset enforcement. The shared Redis-backed global limiter is mounted only for Workers.

## Required change

Use a cross-replica/restart-safe production limiter for Bun, or fail closed unless an exact single-instance acknowledgement is configured. Memory may remain for development/test. Read posture through the immutable runtime environment and reflect any required durable dependency in readiness.

## Acceptance evidence

Two independent Bun contexts share the boundary or both fail closed; restart cannot reset production enforcement; Redis absence/outage posture; explicit dev/single-instance behavior; mounted request evidence; API build/typecheck and hosted exact checks.

lalalune:
Reopened after exact post-merge audit of PR #794 final head eebfc2229 / develop 4b2daac407b894089fffaae7c70d745685f902b0. Final PR #794 bytes still mutate isolate-global process.env per Worker request while auth and readiness paths read it; overlapping bindings can reinterpret identity and outage posture. The durable Redis key also omits the effective cap. Keep open for a current-develop corrective with hostile production tests, independent review, and terminal hosted gates.

lalalune:
PR #859 explicitly said it did not close #786 and did not repair the broader Worker hydrateProcessEnv/process-global authority boundary, but GitHub parsed the text as a closing reference and auto-closed this issue. Reopened: the exact final head changes the capability limiter only; the process-global Worker authority blocker remains.

lalalune:
PR #867 explicitly excludes the broader request-local Worker/global authority lane and merged with multiple known failing checks and no review. This issue remains open and must not be treated as covered by #867.

lalalune:
Exact PR #867 audit confirms this issue remains out of scope and open. #867 only changes API capability readiness/Redis probing; it does not close broader Bun/Worker global request-local authority. #867 itself also has an uncancelled late-EVAL readiness P1 and failed exact checks.

lalalune:
[...]

Issue #783, opened 2026-08-21
docs(deploy): use the API runtime's canonical PORT variable


lalalune:
## Problem

Three active deployment guides configure `STEWARD_PORT`, but the API runtime reads `PORT` and the shipped Compose artifacts use `PORT`. Operators following these guides cannot change the listen port with the documented variable, and may incorrectly believe the setting took effect.

Affected active guides:

- `docs/guides/self-hosting.mdx`
- `docs/guides/docker.mdx`
- `docs/guides/cloud-deployment.mdx`

## Required change

- Replace active `STEWARD_PORT` examples with the canonical `PORT`.
- Add or extend a deploy/docs contract so `STEWARD_PORT` cannot recur in active guides.
- Keep all examples consistent with the canonical default port 3200.

## Acceptance evidence

Focused deploy/docs contract tests, active-doc link/config checks, Biome/diff check, and exact-head hosted checks.

PR #791, opened 2026-08-22
fix(runtime): make production authority and limits durable

### LINKED ISSUE #764: fix(worker): make custody authority request-local across binding rotations
## Problem

Worker custody configuration is not request-local even though the Worker explicitly supports isolate reuse across binding-only deployments and concurrent requests.

- `packages/api/src/services/context.ts` captures `STEWARD_MASTER_PASSWORD` at module import and later uses it as a fallback.
- direct `SecretVault` / `KeyStore` consumers in secret, KMS, provider, tenant-config, and plugin paths retain that first authority through the cached composed app.
- the late-bound vault factory still reads mutable `process.env`, so overlapping Worker requests can select another request's hydrated password.
- the vault cache identity omits `STEWARD_KDF_SALT` and other custody parameters.
- removal of a required binding can silently continue using the old captured password instead of failing closed.

This makes sequential secret/password/KDF rotation on a reused isolate stale and makes overlapping requests capable of selecting the wrong custody root. The result can be durable ciphertext unreadability, signing/decryption failure, or continued use of retired authority after isolate recycle/rotation. This is a production integrity/availability boundary, not a client-controlled secret-injection issue.

Cloudflare documents that binding-only deploys may reuse isolates and that isolates may serve concurrent requests:

- https://developers.cloudflare.com/workers/runtime-apis/bindings/
- https://developers.cloudflare.com/workers/reference/how-workers-works/

## Required fix

- Resolve a complete immutable custody authority through the request-local runtime environment: master password, KDF salt, custody mode/provider identity/region, local-custody acknowledgement, and any relevant chain/RPC/legacy gates.
- Remove the exported import-time `MASTER_PASSWORD` fallback and fail closed when required current bindings are absent.
- Replace route/plugin `SecretVault` and `KeyStore` singletons with authority-keyed accessors, passing KDF salt explicitly.
- Key any safe cache by the canonical full authority tuple/fingerprint, not password alone.
- Preserve Bun behavior through the same runtime-environment resolver.

## Acceptance evidence

- Sequential Worker binding reuse A -> B proves every secret/KMS/provider/plugin path uses B, and removing B fails closed rather than falling back to A.
- Hostile overlap suspends request A, runs B, then resumes A and proves neither request selects the other's vault/key store.
- Same password with different KDF salts produces distinct authorities/instances and no stale-cache reuse.
- Mounted secret create/use or shared factory behavior proves ciphertext created in each scope decrypts only under that scope.
- Exact API/Worker focused tests, affected builds, lint/typecheck, and hosted Worker checks pass.

### LINKED ISSUE #767: docs(architecture): describe current RLS and encryption boundaries accurately
## Problem

`docs/concepts/architecture.mdx` describes two security boundaries that no longer match production:

- it says database row-level security is not implemented, but `develop` ships the 0111 policy inventory, activation/rollback operator scripts, runtime `assertRlsDeploymentSafety`, and Bun/Worker startup/readiness enforcement. The accurate current boundary is that PostgreSQL production requires separately provisioned roles plus explicit policy activation; PGLite/local mode relies on the service layer.
- it claims a master -> per-tenant -> secret encryption-key hierarchy. `KeyStore` actually derives a deployment/domain root with scrypt, then derives a per-record key from a random salt and binds tenant/agent/name/version context as AES-GCM AAD. There is no independently derived/stored per-tenant key layer.

These are active architecture docs, not historical notes. The first understates a shipped defense; the second overstates cryptographic tenant key separation.

## Required fix

- Rewrite the tenant-isolation section to describe service filtering/composite FKs plus the exact production RLS activation/readiness boundary and the local/PGLite distinction.
- Replace the key-hierarchy diagram/text with the implemented KDF-domain root -> random per-record derivation + tenant-context AAD model, including the legacy-root compatibility gate without implying a per-tenant key.
- Link to the current RLS rollout/operator and key-rotation documents.
- Verify all architecture links and ensure no other active docs repeat the obsolete claims.

No matching open issue or PR was found at exact `develop` `3559ac2108d118730ff99e0b8a2a9fdc55895706`.

### LINKED ISSUE #768: fix(worker): make adapter selection request-local and fail closed on binding changes
## Problem

The global adapter registry caches the first resolved adapter for the lifetime of a Worker isolate, outside the request-local runtime environment.

- `packages/adapters/src/registry.ts` exports one module-global `new AdapterRegistry()` using `process.env`.
- each category is memoized in `resolved` after its first access.
- API adapter routes import that global registry directly, while the Worker composes/caches the app once per isolate and mutates `process.env` for each request.
- binding-only deploys may reuse the isolate, so a changed `STEWARD_<CATEGORY>_ADAPTER`, `NODE_ENV`, or `STEWARD_ALLOW_MOCK_ADAPTERS` does not invalidate the cached category.

The most dangerous transition is removal of an accidental/staging mock opt-in: once a category has resolved to its mock, a later binding set that disables mocks can continue returning the old enabled mock until isolate recycle. Provider selection/disable changes are likewise stale. Overlapping requests also read mutable global configuration rather than one immutable request snapshot.

## Required fix

- Resolve adapter policy from the immutable request-local runtime environment, not global mutable `process.env`.
- Separate the durable registry of explicitly registered provider implementations from request-scoped selection/policy.
- Key any safe resolved cache by a canonical authority tuple containing runtime mode, mock acknowledgement, and every category selection, or resolve cheaply per request.
- A missing/removed production binding must fail closed immediately; it must never reuse an adapter selected under an older binding set.
- Preserve explicit plugin registration collision rules and Bun/non-Worker configuration.

## Acceptance evidence

- Sequential Worker binding reuse proves `mock allowed -> mock removed` becomes disabled on the next request without isolate recycle.
- Sequential provider A -> B and configured -> missing transitions select B/fail closed immediately.
- Hostile overlap suspends A, runs B, resumes A, and proves each request retains its own adapter policy.
- All adapter categories and plugin-registered providers are covered; production cannot use mocks without the current request's exact acknowledgement.
- Mounted adapter routes plus Worker runtime tests and affected builds/lint pass.

### LINKED ISSUE #769: fix(worker): bind webhook transport and SSRF policy to each request
## Problem

Webhook transport and network-destination policy is frozen at module import instead of being resolved from the current Worker request environment.

At exact `develop` `3559ac2108d118730ff99e0b8a2a9fdc55895706`, `packages/api/src/services/webhook-url.ts` captures:

```ts
const ALLOW_INSECURE_WEBHOOK_URLS = process.env.STEWARD_ALLOW_INSECURE_WEBHOOK_URLS === "true";
```

`validateWebhookUrl()` then uses that frozen value for every registration/update/delivery validation. Worker binding-only deploys may reuse an isolate. If a staging/break-glass deployment first imports the module with insecure public HTTP enabled, removing the binding does not restore HTTPS-only policy until isolate recycle. Conversely, enabling it later is not observed. Overlapping requests also do not retain independent immutable policy snapshots.

The delivery package separately captures `STEWARD_ALLOW_PRIVATE_WEBHOOK_NETWORKS` at module import and copies it into the module-global `WebhookDispatcher` constructed in API context. If an isolate first starts with that break-glass acknowledgement, removing it does not restore the DNS/IP SSRF guard: the cached dispatcher can continue fetching loopback, link-local, metadata, or other private destinations. This is a direct stale-authority SSRF boundary as well as a transport-confidentiality/integrity regression for cleartext HTTP.

## Required fix

- Resolve both insecure-HTTP and private-network acknowledgements from `runtimeEnvironmentValue` at each request-scoped validation/delivery, with a Bun fallback through the same resolver.
- Default/fail closed when the current binding is absent; never fall back to an import-time or another request's value.
- Keep string and DNS-resolved registration/delivery validation on the same current policy.
- Replace the module-global policy-bearing dispatcher or key it by the immutable request authority; a dispatcher created under an older acknowledgement must never serve a later request.

## Acceptance evidence

- Sequential Worker reuse `enabled -> binding removed` accepts public HTTP only in the first request and rejects it in the second without isolate recycle.
- The same transition allows then rejects a private/loopback destination at delivery time, with zero network request after removal.
- Hostile overlap suspends enabled request A, runs disabled B, resumes A, and proves each validation keeps its own policy.
- Missing binding, malformed/case-varied values, HTTPS, private/local hosts, DNS rebinding, and delivery-time revalidation remain covered.
- Worker/API focused suites, affected build, lint/typecheck pass.

### LINKED ISSUE #771: refactor(api): remove the unused process-local SIWE nonce timer
## Problem

`packages/api/src/services/context.ts` still exports a process-local "SIWE nonce store" and a five-minute cleanup interval, but no production or test code reads or writes the map.

The only external reference is `packages/api/src/index.ts`, which imports the timer solely to clear it during shutdown. Current SIWE/auth nonce flows use the auth store backends instead. The dead map therefore provides no protection, creates an unnecessary long-lived timer in Bun, and misleadingly suggests SIWE has a process-local authority path.

## Required fix

- Remove `nonceStore`, `nonceCleanupTimer`, the shutdown import/clear, and the stale section comment.
- Verify the whole-tree reference inventory is empty.
- Run API typecheck/build plus focused auth/SIWE tests and lint/diff checks.
- Do not replace it with another fallback; durable auth stores remain authoritative.

No matching issue or PR was found at exact `develop` `3559ac2108d118730ff99e0b8a2a9fdc55895706`.

### LINKED ISSUE #774: fix(agent-trader): fail fast on missing credentials and use canonical API port
## Problem

`packages/agent-trader/src/config.ts` defaults the API URL to `http://localhost:3000` even though Steward's canonical API port and the checked-in example use 3200. It also defaults `apiKey` to an empty string and validation does not reject that value.

An enabled autonomous agent can therefore start (including with `dryRun: false`) against the wrong endpoint with no usable credential; the initial connectivity failure is non-fatal.

## Required change

- Use the canonical localhost API port 3200.
- Reject a missing/blank API key whenever any configured agent is enabled.
- Preserve credentialless configuration only when no API-driven agent loop can start.
- Add behavioral fail-fast configuration/startup tests.

## Acceptance evidence

Focused package tests, package typecheck/build, Biome/diff check, and exact-head hosted checks before merge.

### LINKED ISSUE #775: fix(plugin-trading): make production rate limits cross-replica and restart-safe

### LINKED ISSUE #777: docs(solana-signer): use the canonical Steward API port
## Problem

The primary @stwd/solana-signer README and runtime guidance advertise http://127.0.0.1:3000. Steward's canonical API port is 3200. Copying the package instructions can target an unrelated listener or fail to connect.

## Required change

- Update active signer instructions and runtime guidance to port 3200.
- Add a lightweight contract test so package examples cannot drift from the canonical API default again.
- Verify all active signer examples remain consistent.

## Acceptance evidence

Focused package tests/typecheck, documentation/source contract, Biome/diff check, exact-head hosted checks.

### LINKED ISSUE #778: docs(self-hosting): state the production JWT secret boundary accurately
## Problem

docs/guides/self-hosting.mdx says STEWARD_JWT_SECRET defaults to the master password while the same example sets STEWARD_EMBEDDED=false. Runtime accepts master-password derivation only in embedded/local mode; non-embedded production requires a JWT/session secret. The example literal is also shorter than the production 32-character minimum.

Operators following the active guide can boot-fail and are given a false key-separation contract.

## Required change

- State that STEWARD_JWT_SECRET is required for non-embedded production.
- Require at least 32 high-entropy characters.
- Describe master-password derivation as embedded/local-only.
- Use a generation command/placeholder rather than a short literal secret.
- Add or extend the docs/deployment contract test.

## Acceptance evidence

Active-doc link/config checks, relevant auth/deploy contract tests, Biome/diff check, exact-head hosted checks.

### LINKED ISSUE #779: fix(auth): accept allowlisted IPv6 loopback OAuth redirects
## Problem

`parseOAuthRedirectUri` intends to allow plain HTTP only for loopback development origins, including IPv6 `::1`. Bun/WHATWG `URL.hostname` serializes an IPv6 literal with brackets (`[::1]`), while the parser compares it to unbracketed `::1`. As a result, an exact tenant allowlist entry such as `http://[::1]:4173/callback` is rejected before allowlist matching.

## Acceptance

- Recognize the WHATWG/Bun canonical IPv6 loopback hostname consistently.
- Keep non-loopback HTTP redirect URIs rejected.
- Add mounted `/oauth/google/authorize` behavior proving an exact tenant allowlist permits `http://[::1]:PORT/callback` and rejects an allowlisted non-loopback HTTP origin.
- Run focused API tests, API typecheck/build where feasible, Biome, and diff check on the exact commit.

lalalune:
## Summary

- bind custody, adapter selection, and webhook transport policy to each request instead of process-cached Worker state
- fail production trading, adapter daily caps, and global API limits closed unless their state is durable; make reservations atomic and retain holds through builders
- fail agent-trader startup closed on missing credentials and remove the unused SIWE nonce timer
- correct OAuth IPv6-loopback handling and deployment/runtime documentation
- bound and redact Railway diagnostics, database migrations, startup phases, and overall API boot

## Exact head

380a0b686 on base 201c1869d40ffca8a207a32a24eecc7eb6f24cd6.

## Local validation

- plugin-trading full isolated package suite: 178 passed, 717 assertions
- plugin-trading production/cache regression files: 37 passed, 193 assertions
- plugin-trading typecheck: passed
- startup phase and migration deadline suites: 10 passed
- webhook SSRF/request-overlap suite in isolation: 10 passed
- database package build/typecheck: passed
- API package build/typecheck: passed using the existing dependency cache
- Biome on the final changed test files: passed
- git diff check: passed

The repository hosted matrix remains the authoritative validation for the full 93-file carrier. A single-process cross-package local aggregation was intentionally not claimed because API test module mocks contaminate webhooks tests when packages are co-scheduled; the protected workflow runs package jobs in their intended isolation. This PR does not alter protections and must not bypass independent review or exact-head checks.

Explicitly not closed here: #765, #766, and #770 did not have an accepted implementation in this carrier.

Closes #764
Closes #767
Closes #768
Closes #769
Closes #771
Closes #774
Closes #775
Closes #777
Closes #778
Closes #779
Closes #783
Closes #786
Closes #787
Closes #788


lalalune:
Closing unmerged after exact-head adversarial audit of `cf25b5b3b1e614e3a73bbcbac426b00f48cdeceb` (the only delta from the prior audited head is formatting).

This 93-file carrier is not safe to accept as a unit:

- `globalRateLimitPosture()` still selects process memory for every non-`production` `NODE_ENV`, including unset/staging/typo postures, so an unknown production runtime can fail open.
- Redis availability/client authority remains module-global and `initRedis(env)` mutates `process.env`; overlapping Worker binding generations can inherit stale client/config authority.
- adapter issuance reservations are not idempotent, so a retry can double-reserve, and they reuse the settled-spend bucket in a way that can double-count later vault accounting.
- the #788 migration path creates `drizzle` schema/table state before validating that the database is a legitimate Steward schema, so a wrong/shared database is mutated before rejection.
- the carrier mixes rate limiting, custody, webhook, deployment, migration, docs, and unrelated runtime changes; the valid pieces cannot make those blockers safe.

The exact commit remains preserved on its branch for comparison. #775/#786/#787 are being repaired in a narrower local-only current-base lane; #788 requires its own topology-safe successor. Do not close the linked issues from this PR.

Issue #761, opened 2026-08-20
fix(capabilities): replace process-local production rate-limit fallback


lalalune:
## Problem

`packages/plugin-capabilities/src/rate-limit.ts` falls back to a process-local `Map` whenever Redis is not configured. Bun explicitly supports production “Redis optional mode” with PostgreSQL auth storage, so this path is reachable in multi-process production: every replica gets an independent 60/minute or 30/minute bucket and restarts reset it.

The comment calls this a local-dev posture, but no development/test gate enforces that assumption. Agent capability invocation and manifest issuance can therefore exceed their documented global limits.

## Acceptance

- In production, use a durable cross-replica reservation/count boundary (Postgres advisory-locked rows or another durable backend) when Redis is absent; configured-but-failing Redis remains fail closed.
- Permit the memory bucket only behind explicit development/test posture.
- Prove two independent service contexts/processes cannot jointly exceed the limit, including exact boundary, restart, and concurrent requests.
- Cover Redis outage, absent Redis in production, and local-dev memory behavior without writing invocation rows for rejected attempts.
- Document the backend/readiness contract and run plugin/API builds, real-Postgres/Redis tests, Biome, and diff check.

lalalune:
Exact merged-head audit of PR #859 at b08c43dc / develop merge 087523c1: source is materially strong, but acceptance remains HOLD. The final head had zero reviews, 61 cancelled checks, one skipped check, and the PR body cited stale 23fb40b4 rather than the merged head. #761 explicitly requires exact hosted real-PostgreSQL/Redis/build evidence. Keep open until a protected exact-head successor or equivalent same-byte hosted acceptance exists.

lalalune:
PR #867 merged unreviewed while its exact matrix had known failures (including metadata, plugin-capabilities, plugin-trading, web, integration PG, and Trust UX) and other jobs nonterminal. Its Redis readiness source is being audited against the full durable capability contract; this issue remains open.

lalalune:
[...]

PR #794, opened 2026-08-22
fix(rate-limits): bind production enforcement to durable authority

### LINKED ISSUE #775: fix(plugin-trading): make production rate limits cross-replica and restart-safe

### LINKED ISSUE #786: fix(api): make Bun global rate limiting production durable

### LINKED ISSUE #787: fix(api): fail adapter daily spend caps closed without durable state

lalalune:
## Summary

- fail unknown, staging, and Worker global/money-path rate-limit postures closed unless Redis is durably available
- bind Redis clients, configuration, and package helper resolution to immutable request authority without mutating process environment
- make adapter issuance daily reservations atomic, idempotent, request-plus-artifact-bound, bounded, and separate from settled execution spend
- require durable Redis before serving Worker traffic, including health probes, and recover safely after a failed first-generation initialization
- require encrypted Redis transport outside explicit local postures
- align isolated hosted fixtures with the durable Redis constructor and production-only memory acknowledgements without weakening runtime defaults

## Exact head

bcd42983cee4d095da9429861d642ed7d6f207a4 on develop base 201c1869d40ffca8a207a32a24eecc7eb6f24cd6.

## Validation

- adapter PGLite: 20/20
- Redis TLS/resolver and wrapper suites: 53/53
- Worker boot/recovery and scheduled-runtime suites: 28/28
- auth session + exact Redis constructor fixture: 35/35
- plugin-trading production-negative and order suites: 35/35
- combined focused rerun: 100/100
- API dependency graph build: 24/24
- actionlint, targeted Biome, and diff-check: clean
- exact hosted package, PostgreSQL, Compose, Docker, and CodeQL jobs plus independent review remain required

This is the narrow replacement for the rate-limit portions of closed unsafe #791; it does not carry unrelated custody, webhook, capability-plugin, migration, or deployment changes. #761 remains open for the plugin-capabilities Postgres fallback.

Closes #775
Closes #786
Closes #787


lalalune:
Independent exact-head audit of bcd42983cee4d095da9429861d642ed7d6f207a4 (tree b539faa3, base 201c1869) is BLOCK (P1). Adapter caps reserve caller-supplied estimatedUsd rather than authoritative value: swap, earn deposit/withdraw, non-external bridge, and Spark fund movers accept arbitrary positive estimates. A high-notional request can declare 0.0001 USD, receive the unsigned artifact, and consume negligible Redis budget; tests only reject caller-declared high estimates and never bind amount/quote/artifact to valuation. Worker isolation is also incomplete: Redis authority is ALS-bound, but auth client identity, proxy hops, Cloudflare trust, and outage soft-fail posture still read process.env while overlapping Worker requests can change the global mirror, so forwarded headers and fail posture can be interpreted under another request authority. The durable Redis key also omits maxRequests, allowing overlapping generations with different limits to reinterpret one counter. Add authoritative valuation/understatement tests, request-local trust authority with hostile overlaps, and generation-safe limit keys. I restored draft; hosted/review gates remain incomplete. Do not merge.

lalalune:
Closed as a rejected stale carrier. Rate-limit valuation and enforcement still depend on raw process-global authority rather than the required request-local durable boundary.

PR #795, opened 2026-08-22
fix(custody): bind Worker authority to each request

### LINKED ISSUE #764: fix(worker): make custody authority request-local across binding rotations

lalalune:
## Summary

- remove secret-derived process-global cache identities across custody, webhooks, Polymarket, and EmailAuth
- bind memoization to opaque request identity and WeakMaps, zero derived buffers, and retire rotated EmailAuth providers
- preserve immutable Worker request-overlap behavior without retaining prior secret generations

## Exact head

e68d3fe6a4f7c4aeff02de78b86122d8578da629 on develop base 201c1869d40ffca8a207a32a24eecc7eb6f24cd6.

## Validation

- 69 focused tests / 494 assertions across runtime, webhooks, trading, vault KMS, custody, EmailAuth, and Worker overlap
- API/Auth/Vault/Webhooks/plugin-trading builds and TypeScript checks pass
- changed-file Biome, diff-check, and secret-derived-identity scan pass
- protected hosted checks and independent approval remain required

Closes #764


lalalune:
Adversarial exact-head audit found a merge blocker: this head hashes secret-bearing custody tuples to key process-global caches and can retain old EmailAuth/provider generations across rotations. Keep draft; do not ready or merge. Repair must use opaque/request-local authority identity (not secret-derived digests), bound and retire unavoidable cross-request caches, zero derived buffers on eviction, and prove old generations are unreachable after rotation. Hosted green status on this head will not be treated as acceptance.

lalalune:
Independent exact audit at 5e02ebca30efd03ac33a7ce41852f70951a79e21: BLOCK; keep draft. The request-local custody factory work is materially sound, but two authority gaps remain. (1) packages/vault/src/secret-route-validator.ts still reads STEWARD_ALLOW_BROAD_SECRET_ROUTES, STEWARD_ALLOW_COOKIE_INJECTION, and STEWARD_SECRET_ROUTE_ALLOWED_HOSTS from mutable process.env, so overlapping Worker request B can change request A's durable route-validation authority. Bind/inject immutable request authority and prove hostile suspended overlap at mounted route creation. (2) packages/webhooks/src/secret-codec.ts accepts a truthy whitespace-only encryption key and permissively decodes hex salts with Buffer.from, allowing valid-prefix/invalid-suffix truncation. Trim/fail blank keys and require strict full-string salt decoding with regressions. Also bound or eliminate the unbounded per-authority email-auth promise/plaintext-key cache if confirmed. No merge at this head; exact hosted checks/review cannot supersede these semantic blockers.

lalalune:
Exact-head blocker repair pushed at `7df0051e94642b31645387c6877f7bc177ef175b`.

This removes the rejected secret-derived cache identity class from custody, webhook roots, Polymarket credential derivation, and EmailAuth. Runtime memoization is now opaque request-identity/WeakMap scoped; webhook and venue-derived buffers are zeroed; EmailAuth invalidation retires the credential-bearing provider; the obsolete global plaintext-provider cache contract was removed. The newer mounted request-authority overlap fixes from the prior head are preserved.

Local exact-byte evidence: 69 focused tests / 494 assertions green across custody, mounted Worker overlap, EmailAuth, webhook rotation, Polymarket A-B-A, KMS, and runtime isolation; Shared/Auth/DB/Vault/Webhooks/Policy Engine/venues/plugin-trading/adapters/provider packages/plugin-capabilities/API builds green; 13-file Biome and diff-check clean; secret-derived-identity source scan empty. Temporary dependency harness removed before push. Hosted checks and independent review remain required before readiness/merge.

lalalune:
Exact head is now `e68d3fe6a4f7c4aeff02de78b86122d8578da629`; the sole delta from the validated repair is a test-title correction from the removed global-cache terminology to the actual request-local reachability contract. Biome and diff-check remain clean; hosted checks restarted on this exact head.

lalalune:
Exact-head audit BLOCK at `e68d3fe6a4f7c4aeff02de78b86122d8578da629` (tree `17e6fa38aef7c5d2a725dd980d19a735d3b95172`). The cached plugin context constructs the EVM simulator once, while `createEnvEvmSimulator` decides availability from global `process.env` and the simulator retains that mutable object by reference. Worker requests then overwrite the same global while reusing the cached app. First-request missing RPC can freeze simulation off for later requests; configured A overlapping B can resume against B’s RPC. This reaches governed swap pre-sign simulation and violates #764 request-local chain/RPC authority. New overlap tests manually mount selected routes under runtime snapshots, not production `worker.fetch` plus cached composition. Keep draft; 60 checks queued and no human review.

PR #796, opened 2026-08-22
fix(adapters): bind selection to request environment

### LINKED ISSUE #768: fix(worker): make adapter selection request-local and fail closed on binding changes

lalalune:
Closes #768.

## Summary

- separate durable provider inventory from request-time adapter policy
- resolve every adapter selection from the immutable request runtime snapshot
- cache only authority-free mock/disabled instances and require explicit production selection per category
- retain provider collision guards while using request-owned factories for binding-dependent providers
- resolve WXMR RPC authority inside each selection and fail closed on missing bindings
- prove A-to-B-to-missing, mock removal, hostile overlap isolation, and Bun fixed-environment behavior across all adapter categories

## Exact source audit

Independent semantic ACCEPT: no cached or module-captured production adapter authority was found. All production adapter routes resolve instances inside request handlers. Hosted checks and review remain required.

## Exact lineage

Head: debe9f12db83d6fa711c2d67a1f852fae396429e
Tree: e0162b6c8e7b71a23fe7f2fa6095e0c3596ae524
Scope: 12 changed files, including provider-factory authority

This PR stays draft until the exact hosted matrix, Docker check, and required review are terminal.


lalalune:
Independent exact source audit at de22c0023edb3db046606b89da76360cbf632cc7: semantic PASS. Selection policy is resolved on every access from immutable runtime authority; registered implementations remain durable; all categories cover mock removal, provider A-to-B, missing/unknown selection, current-request acknowledgement, and hostile overlap. Worker without NODE_ENV fails closed while explicit Bun dev/test behavior is preserved. Diff check is clean. Local re-execution in the preserved worktree was dependency-blocked (workspace packages absent), so exact hosted package/build checks and independent human review remain required; keep draft.

lalalune:
Exact-head follow-up found a merge blocker: request-local provider selection still resolves to the module-cached WXMR adapter created from the first Worker binding generation. A later or overlapping generation can select wxmr but execute through the prior RPC endpoint/client. Keep draft; do not ready or merge until binding-dependent adapters are request-authority factories (or resolve runtime bindings per operation) and a mounted A -> B -> A/remove regression proves endpoint authority and fail-closed removal.

lalalune:
Exact-head Worker authority repair pushed at `debe9f12db83d6fa711c2d67a1f852fae396429e`.

AdapterRegistry now supports typed request factories; the WXMR contribution resolves its RPC endpoint from the immutable current runtime snapshot for each selection, while only authority-free bounded session state is shared. Mounted real-adapter coverage proves suspended A -> overlapping B -> resumed A, a later A request, and endpoint removal failing closed with zero network calls. Audit found WXMR was the only binding-dependent production adapter contribution.

Local exact-byte evidence: registry/WXMR/plugin-host/mounted Worker suites 51/51 (272 assertions), adapter policy gates 18/18 (112 assertions), dependency-ordered API graph builds/typechecks 24/24, 9-file Biome and diff-check clean. Temporary harness removed before push. Hosted exact-head checks and independent review are still required before merge.

lalalune:
Independent exact-current audit: semantic source ACCEPT / delivery HOLD at debe9f12db83d6fa711c2d67a1f852fae396429e (tree e0162b6c8e7b71a23fe7f2fa6095e0c3596ae524). Registry policy is request-local through immutable runtime snapshots; caches retain only authority-free instances; production requires explicit per-category selection; all categories cover A-to-B-to-missing, mock removal, and hostile overlap. Provider collision guards remain, binding-dependent providers use request-owned factories, and reachable WXMR resolves RPC per selection with no fallback. Bun fixed-environment behavior is preserved. The stale body has been corrected to this exact head/tree and 12-file provider-factory scope. Keep draft until exact hosted checks, Docker, and required review are terminal.

PR #792, opened 2026-08-22
test(api): execute remaining control-plane boundaries

lalalune:
## Summary\n\n- replace user-wallet, secret, readiness, sponsorship, approval, app-client, and external-JWT source claims with mounted behavior\n- make secret create/rotate/delete and auto-approval mutations audit-atomic\n- exercise real PostgreSQL lock/race behavior for gas caps, secrets, approvals, rules, and app-client rotation\n- reuse the pre-staged transaction anchor for sponsored SPL execution instead of duplicate-inserting its action id\n- preserve secret lineage serialization while allowing explicit PGLite-mode execution\n\n## Exact head\n\nad890288a260ffe1244780b0e38545605593fd94 on develop base 201c1869d40ffca8a207a32a24eecc7eb6f24cd6.\n\n## Validation\n\n- prior carrier parent: 116 focused mounted/PGLite/unit passes\n- prior carrier parent: 11/11 real-PostgreSQL concurrency passes after all 94 migrations\n- API and Vault TypeScript builds passed on the prior carrier parent\n- sponsored SPL anchor delta: changed-file Biome and diff-check clean\n- explicit PGLite lineage delta: secret audit atomicity 3/3 at exact head\n- exact hosted matrix and independent approval remain required at ad890288\n\nThis carrier contains the exact independently accepted #734 readiness patch (stable patch-id 22edef98b41aba87b47b8fe4498797b0320360d8); duplicate #798 was closed unmerged.\n\nCloses #720\nCloses #729\nCloses #732\nCloses #733\nCloses #734\nCloses #735\nCloses #736\nCloses #737\nCloses #738\nCloses #741

PR #798, opened 2026-08-22
test(api): mount Bun readiness behavior

### LINKED ISSUE #734: test(api): make Bun readiness auth-store checks behaviorally testable
## Gap

On exact develop `3559ac210`, `auth-import-session-store-sources.test.ts` explicitly source-scans `index.ts` because importing the Bun entry immediately starts timers, migrations, schedulers, custody initialization, and `Bun.serve`. The test therefore cannot call `/ready`; it only checks that strings such as `getAuthStoreSources` and `source === "memory"` appear nearby.

This is both weak evidence and an entrypoint testability seam: the readiness handler cannot be mounted independently from process startup.

## Acceptance

- Extract the Bun readiness handler/factory from process startup without weakening the production startup order.
- Mount it in tests with injected DB/Redis/proxy/auth-store health dependencies.
- Prove production memory-backed import-session (and every other auth store) returns 503 unless the explicit single-instance acknowledgement is present; durable stores pass.
- Assert unauthorized probes expose only redacted per-check status, authorized probes expose bounded diagnostics, and no token/backend identity leaks.
- Delete the self-referential source-only test. Keep one narrow entrypoint composition assertion only if unavoidable.

lalalune:
## Summary

- extract the Bun /ready policy into a side-effect-free handler with injected production dependencies
- preserve blocking migration, RLS, Redis/auth-store, scheduler, custody, then Bun.serve startup order
- replace source scans with mounted behavior for every auth-store source, optional dependencies, failure redaction, and authorized diagnostics

## Exact lineage

- base: 201c1869d40ffca8a207a32a24eecc7eb6f24cd6
- head: e923a4091dc156df22b8cde5ee0b18b4a1f0d0ce
- tree: 6bf07241d628aeecfdee2e5380ab8784b0be0b10

## Validation

- focused readiness/security tests: 22 pass, 0 fail, 120 assertions
- @stwd/api TypeScript build: pass
- changed-file Biome: pass
- git diff --check: pass
- direct readiness-module import exits without listener/timer side effects

Closes #734.

Held draft until exact hosted matrix and independent review complete.

### DIFF
--- a/packages/api/src/__tests__/auth-import-session-store-sources.test.ts
+++ b/packages/api/src/__tests__/auth-import-session-store-sources.test.ts
@@ -1,6 +1,12 @@
 import { afterEach, describe, expect, it, mock } from "bun:test";
-import { readFileSync } from "node:fs";
-import { join } from "node:path";
+import { Hono } from "hono";
+import {
+  createReadinessHandler,
+  type ReadinessDependencies,
+  type ReadinessEnvironment,
+} from "../readiness";
+import type { AuthStoreSources } from "../routes/auth";
+import type { AppVariables } from "../services/context";
 
 const redisSetMock = mock(async () => "OK");
 const redisGetMock = mock(async () => null as string | null);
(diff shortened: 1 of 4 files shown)


lalalune:
Closing unmerged as an exact duplicate of the #734 patch already carried in draft #792. Commit e923a409 and #792 commit 9769765b have identical stable patch-id 22edef98b41aba87b47b8fe4498797b0320360d8; independent audit evidence was 22/22 focused tests, API typecheck, Biome, diff, and side-effect-free import. #792 remains draft/review-gated and is not accepted for its other changes yet. Local candidate is preserved.

PR #799, opened 2026-08-22
fix(vault): make Solana recovery evidence and effects durable

### LINKED ISSUE #651: fix(api): anchor every Solana broadcast before external I/O

### LINKED ISSUE #673: security: make Solana recovery ownership and accounting atomic

lalalune:
## Summary

- bind recovery to normalized Solana message intent, excluding mutable signatures/recent blockhash while retaining every other message byte
- route primary and approval signing through governed digest/token assertions and exact signer inventory
- allow expired-lease takeover by a different authorized admin with atomic transaction/queue ownership CAS
- checkpoint broadcast evidence and serialize terminal state/effects without post-sign relabel or duplicate execution
- price native SOL and canonical wSOL at 9 decimals on both chain 101 and 102
- keep SPL valuation mint-aware/fail-closed and exclude token base units from native spend aggregates
- retry pending durable effects for already-terminal matching reconciliation without a second Solana RPC call

## Exact lineage

- base: 201c1869d40ffca8a207a32a24eecc7eb6f24cd6
- head: 65d2cf599a88a503c8e02b07c3efb066aa8fbdd5
- tree: 6e358322c245ec9ddf8a43656cd25a96d485931a

## Validation

- token/oracle including 101/102 wSOL valuation: 16/16
- mounted transaction stats/native isolation: 9/9
- Solana recovery/effects replay: 10/10
- approval/takeover gates: 13/13
- wallet actions: 11/11
- governed digest and signer inventories: 11/11
- targeted Biome and diff-check: clean
- API typecheck reached only an absent optional local Polymarket SDK dependency; exact hosted build is required

Real PostgreSQL takeover races, real Redis accounting, the complete hosted matrix, and independent review are required before ready/merge.

Closes #651
Closes #673


lalalune:
Exact-head audit found a merge blocker: parsed/SPL Solana paths replace the raw-vault envelope with an unbound allowParsedSign boolean. That flag carries no execution token, claim/message digest, parsed effects, policy revision, or custody-bound assertion, and production callers toggle it directly. Keep draft; do not ready or merge until this is a typed governed parsed-Solana authorization that binds the normalized message digest + parsed effects/policy revision + txId/execution token and consumes/asserts the durable claim immediately before raw key use, with the complete raw Solana call-site inventory pinned. The eventual successor also needs restack/composition testing with #792.

lalalune:
Exact-head correction published at `3975edf3aca92bf2cd497b0edbca3e4598f1b67b`.

This removes the raw `allowParsedSign` exemption. Parsed/SPL signing now requires a typed durable claim bound to normalized message digest, parsed effects, policy revision, txId, execution token, payload digest, and claim digest; the claim is atomically consumed immediately before raw custody. All parsed production paths use the governed gateway, and remaining blind/native raw callers are inventory-pinned. The head also composes accepted #792 (`ad890288a`) and preserves sponsored-SPL anchor reuse.

Local exact-head evidence: 111 focused tests / 754 assertions; API and Vault typechecks; 60 changed files Biome-clean; diff-check clean. Hosted checks and independent approval remain required.

lalalune:
LIVE EXACT HEAD 3975edf3aca92bf2cd497b0edbca3e4598f1b67b is BLOCK and supersedes all 65d2cf59 acceptance evidence. It is a merge commit with parent ad890288 from the broad PR792 behavior branch; relative to the body-listed 65d head it adds 45 files, +6105/-1317, including unrelated readiness, agent-JWT, approval, gas-sponsorship, secret CRUD, tenant app-client, user-wallet, and global-wallet changes. Several imported lines are independently blocked in their own PRs: gas sponsorship lacks the required atomic/mounted race proof (#818), secret CRUD canonical-writer locking is incomplete (#820), and tenant app-client tests bypass restricted production auth/RLS and omit real generated-secret authentication (#822). The PR body still claims exact head 65d2cf59/tree 6e358322 and only #651/#673 scope, while the new head has no attached exact checks. Do not treat this merge as a Solana correction or inherit old test results. Split/restack only the independently accepted Solana delta onto current develop, exclude blocked/unrelated parents, refresh exact lineage/body, and rerun the full hosted/PG/Redis/review gates. Keep draft; do not merge.

lalalune:
Additional independent P1 on exact 3975edf3: parsed Solana execution claims are marked consumed durably before the raw signer runs. A crash after parsedClaimConsumedAt is written but before signing/checkpoint permanently strands the approved transaction, because both direct replay and expired-lease takeover reject any row with parsedClaimConsumedAt. Takeover can also fail before consumption: the first claim computes executionDigest before adding message/policy/parsedExecution fields, while takeover recomputes from the augmented payload, so the stable-digest comparison can reject its own prior lease. The real-PG takeover test uses a synthetic unparsed row and misses both crash paths. Consumption must be a recoverable lease/CAS tied to durable checkpointed signing evidence, with expired takeover and two-connection crash-window proof. This blocker remains even after the unrelated merge-parent contamination is removed.

lalalune:
Closed as a rejected stale carrier. Native Solana recovery evidence, durable nonce and blockhash authority, strict RPC validation, evidence-locked CAS, and post-sign durability remain incomplete.

PR #800, opened 2026-08-22
fix(api): make intent finalization durable and recoverable

### LINKED ISSUE #710: test(api): prove intent finalization audit and webhook ordering behaviorally
`intent-finalization-webhooks.test.ts` on exact develop only slices `routes/intents.ts` and compares string positions. It does not prove that transfer/send-calls success webhooks are absent when required `intent.executed` audit fails, that each webhook is emitted once after commit, or that retries cannot duplicate/misorder effects. Replace it with mounted PGLite/real-PostgreSQL fault tests using the actual intent routes and webhook dispatcher: inject completion-audit failure after execution result staging and assert rollback/no success webhook; successful finalize asserts ordered durable audit then exactly one wallet-action and intent webhook; lost-ACK/replay asserts no duplicates; concurrent finalizers produce one winner. Remove source-order claims. Local-only until #657 protection is durable.

lalalune:
## Summary

- atomically finalize successful intents with required audit evidence before terminal status
- reserve autonomous finalization effects durably before external send and recover interrupted execution without rebroadcast
- enforce a database-level generic-intent agent-deletion fence
- serialize persistent webhook claims and preserve cross-event ordering and lost-ACK retry behavior
- require request-owned, closeable database authority for autonomous recovery

## Exact head

6fc45cd93767bd5350cef7147adc84b1fc18860a on develop base 201c1869d40ffca8a207a32a24eecc7eb6f24cd6.

## Validation

- focused API and real-Postgres behavior is included for deletion races, persistent ordering, worker claims, and finalization recovery
- protected hosted real-Postgres/full matrix and independent approval remain required on this exact head

Closes #710
## Blocking composition gate

- This head is not mergeable standalone: its journal jumps from 0113 to 0118 and the migration depends on absent 0114–0117.
- Compose it after the accepted migration chain or renumber it and prove a contiguous upgrade before readiness.


lalalune:
Exact-head BLOCK at efd673fbd33d3afcd0f8fddb69809fa888371f98.\n\nP1: the production webhook dispatcher is at-least-once and does not enforce cross-row wallet_action-before-intent ordering. Both rows receive independent due timestamps; concurrent SKIP LOCKED workers can split them, and a crash after remote 2xx but before the delivered-state update retries the external effect. A stable delivery ID supports receiver dedupe but is not exact-once delivery.\n\nP1: the routed service guard blocks generic executing intents, but the database deletion trigger from migration 0110 still covers only provider-action intents while intents.agent_id cascades on delete. A direct or concurrent agent deletion can erase a generic reserved/completed/outcome_unknown fence. Add a new migration-level invariant and a real two-connection race.\n\nAcceptance gap: the focused webhook suite mocks enqueue as an in-memory push. The mounted Postgres test proves reservation/outbox rows and no rebroadcast, but does not execute PersistentWebhookQueue, concurrent claims, lost ACK, actual remote effects, or Worker request-owned outer plus autonomous connection lifecycle.\n\nThe audit/status/outbox reservation transactions otherwise appear materially sound. Keep draft; do not merge this exact head.

lalalune:
Corrective successor published at `6fc45cd93767bd5350cef7147adc84b1fc18860a`. It closes both prior hosted failures with durable finalization recovery, a generic intent delete fence, dependency-ordered persistent delivery, and worker-claim fencing. Exact local rerun: 33 passed, 10 expected real-Postgres skips, 0 failed; `git diff --check` is clean. Fresh hosted exact-head PostgreSQL/full-matrix checks and independent review remain required.

Issue #696, opened 2026-08-20
blocker(security): repair merged personal-tenant account lifecycle invariants


lalalune:
Merged PR #692 (`05944ddaa50418226e929a43573dc3648ae58325`) shipped an incomplete personal-account lifecycle. This issue remains open until one current-`develop` corrective closes all four gates together.

## Required corrections

1. **Canonical personal authority at every write boundary**
   - A `personal-*` tenant must have exactly one canonical owner membership and no extra memberships or invitations.
   - Reject personal/reserved tenant membership and invitation creation at both reachable route boundaries (including `POST /me/tenants/:tenantId/invitations`) and a database invariant so future writers cannot bypass it.

2. **Production-safe privileged function upgrade**
   - Do not run `CREATE OR REPLACE steward_bootstrap.*` as the ordinary restricted migrator after bootstrap transferred ownership to the NOLOGIN bootstrap owner.
   - Put privileged replacement in the explicit admin/bootstrap upgrade lane.
   - Prove the real upgrade topology: migrate through 0112, split/transfer roles, apply the privileged upgrade, then apply 0113 as the actual restricted migrator; verify owner/ACL preservation and rerun safety.

3. **Canonical lock ordering and real concurrency**
   - Deactivation, user deletion, and personal-tenant deletion must acquire locks in one order: user account -> tenant-owner lifecycle -> tenant row (with any tenant-delete fence ordered consistently).
   - Add deterministic two-connection PostgreSQL writer-first/delete-first and deadlock/race proofs. A single-connection `Promise.all` test is not acceptance evidence.

4. **Retained provenance/provider evidence**
   - Preflight and handle shared `ON DELETE RESTRICT` provenance (`workspaces.created_by`, provider role/grant actor columns, and other retained evidence).
   - Preserve or safely detach terminal provider/provenance evidence. Do not revoke externally and then return 500, silently cascade evidence, or make compliant deletion permanently impossible.

## Acceptance

- One current-develop draft PR closes this issue and all four gates.
- Mounted route tests plus real PostgreSQL upgrade/concurrency/provenance tests pass.
- Full exact-head CI/CodeQL is terminal green and an independent reviewer approves the final pushed SHA.
- No production activation or merge before those gates.

Refs #692, #675, #343, #654.

lalalune:
[...]

PR #807, opened 2026-08-22
test(auth): prove refresh lifecycle concurrency

### LINKED ISSUE #704: test(auth): replace refresh-token lifecycle source scans with real concurrency proofs
## Gap

On exact current develop `3559ac210`, the primary evidence for refresh-token reuse, user-wide revocation serialization, tenant switching, and successor revoke binding is still `auth-refresh-reuse.test.ts` plus `auth-refresh-revocation-race.test.ts`, which read `routes/auth.ts` as text and assert strings/order. The operator PostgreSQL suite proves the definer can rotate one authorized row, but it does not exercise the mounted `/auth/refresh`, `/auth/revoke`, `/auth/sessions`, or MFA revocation interleavings.

A rename/refactor can fail these tests without changing behavior, while a real lock/CAS/reuse regression can pass as long as the expected strings remain.

## Acceptance

- Replace the source scans with mounted route behavior against real PostgreSQL and the real revocation store.
- Deterministically barrier-test: two concurrent rotations of one token; rotate vs user-wide revoke; rotate vs single-session revoke; same-token reuse after winner commit; tenant switch vs membership removal/deactivation; MFA enable/disable vs refresh.
- Assert exact winner/loser statuses, successor binding, audit actions, tenant authority, refresh-row state, and revocation cutoff.
- Prove replay/reuse cannot mint a second live successor and cannot escape a concurrently established revocation.
- Keep the existing operator ACL proof as a separate database-authority contract.

lalalune:
Closes #704

Exact head: 7b9985cc603a2e5e2ddf9d2e41008a5bc8840c1b

Replaces refresh-token lifecycle source scans with mounted multi-replica behavior against real PostgreSQL and Redis. Deterministic barriers cover concurrent rotation, committed-token replay, user-wide and single-session revocation, tenant membership removal and deactivation, and MFA enable/disable against an in-flight refresh.

The production delta serializes membership/deactivation mutations with the same user-session advisory lock as refresh rotation and binds the definer timestamp argument explicitly as timestamptz.

Local exact evidence:
- API typecheck/build green
- targeted Biome, diff-check, and metadata green
- integration file registers 7 real-service gates and skips cleanly when DATABASE_URL/REDIS_URL are absent

Draft only. The local environment did not provide PostgreSQL/Redis, so hosted Integration Tests (Postgres), the full exact-head matrix, and independent approval are mandatory before readiness or merge.


lalalune:
Exact source audit BLOCK at `7b9985cc603a2e5e2ddf9d2e41008a5bc8840c1b`: the production delta adds `lockUserSession()` **after** `lockTenantOwnerLifecycle()` in platform membership removal and both user-admin deactivation/removal paths. That is opposite the canonical lifecycle order required by #696 (user session/account -> tenant-owner lifecycle -> tenant row/delete fence) and can deadlock against deletion/deactivation paths that acquire the user lock first. Preserve the real PG/Redis refresh proof, but reorder these mutations to the single canonical lock order and add/compose the deterministic lifecycle two-connection deadlock regression before publication. This PR must remain draft and should restack with the canonical #696 lane rather than independently cementing the inverse order.

lalalune:
Independent exact-head audit: ACCEPT at 7b9985cc603a2e5e2ddf9d2e41008a5bc8840c1b on base 201c1869d40ffca8a207a32a24eecc7eb6f24cd6. The same user-session advisory lock now serializes refresh rotation with tenant membership removal/deactivation and platform deactivation; refresh revalidates the token, user state, membership, and revocation state after acquiring it. The mounted fixture uses separate Bun processes plus real PostgreSQL/Redis, deterministic pg_stat_activity advisory-lock barriers, exact successor/replay/revocation state, and exact audit-action multisets. Exact hosted checks are terminal green (63/63). Ready is semantically justified, but merge remains blocked on REVIEW_REQUIRED; do not merge without the required approval.

PR #808, opened 2026-08-22
fix(api): make linked-account mutations audit-atomic

### LINKED ISSUE #702: security: make linked-account unlink mutation and audit atomic
## Gap

Exact current develop `3559ac210` still implements `DELETE /user/me/accounts/:provider/:providerAccountId` as:

1. revoke user tokens inside a DB transaction;
2. delete account/passkey and all refresh-token rows;
3. commit;
4. append the required `user.account.unlink` audit in a separate transaction;
5. on audit failure, reconstruct deleted rows from snapshots with `onConflictDoNothing()`.

`user-account-unlink-audit-rollback.test.ts` only source-scans that restoration sequence. It cannot prove behavior or concurrency. Once the first transaction commits, another relink/token rotation can occur before audit failure; whole-row restoration can silently skip current rows, reinsert obsolete refresh credentials, or produce a state that never existed. The external revocation cutoff is also irreversible and remains applied even when the route returns an audit failure.

## Acceptance

- Keep the irreversible revocation cutoff fail-closed, but perform the account/passkey deletion, refresh-token deletion, and required completion audit in one tenant/user-locked DB transaction after the cutoff is established.
- Remove snapshot/restore compensation for this path.
- Preserve last-login-method and exact linked-account revalidation under the same lock/transaction.
- Define and test the response/evidence contract when external revocation succeeds but the DB/audit transaction fails.
- Add mounted behavior plus deterministic real-PostgreSQL interleavings for concurrent relink/refresh rotation and injected completion-audit failure; prove no stale credentials are resurrected and no unaudited mutation commits.

No production carrier should be published until #657 protection is durable.

lalalune:
Closes #702

Exact head: 10cc70c889786871941b10735ebb46f504909387

Makes authenticated unlink, tenant remediation, and platform linked-account transfer execute the database mutation and required completion audit in one locked transaction after the irreversible session-revocation cutoff. It removes snapshot restoration, preserves last-login-method and tenant authority checks under the lock, and returns typed failure evidence when the cutoff succeeds but the database/audit unit fails.

The SDK exposes that failure evidence and documents the reauthentication contract. Deterministic PostgreSQL tests cover unlink versus refresh rotation, relink, and tenant remediation without stale credential resurrection.

Local exact evidence:
- mounted API suites: 25 pass, 3 expected PostgreSQL-only skips, 192 assertions
- SDK client: 143/143, 803 assertions
- SDK and API builds green
- targeted Biome, diff-check, and public-package metadata green

Draft only. Hosted exact-head PostgreSQL concurrency, full matrix, and independent approval remain required before readiness or merge.


lalalune:
Exact source audit at 7a9cf569: BLOCK; keep draft (independent of the still-running matrix's PostgreSQL failure).

This patch explicitly establishes `tenant audit advisory lock -> user-session lock` in `withPlatformLinkedAccountTransaction`, and the user unlink/remediation paths do the same by entering `withTenantAuditedTransaction` before `lockUserSession`. That is the reverse of the canonical lifecycle ordering required by the active #696 repair (`user -> owner/membership -> tenant`, then audit append in the same transaction). A lifecycle path holding the user/owner lock and appending the tenant audit can deadlock with this patch holding the tenant audit lock and waiting for the user.

Restack these linked-account mutations on the canonical lifecycle helper/order and add a deterministic two-connection race against personal/user deletion and tenant leave/deactivation. Do not preserve the current comment claiming audit-before-user is authoritative. The existing linked-account-only races cannot prove cross-lifecycle deadlock freedom.

lalalune:
Hosted exact-head evidence is now terminal red: PR Checks run `32547763496`, Integration Tests (Postgres) job `96969253196` failed `user-account-unlink-postgres-concurrency.integration.test.ts` (suite reports 1 pass / 2 fail). The visible failing remediation-vs-relink case expected the stable `{ accountUnlinked:false, sessionsRevoked:true }` failure envelope but the mounted route returned only `{ ok:false, error:"Internal server error" }`. This is consistent with the previously recorded lifecycle ordering/remediation blocker; 278/279 files passing does not accept the exact head. Keep draft and repair on the canonical lifecycle lock order with deterministic two-connection races.

lalalune:
Root cause fixed at `10cc70c88`: required audit writes now execute inside the same request-owned savepoint as the linked-account mutation. A forced audit failure can therefore roll back the atomic unit before the route returns its sanitized split-outcome evidence instead of aborting the outer tenant transaction. Exact local PostgreSQL 16 concurrency suite: 3/3 passed (15 assertions); audit-chain focused suites: 8/8; linked-account route suite: 22/22. The temporary diagnostic error exposure was reverted.

lalalune:
Exact-head audit BLOCK at `10cc70c889786871941b10735ebb46f504909387` (tree `1618dd725083dfbf9d93b04a8ca8ed355fcbafc4`). Unlink establishes the irreversible session cutoff and takes the user-session lock, but production OAuth link/upsert paths do not share that lock or recheck the cutoff. A link authenticated before unlink can finish provider exchange after cutoff and recreate/update the account. The PG test hides this by fabricating a relink transaction that explicitly calls `lockUserSession` instead of invoking production relink; it also uses a DDL-capable connection rather than restricted app-role middleware topology. Make every relink path participate in the same cutoff/lock protocol and prove it through a mounted two-connection race. Keep draft; 60 checks queued and no human review.

PR #810, opened 2026-08-22
fix(api): retire process-local tenant policy defaults

### LINKED ISSUE #711: fix(api): remove process-local tenant default-policy mutation
## Gap

Exact develop `3559ac210` keeps legacy `PUT /tenants/:id/webhook` active for `defaultPolicies`. After rejecting `webhookUrl`, it mutates only the process-local `tenantConfigs` map, appends the completion audit separately, and restores a copied map value if that audit fails.

Consequences:

- the update is not durable across restart and is not shared across API/Worker instances;
- concurrent updates can be lost when one audit failure restores an older whole config over a newer winner;
- `tenant-audit-rollback.test.ts` only source-scans the compensation sequence;
- the route name/deprecation contract obscures that it still carries an active policy mutation.

## Acceptance

Either fully deprecate this legacy endpoint and direct callers to the durable tenant-config API, or migrate default-policy writes to the canonical database-backed tenant-config service. In either case:

- no production write depends on a process-local map as authority;
- mutation + required audit are one transaction/CAS;
- cache invalidation happens only post-commit;
- mounted multi-instance/restart behavior and concurrent audit-failure races are tested;
- replace `tenant-audit-rollback.test.ts` and legacy source scans with reachable behavior;
- update SDK/docs/OpenAPI callers and current-state comments.

Keep local-only until #657 is durable.

lalalune:
## Summary

- reject legacy tenant defaultPolicies writes before database, cache, or audit mutation
- make durable agent policy rows the sole policy authority and remove retired fields from tenant reads
- remove remaining CLI, example, and dashboard callers of the retired tenant webhook/default-policy carrier
- direct operators to the signed webhook control plane and durable policy APIs

## Local validation

- tenant policy/deprecation API selection: 14 pass, 11 expected service/real-Postgres skips, 128 assertions
- CLI tests: 61/61
- API, CLI, waifu example, React package, and web typechecks/builds: pass
- targeted Biome, public-package metadata, and git diff check: pass

Hosted real-Postgres and the full protected matrix remain required. This PR stays draft until exact-head checks and independent review complete.

Closes #711

## Exact lineage

Head: 0d4daa54807cf33214402d4f2d481bb0b3765b2f


lalalune:
Exact-head correction published at `9f40f7db8340e931e9dc1a1e4262c10a1100b689`.

The public `defaultPolicies` carrier is removed from shared `TenantConfig`, platform create request/response types, and docs. Tenant create/update and platform create/PUT now share one exact 410 retirement contract. A mounted platform regression proves the rejected legacy field creates no tenant and leaves audit rows byte-equivalent.

Local exact-head evidence: platform 6/6 (18 assertions), tenant retirement 3/3 (15 assertions), Shared/API TypeScript green, Biome and diff-check clean. Hosted checks and independent approval remain required.

lalalune:
Exact-head audit BLOCK at live `9f40f7db8340e931e9dc1a1e4262c10a1100b689` (tree `5ef96fecd35991f17788056b96169e294e264b2b`). The source retirement direction is sound, but the newly changed gated condition-set test expects an empty durable policy set to approve with no results, directly contradicting the policy engine’s fail-closed contract. The claimed multi-instance/restart proof is two Hono objects in one process sharing one imported module/map; restart is only a map deletion. There is also no restricted-role real-PG proof that tenant A cannot resolve tenant B policy, even though `getPolicySet` ignores its tenant argument and relies on RLS fencing. Fix the red expectation, run the gated suite, and add separate-process restricted-role PG evidence. Body lineage is stale (`0d4daa…` vs live `9f40f7…`); zero reviews and nearly all checks queued. Keep draft.

lalalune:
Closed as a rejected stale carrier. Empty policy state is still treated inconsistently with fail-closed semantics and restart or restricted-RLS evidence is not production-real.

PR #816, opened 2026-08-22
test(api): mount secret route validation

### LINKED ISSUE #733: test(api): execute secret route validation and persistence boundaries
## Gap

On exact develop `3559ac210`, `packages/api/src/__tests__/secret-routes-validation.test.ts` only reads `routes/secrets.ts` and `@stwd/vault/secret-route-validator.ts` as text. It claims oversized-format, header-only injection, type validation, line-break rejection, header-name validation, persistence ordering, and no-store behavior without mounting either create/update route or inspecting stored rows.

## Acceptance

- Replace the source scan with migrated-PGLite mounted behavior for both secret-route create and update entrypoints.
- Table-drive query/body injection modes, oversized formats, invalid header names, non-integer priority, non-boolean enabled, and CR/LF secret values.
- Assert stable 400s, zero secret/route/audit mutation, and no vault persistence on every invalid case.
- Prove a normalized valid header route persists exactly once and both inventory/topology responses carry the full no-store header set.
- Retain a narrow whole-tree absence assertion only if needed to prove no hidden query/body opt-in exists; do not use source order as behavioral evidence.

lalalune:
## Summary

- preserve the mounted validator matrix and add CR/LF coverage for POST /secrets/:id/rotate
- mount core idempotency and production routes through createApp against PostgreSQL
- use real signed owner, member, and stale-MFA JWTs with seeded tenant memberships and production tenantAuth
- execute secret create, PUT rotate, POST rotate, route create, and route update
- prove no-store/non-disclosure, zero persistence on denied/invalid requests, exact authorized/completion actors, and the full ten-event audit chain/head

## Exact local evidence

- focused suites: 5/5, 95 assertions
- API TypeScript no-emit: pass
- targeted Biome and diff check: clean
- PostgreSQL suite registers 3 tests and skipped locally only because DATABASE_URL was unavailable

This stays draft until the exact hosted Integration PostgreSQL job executes the mounted suite, the full matrix is terminal, and independent review is complete.

Closes #733

## Exact lineage

Head: 60154282e6d693017ca0d2cfdcdf8b1d4b108701
Tree: 9fb8160964bff6e969c4ec276b7e638c872d3b6d

### DIFF
--- a/packages/api/src/__tests__/secret-routes-production-postgres.integration.test.ts
+++ b/packages/api/src/__tests__/secret-routes-production-postgres.integration.test.ts
@@ -1,0 +1,282 @@
+import { afterAll, beforeAll, describe, expect, it, setDefaultTimeout } from "bun:test";
+import {
+  __resetAuditHmacKeyCacheForTests,
+  agents,
+  auditChainHeads,
+  auditEvents,
+  getDb,
+  secretRoutes,
+  secrets,
+  tenants,
+  users,
+  userTenants,
+} from "@stwd/db";
+import { and, asc, eq, isNull } from "drizzle-orm";
+import type { Hono } from "hono";
+import { verifyAuditChain } from "../services/audit";
+import type { AppVariables } from "../services/context";
+
+setDefaultTimeout(120_000);
+
+const hasDatabaseUrl = Boolean(process.env.DATABASE_URL?.trim());
+const describeWithPostgres = hasDatabaseUrl ? describe : describe.skip;
+const SUFFIX = crypto.randomUUID();
+const TENANT_ID = `secret-production-${SUFFIX}`;
+const AGENT_ID = `secret-production-agent-${SUFFIX}`;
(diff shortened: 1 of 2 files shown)


lalalune:
Exact-head evidence audit of 8deaaec0c1ab90990010e4dc878638f23de895e5 is BLOCK. The test imports the real secrets router after migrated PGLite and its direct route assertions are useful, but it does not mount the production app or tenantAuth: a custom middleware manually injects tenant ID, session role, and MFA variables. It therefore cannot prove reachability through app.ts auth, real JWT/session handling, or the request-bound outer tenant transaction/savepoint. It also omits the third secret-value writer POST /secrets/:id/rotate, so CR/LF validation can regress on that alias while collection POST and PUT stay green. Audit verification checks only a +2 row count, not the exact authorized/completion actions or chain. Add full-app production-auth mounted success/failure coverage, rotate CR/LF cases, and exact audit action/chain assertions. I restored draft because the body requires it and hosted/review gates are incomplete; do not merge.

PR #817, opened 2026-08-22
test(api): mount Bun readiness behavior

### LINKED ISSUE #734: test(api): make Bun readiness auth-store checks behaviorally testable

lalalune:
## Summary

- extract the Bun readiness handler from process startup while preserving entrypoint startup order and production dependency wiring
- mount readiness with injected database, Redis, proxy, Vault, and auth-store health dependencies
- prove every memory-backed production/Workers auth store fails 503 unless explicitly acknowledged
- redact public diagnostics, use constant-time probe-token authorization, bound operator details to the supplied checks, and make every response no-store

## Local validation

- readiness and middleware hardening suites: 22/22, 120 assertions
- API typecheck, targeted Biome, public-package metadata, and git diff check: pass

This PR stays draft until the exact-head hosted matrix and independent review complete.

Closes #734

## Exact lineage

Head: e923a4091dc156df22b8cde5ee0b18b4a1f0d0ce

### DIFF
--- a/packages/api/src/__tests__/auth-import-session-store-sources.test.ts
+++ b/packages/api/src/__tests__/auth-import-session-store-sources.test.ts
@@ -1,6 +1,12 @@
 import { afterEach, describe, expect, it, mock } from "bun:test";
-import { readFileSync } from "node:fs";
-import { join } from "node:path";
+import { Hono } from "hono";
+import {
+  createReadinessHandler,
+  type ReadinessDependencies,
+  type ReadinessEnvironment,
+} from "../readiness";
+import type { AuthStoreSources } from "../routes/auth";
+import type { AppVariables } from "../services/context";
 
 const redisSetMock = mock(async () => "OK");
 const redisGetMock = mock(async () => null as string | null);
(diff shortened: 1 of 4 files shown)


lalalune:
Exact-head source audit of `e923a4091dc156df22b8cde5ee0b18b4a1f0d0ce` is **ACCEPT**. The extraction preserves Bun entrypoint ordering and production dependency wiring: the mounted handler receives the real DB/migration/RLS, Redis, proxy, Vault, auth-store, and optional scheduler checks, while `migrationsRan` remains bound to boot completion. Public responses expose only bounded ok/required flags, every path sets no-store headers before dependency work, probe-token comparison is constant-time over fixed-length hashes, and each memory-backed auth-store case fails readiness in durable runtimes unless explicitly acknowledged. The new tests exercise the Hono route rather than source text and cover dependency rejection/redaction/optional checks. Keep draft until this exact head has terminal hosted checks and independent required review; this is source acceptance, not permission to merge.

PR #818, opened 2026-08-22
fix(api): prove sponsored transfer accounting

### LINKED ISSUE #735: test(api): replace gas sponsorship source-order claims with mounted accounting proofs
## Gap

On exact develop `3559ac210`, the final three cases in `gas-sponsorship-config.test.ts` read `services/gas-sponsorship.ts` and `routes/vault.ts` as text to claim reservation-before-signing, manual-approval reservation/finalization, rollback/failure handling, and signed-only rejection. These are money-path concurrency and irreversible-I/O contracts; string ordering cannot prove them.

## Acceptance

- Retain the pure normalization unit cases, but replace the three source-order cases with mounted route behavior against migrated PGLite plus deterministic real PostgreSQL where locking matters.
- Fault-inject signer and required-audit failures and prove sponsorship reservation state/value, transaction row, and audit rows remain consistent with zero premature signing.
- Drive manual approval queue -> approval -> broadcast and assert one reservation/finalization and correct persisted sponsorship metadata.
- Prove signed-only sponsored transfer rejects before reservation/provider/signing, including replay.
- Race concurrent requests at the tenant/agent cap and prove one bounded winner without double reservation.

Coordinate with #701 durable spend effects and #717 policy-template work; keep local-only until #657 confirms durable branch protection.

lalalune:
## Summary

- replace gas-sponsorship source-order claims with mounted transfer, approval, signing-fault, audit-fault, replay, and cap-race behavior
- stage sponsored transaction authority before signing and keep reservation/failure state consistent
- serialize PGLite reservations per tenant while production PostgreSQL uses the advisory transaction lock
- make reservation upserts cap-neutral for the same existing sponsorship record

## Local validation

- gas sponsorship suites: 10 pass, 1 explicit real-Postgres skip, 51 assertions
- API typecheck, targeted Biome, public-package metadata, and git diff check: pass

The local environment did not provide PostgreSQL, so the two-connection advisory-lock race remains an exact-head hosted gate. This PR stays draft until that proof and independent review complete.

Closes #735

## Exact lineage

Head: 670dd887875d9cde7d5878671057fda32bcbd059

### DIFF
--- a/packages/api/src/__tests__/gas-sponsorship-config.test.ts
+++ b/packages/api/src/__tests__/gas-sponsorship-config.test.ts
@@ -1,6 +1,4 @@
 import { describe, expect, it } from "bun:test";
-import { readFileSync } from "node:fs";
-import { join } from "node:path";
 import {
   normalizeGasSpendQuery,
   normalizeGasSponsorshipConfig,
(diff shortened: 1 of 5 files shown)


lalalune:
Exact source audit at 670dd887: BLOCK; keep draft.

The new PostgreSQL test does not prove the claimed two-connection advisory-lock race. `Promise.all(reserveSponsoredGasEvent(...))` has no backend-PID assertion or synchronization point inside two open transactions. The same one-winner result occurs under completely sequential execution because the second reservation simply observes the first row. Add a deterministic two-connection overlap/barrier (and distinct `pg_backend_pid()` evidence) that would exceed the cap without the production lock.

There is also a crash-consistency gap in the mounted transfer path. The sponsored `transactions` row is inserted, then `reserveSponsoredGasEvent` commits in a separate transaction, and only afterward does signing begin. `actionId` is random for this EVM path. A process death after either commit strands a pending transaction and/or a cap-consuming `reserved` sponsorship record; a retry gets a new action ID and cannot reconcile the old reservation. No expiry/reconciler is present in this patch. Make staging + reservation one durable atomic transition keyed to stable request identity, or add authoritative recovery/expiry that proves a crash cannot permanently consume cap or double-authorize spend.

Hosted green alone will not close these semantic gates.

lalalune:
Exact-head audit BLOCK at `670dd887875d9cde7d5878671057fda32bcbd059` (tree `fc49c6267ea3e12843cd3ce4dc01d331c0707bb0`). Sponsorship is authorized before several awaits, then the locked reservation path rechecks only spend caps. It does not revalidate enabled state, circuit breaker, client permission, provider/mode, allowed chain, or `maxPerTxUsd`; an admin revocation or limit reduction can race the stale request and still reach reservation/signing. Mounted coverage uses PGLite/injected auth, while real-PG invokes the reservation service directly under a broad role and proves cap serialization only. Add a deterministic mounted revocation race and revalidate the full current authority inside the locked transaction. All 62 checks are green, but review is absent; keep draft.

PR #820, opened 2026-08-22
fix(security): make secret CRUD audit-atomic

### LINKED ISSUE #736: security: make secret create, rotate, and delete audit-atomic
## Defect

Exact develop `3559ac210` still leaves core secret CRUD on explicit snapshot/compensation while secret-route CRUD is transaction-atomic. `secrets-audit-order.test.ts` acknowledges this in source comments and only searches route text. Create, rotate, and delete perform Vault/database mutation and then required completion audit; rollback reconstructs/deletes rows afterward. A concurrent route/version mutation can be overwritten or resurrected by stale compensation, and a process death between mutation and compensation leaves state without required evidence.

## Acceptance

- Add transaction-bound SecretVault create/rotate/delete primitives and execute each DB mutation plus required completion audit in one tenant transaction.
- Establish one canonical tenant -> secret -> route/version lock order shared by create, rotate, delete, and route CRUD.
- Re-read ownership/status under lock; return stable 404/409 for stale races.
- Preserve authorization audits before sensitive mutation; invalidate/refresh any caches only after commit.
- Replace `secrets-audit-order.test.ts` source claims for these paths with mounted PGLite audit-failure rollback and deterministic real-PostgreSQL rotate/delete/route races.
- Prove no plaintext read-back, exact version lineage, no orphan/resurrected route, and one completion audit for the winner.

Coordinate with #733 validation tests. Keep local-only until #657 confirms durable branch protection.

lalalune:
Closes #736.

Moves secret create, rotate, and delete plus their audit events into the same tenant transaction. Adds canonical lineage locking, transaction-aware Vault methods, authorization/audit failure coverage, and a file-backed PGLite regression for the supported lock path.

Exact head: 80ec2ac50fa7f3d120d0a5804c06cb02a19eaca9

Local acceptance:
- API secret atomicity/order: 6 pass, 1 real-Postgres skip
- Vault lifecycle/no-read-back: 15 pass
- API build: pass
- Vault build: pass

Hosted acceptance still requires the mounted real-Postgres concurrency/rollback gate and the full exact-head matrix. Keep draft until required checks and independent review are complete.


lalalune:
Independent exact-head audit of `80ec2ac5` is **BLOCK**.

The new `/secrets` transaction path is stronger, but the claimed canonical lineage lock is not shared by all production writers. KMS still calls legacy `createSecret`/`rotateSecret`; legacy rotation reads current state before its transaction and never takes the `steward_secret_*` advisory/row lock, then audits after commit. Concurrent KMS/route operations can produce duplicate-version/non-atomic evidence and can leave a newly inserted route pointing at soft-deleted v1 when KMS repoints routes before the route insert lands. Legacy `deleteSecret` remains an unlocked public path too.

There is also a governed-authority bypass: generic route DELETE does not reject `governed_v2`, and generic secret deletion removes every route in the lineage, including governed/provider routes, without the provider-operation disconnect/revocation lifecycle; provider/account state can continue referencing a soft-deleted credential.

The PG fixture injects Hono variables instead of mounting `tenantAuth`, so it does not prove production's outer transaction/savepoint or authorization-receipt semantics, and it omits KMS/provider cross-writer races.

Require legacy APIs to delegate to the locked primitives (or remove their production callers), a governed delete/repoint authorization fence, and mounted tenantAuth multi-connection KMS/provider/route races with audit chain/head receipts. Keep draft/unmerged.

lalalune:
Exact-head audit BLOCK at `80ec2ac50fa7f3d120d0a5804c06cb02a19eaca9` despite 62/62 hosted checks. The HTTP secret CRUD paths now use transaction-bound primitives and the focused rollback/race evidence is materially stronger, but the required canonical lock order is not shared by every production writer. `SecretVault.createSecret`, `rotateSecret`, and `deleteSecret` remain public unlocked implementations, and production callers such as `provider-x-connect.ts` and `routes/kms.ts` still invoke them directly. Those writers can race the newly locked route/lineage paths, repoint/delete routes under a different order, or create version lineage without the canonical advisory/row locks. Migrate every production writer to the transaction-bound canonical path (or make the legacy methods internally acquire the same locks) and add a hostile cross-entrypoint race. Zero reviews; keep draft.

lalalune:
Closed as a rejected stale carrier. Legacy production SecretVault writers bypass the canonical lock and audit-atomic mutation path.

PR #823, opened 2026-08-22
fix(api): make ERC-8004 registration audit-atomic

### LINKED ISSUE #718: security: make ERC-8004 registration audit-atomic
Exact develop `3559ac210` upserts `agent_registrations`, appends required `erc8004.register` audit afterward, and restores a prior registration snapshot on audit failure. A concurrent registration update after the upsert commit can be overwritten/deleted by the stale compensation. Move the locked tenant/agent/chain upsert and completion audit into one tenant-audited transaction; remove `restoreAgentRegistration`; keep authorization audit and SSRF/API URL validation before mutation. Add mounted plus deterministic two-connection PostgreSQL update-vs-update/audit-failure tests, asserting the concurrent winner and exact audit remain. Local-only until #657 is durable.

lalalune:
Closes #718.

Moves the locked ERC-8004 registration upsert and required completion audit into one tenant-audited transaction and removes stale snapshot compensation. Adds mounted deterministic PostgreSQL rollback and concurrent-winner proofs.

Exact head: cbcb0e34b2f6b8bdd7215d17167bab0a48f1879e

Local acceptance:
- focused audit ordering: 8 pass, 38 assertions
- mounted PostgreSQL: 2 locally skipped and required in hosted Integration Tests
- API build, targeted Biome, diff check, public-package metadata: pass

### DIFF
--- a/packages/api/src/__tests__/erc8004-audit-atomicity-postgres.integration.test.ts
+++ b/packages/api/src/__tests__/erc8004-audit-atomicity-postgres.integration.test.ts
@@ -1,0 +1,304 @@
+import { expect, it } from "bun:test";
+import {
+  __resetAuditHmacKeyCacheForTests,
+  agentRegistrations,
+  agents,
+  auditChainHeads,
+  auditEvents,
+  createDb,
+  tenants,
+} from "@stwd/db";
+import { and, eq } from "drizzle-orm";
+import { Hono } from "hono";
+import { correlationId } from "../middleware/correlation";
+
+const databaseUrl = process.env.DATABASE_URL;
+const realPostgresIt = databaseUrl && !process.env.STEWARD_PGLITE_MEMORY ? it : it.skip;
+const registryAddress = "0x0000000000000000000000000000000000008004";
+
+realPostgresIt(
+  "rolls back a mounted ERC-8004 registration when its required audit fails",
+  async () => {
+    const suffix = crypto.randomUUID().replaceAll("-", "");
+    const tenantId = `erc8004-atomic-${suffix}`;
+    const agentId = `erc8004-agent-${suffix}`;
+    const requestId = `failed-${suffix}`;
+    const triggerFunction = `fail_erc8004_audit_${suffix}`;
+    const triggerName = `fail_erc8004_audit_${suffix}`;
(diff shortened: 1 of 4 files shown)


lalalune:
Independent exact-head audit of `cbcb0e34`: **source logic ACCEPT, acceptance evidence HOLD**.

The registration upsert and required completion audit now share one tenant-audited transaction; the pre-mutation authorization receipt remains separate, and this route performs no external/on-chain side effect.

The mounted PostgreSQL proof does not, however, mount `tenantAuth` or the middleware-owned outer tenant transaction. Its app only sets Hono variables, so it never executes production's `hasTenantTransactionDatabase` savepoint/raw-outer-audit path. The concurrent “winner” is a direct `withTenantAuditedTransactionOnDb` subprocess rather than a second authenticated mounted route and omits the winner's authorization receipt. Global advisory-waiter counting can also match unrelated sessions.

Before acceptance, mount the real tenant middleware and use two independent authenticated route connections; assert failed authorized-only plus winner authorized/completion events, chain/head integrity, and scoped cleanup. Keep draft until that exact real-PG topology proof, full matrix, and review are green.

lalalune:
Independent exact-head audit: BLOCK at cbcb0e34b2f6b8bdd7215d17167bab0a48f1879e (tree ec67aafbb7120f282158edd3fdd2f3fc5519bffd). The atomic mutation/completion-audit repair is sound, but authority binding remains TOCTOU and cross-tenant unsafe: the route resolves agent and wallet outside the transaction, then upserts tenantId/agentId without locking/reselecting the agent or using an insert-select constrained to the same tenant. agent_registrations has separate tenant and agent foreign keys, not a composite tenant-agent FK. Delete the tenant-A agent after preflight and recreate the global ID under tenant B, or rotate its wallet, and the route can commit a tenant-A registration bound to tenant-B/stale authority. Fix with in-transaction locked composite lookup/insert-select plus composite FK and hostile delete-reuse/wallet-rotation races. Tests inject owner/MFA manually and use unrestricted DATABASE_URL for both route and writer, so they do not prove production tenant middleware, restricted app role, or RLS. #718 linkage is correct. Keep draft.

PR #824, opened 2026-08-22
fix(api): make webhook mutations audit-atomic

### LINKED ISSUE #719: security: make webhook configuration and retry mutations audit-atomic
## Gap

Exact `origin/develop` `3559ac210` still commits webhook authority/delivery mutations before required completion audits and compensates afterward in `packages/api/src/routes/webhooks.ts`:

- create inserts a disabled credential-bearing webhook, audits separately, deletes on failure, then enables in another write;
- update commits the new URL/events/secret-related configuration, audits separately, then whole-row restores;
- delete commits, audits separately, then reinserts the encrypted-secret snapshot;
- delivery retry status mutation similarly commits before audit and restores stale delivery state.

A concurrent legitimate update/delete/recreate or delivery worker transition can land after the first commit and then be overwritten or collide with compensation. Create also exposes a partially initialized row and does not bind enablement to the required audit. Existing behavior tests do not deterministically exercise the post-commit interleavings.

## Acceptance

- Re-read and lock the tenant/webhook row under one audited transaction; commit create/update/delete/retry state and required completion audit atomically.
- For create, return the one-time secret only after the audited row is committed in its final enabled state; never publish a partially initialized authority row.
- Remove delete/reinsert and whole-row restore compensation.
- Preserve URL resolution/SSRF validation, quota and duplicate-URL serialization, MFA/auth gates, secret redaction, and authorization-audit ordering.
- Add mounted behavior plus deterministic real-PostgreSQL races for create-vs-create, update-vs-update/delete, delete-vs-recreate, retry-vs-worker transition, and injected completion-audit failure. Prove one audited winner, no stale encrypted-secret resurrection, and exact delivery status.

lalalune:
Closes #719.

Moves webhook create, update, delete, and retry state changes plus required completion audits into locked audited transactions. Removes stale compensation, returns one-time secrets only after the final enabled row commits, and adds mounted PostgreSQL create/update/delete/retry races. The PostgreSQL suite now registers safely when DATABASE_URL is absent.

Exact head: cfd601b20b60783134401f7a8dca990dc695586a

Local acceptance:
- webhook behavior/order: 23 pass, 162 assertions
- mounted PostgreSQL races: 10 locally skipped and required in hosted Integration Tests
- API build, targeted Biome, diff check, public-package metadata: pass

### DIFF
--- a/packages/api/src/__tests__/fixtures/webhook-concurrent-request.ts
+++ b/packages/api/src/__tests__/fixtures/webhook-concurrent-request.ts
@@ -1,0 +1,49 @@
+import { closeDb } from "@stwd/db";
+import { Hono } from "hono";
+import { correlationId } from "../../middleware/correlation";
+import type { AppVariables } from "../../services/context";
+
+function required(name: string): string {
+  const value = process.env[name];
+  if (!value) throw new Error(`${name} is required`);
+  return value;
+}
+
+const tenantId = required("TEST_TENANT_ID");
+const userId = required("TEST_USER_ID");
+const requestId = required("TEST_REQUEST_ID");
+const method = required("TEST_REQUEST_METHOD");
+const path = required("TEST_REQUEST_PATH");
+const requestBody = process.env.TEST_REQUEST_BODY;
+
+const [{ createSessionToken }, { tenantAuth }, { webhookRoutes }] = await Promise.all([
+  import("../../routes/auth"),
+  import("../../services/context"),
+  import("../../routes/webhooks"),
+]);
+
+const token = await createSessionToken("0x0000000000000000000000000000000000000719", tenantId, {
+  userId,
+  tenantId,
+  mfaVerifiedAt: Date.now(),
+  mfaMethod: "totp",
+});
+const app = new Hono<{ Variables: AppVariables }>();
+app.onError((error, c) => c.json({ ok: false, error: error.message }, 500));
+app.use("*", correlationId);
+app.use("*", tenantAuth);
+app.route("/webhooks", webhookRoutes);
+
+const response = await app.request(path, {
+  method,
+  headers: {
+    Authorization: `Bearer ${token}`,
+    "Content-Type": "application/json",
+    "X-Request-Id": requestId,
+  },
+  ...(requestBody === undefined ? {} : { body: requestBody }),
+});
+const body = await response.json();
+await closeDb();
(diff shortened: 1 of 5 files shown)


lalalune:
Independent exact-head audit of `cfd601b2` is **BLOCK**.

1. Production worker dispatch is not fenced against config update/delete. `PersistentWebhookQueue` commits its delivery claim, then reads the enabled config without a retained lock and performs network I/O. A delete can commit and return 200 after that read but before the old worker sends; update can likewise permit a post-commit delivery to the retired URL.

2. Expired-claim/manual-retry outcomes have no claim epoch/token. Manual retry can rewrite expired `processing` to pending, while the old worker's terminal success/failure updates are unconditional by delivery ID. A stale worker can overwrite the new claim or re-send concurrently.

3. The mounted PG “worker” race is a raw status UPDATE, not `PersistentWebhookQueue.processQueue`; it never proves the production claim predicate, visibility timeout, config read, network dispatch, or budget boundary.

4. Authorization-attempt receipts regress: `.authorized` events were moved into the same savepoint as completion, and the failure test expects zero request audits. A failed sensitive attempt now disappears instead of retaining its authorization decision.

Require claim-token/epoch CAS on every terminal outcome and manual retry, a config-generation/in-flight fence preventing old-authority dispatch after update/delete, real production-queue multi-connection races, and a durable authorization-attempt receipt. The route-local mutation/audit transactions are otherwise sound. Keep draft/unmerged.

lalalune:
Independent exact-head audit: BLOCK at cfd601b20b60783134401f7a8dca990dc695586a (tree a5b5ce0cd625a1618eff617ca4793b6d2b2d1a9b). P1 config-epoch TOCTOU: the persistent worker claims a delivery, later reads a matching enabled config, and then dispatches the snapshotted old URL/secret without holding a config lock or epoch through network submission. Concurrent PUT/DELETE can commit after validation but before submit, so a revoked/changed endpoint still receives the payload. Route retry FOR SHARE ends at retry commit and carries no epoch into the later worker claim. The new PG race fakes worker completion with a raw UPDATE and never executes production claim/config/dispatch. Tests also reuse one privileged DATABASE_URL for app, writers, and fixture DDL, so restricted app-role/cross-tenant RLS are unproven. Mutation plus authorized/completion audit rollback itself is sound. #719 linkage is correct. Keep draft.

PR #825, opened 2026-08-22
test(api): prove secret-bearing responses are non-cacheable

### LINKED ISSUE #732: test(api): prove secret-bearing responses are non-cacheable at runtime
## Gap

On exact develop `3559ac210`, `packages/api/src/__tests__/secret-bearing-cache-control.test.ts` only reads route source and compares string positions for app-client secrets, invitation tokens, Vault/user/global-wallet signatures, audit/dashboard reads, and auth token responses. This can pass while a route returns secret-bearing JSON before headers are applied or through a different response branch, and it fails harmless refactors.

Some individual routes have real header assertions, but the source-only matrix still claims coverage for many signing and token branches without executing them.

## Acceptance

- Replace the source scan with mounted table-driven behavior for every reachable secret-bearing response family it claims.
- Assert `Cache-Control: no-store, max-age=0`, `Pragma: no-cache`, and `Expires: 0` on successful one-time secret/token/signature responses and relevant error responses after authorization.
- Cover app-client secret rotation, both invitation-token entrypoints, representative Vault signature modes (including Solana and typed/raw/message paths), user/global-wallet signing, audit/dashboard reads, and auth identity/refresh/OAuth token responses.
- Reuse existing mounted suites where they already prove a branch; delete redundant source assertions instead of duplicating fixtures.
- Include a branch inventory so newly added secret-bearing response types cannot silently omit the contract.

lalalune:
Closes #732.

Replaces the secret-bearing cache source scan with a mounted 19-family runtime inventory. Exercises actual app-client and invitation secrets, Vault signature modes, user/global-wallet signing, audit/dashboard reads, and auth token responses; production-disabled user-operation and authorization branches are represented by mounted post-auth fail-closed evidence rather than fabricated success. Adds route-level no-store coverage where reachable responses lacked it.

Exact head: 55fd0b96b0e00ff4090c08045ac459e720a44481

Local acceptance:
- runtime inventory: 19 pass, 193 assertions
- isolated supporting files: 13 pass
- API dependency build: 24/24
- dashboard mounted PostgreSQL: 5 locally skipped and required in hosted Integration Tests

### DIFF
--- a/packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts
+++ b/packages/api/src/__tests__/auth-oauth-nonce-exchange.test.ts
@@ -62,13 +62,17 @@ function makeApp(): Hono {
 async function postExchange(
   app: Hono,
   body: Record<string, unknown>,
-): Promise<{ status: number; json: Record<string, unknown> }> {
+): Promise<{ status: number; json: Record<string, unknown>; headers: Headers }> {
   const res = await app.request("/auth/oauth/exchange", {
     method: "POST",
     headers: { "Content-Type": "application/json" },
     body: JSON.stringify(body),
   });
-  return { status: res.status, json: (await res.json()) as Record<string, unknown> };
+  return {
+    status: res.status,
+    json: (await res.json()) as Record<string, unknown>,
+    headers: res.headers,
+  };
 }
 
 describe("POST /auth/oauth/exchange", () => {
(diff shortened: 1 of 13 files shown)


lalalune:
Exact-head source audit of `55fd0b96b0e00ff4090c08045ac459e720a44481` is **ACCEPT / hosted-PG HOLD**. The 19-family inventory matches the prior secret-bearing surface, and the reachable success proofs exercise actual token/signature/dashboard responses rather than source-only markers. The two production changes correctly move no-store policy ahead of authenticated validation for all global-wallet RPC responses and both user-wallet signing routes; tenant-admin invitation creation sets it immediately after the authorization boundary, before token generation or validation errors. Existing auth/audit/Vault/platform/app-client policies remain mounted and asserted with the full Cache-Control/Pragma/Expires contract. The PostgreSQL dashboard proof is intentionally skipped locally, so do not promote from source evidence alone: keep draft until exact-head hosted Integration Tests, the full matrix, and required review are terminal.

lalalune:
Exact-head audit BLOCK at `55fd0b96b0e00ff4090c08045ac459e720a44481` despite 62/62 hosted checks. The mounted behavior added for the listed families is useful, but the required completeness inventory is still a hard-coded 19-entry list. Its only enforcement reads named test source files and searches for generic status/header strings; adding a new secret-bearing response family silently leaves the list at 19 and passes, and unrelated assertions in the named file can satisfy an entry. That preserves the core source-claim weakness and does not meet #732 acceptance that newly added secret response types cannot silently omit the contract. Derive the inventory from the production route/response registry or another authoritative typed manifest that production additions must update, then bind each entry to executable route-specific evidence. Zero reviews; keep draft.

PR #826, opened 2026-08-22
fix(api): make platform tenant creation audit-atomic

### LINKED ISSUE #712: security: make platform tenant creation audit-atomic
Exact develop `3559ac210` creates the tenant row and publishes it into `tenantConfigs`, then appends required `tenant.create` audit separately. On audit failure it deletes the cache entry and deletes the tenant as compensation. `tenant-audit-rollback.test.ts` only source-scans that order. Between commit and compensation, another request/job can observe the tenant and create dependent authority/data; the compensating delete can cascade or fail, leaving either unaudited state or deleting a concurrent winner. Move tenant creation + required completion audit into one platform-authority audited DB transaction, publish cache only after commit, and remove delete compensation. Add mounted and deterministic real-PostgreSQL audit-failure/concurrent-first-use tests; prove no tenant/dependent rows become visible before the audited commit. Keep authorization audit ordering and platform scope gates. Local-only until #657 is durable.

lalalune:
Closes #712.

Moves platform and legacy tenant creation, API-key custody evidence, and required completion audits into a single audited database transaction; publishes tenant cache state only after commit and removes compensating deletion. Adds a mounted PostgreSQL completion barrier with independent observer and dependent child writer to prove invisibility before commit, coherent success, and full rollback on audit failure.

Exact head: 52d64d1ddcc220015afc1fbb988b86673f35c024

Local acceptance:
- focused tenant/platform suites: 3/3 file processes pass
- API dependency build: 24/24
- mounted PostgreSQL proof: locally skipped because DATABASE_URL is unavailable and required in hosted Integration Tests

### DIFF
--- a/packages/api/src/__tests__/fixtures/tenant-create-dependent-writer.ts
+++ b/packages/api/src/__tests__/fixtures/tenant-create-dependent-writer.ts
@@ -1,0 +1,30 @@
+import { agents, createDb } from "@stwd/db";
+
+const required = (name: string): string => {
+  const value = process.env[name]?.trim();
+  if (!value) throw new Error(`${name} is required`);
+  return value;
+};
+
+const databaseUrl = required("DATABASE_URL");
+const tenantId = required("TEST_TENANT_ID");
+const agentId = required("TEST_AGENT_ID");
+const database = createDb(databaseUrl);
+
+try {
+  await database.db.insert(agents).values({
+    id: agentId,
+    tenantId,
+    name: agentId,
+    walletAddress: "0x7120000000000000000000000000000000000000",
+  });
+  console.log(JSON.stringify({ ok: true }));
+} catch (error) {
+  const code =
+    typeof error === "object" && error !== null && "code" in error
+      ? String((error as { code?: unknown }).code ?? "unknown")
+      : "unknown";
+  console.log(JSON.stringify({ ok: false, code }));
+} finally {
+  await database.client.end();
+}
--- a/packages/api/src/__tests__/platform-security-hardening.test.ts
+++ b/packages/api/src/__tests__/platform-security-hardening.test.ts
@@ -110,11 +110,11 @@ describe("platform security hardening", () => {
       tenantCreateStart,
       platformSource.indexOf('platform.get("/tenants"', tenantCreateStart),
     );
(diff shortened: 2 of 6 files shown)


lalalune:
Independent exact-head audit of `52d64d1d` is **BLOCK (P1)** on tenant-ID retained-evidence reuse.

Both creation routes acquire only the tenant audit advisory lock, then check `secrets`, `secret_routes`, `proxy_audit_log`, and `audit_events` before inserting the tenant. That lock serializes same-ID creators and audit-chain writers, but it does not fence raw retained-state writers. Proxy audit writes directly insert `proxy_audit_log`; that immutable table deliberately has no tenant FK, and an old in-flight proxy completion can land after its tenant was deleted.

Concrete interleaving: creator holds the audit lock and sees no retained rows; an old in-flight completion inserts retained proxy evidence after the check; creator commits the reused tenant ID, aliasing old immutable evidence to the new authority. FK-less secret/route evidence needs the same lifetime fence.

Repair requires one canonical tenant-ID lifetime lock/generation contract acquired by creation, deletion, and every retained-state insert, plus a two-connection deletion -> late retained writer -> reuse regression. Same-ID creators themselves are correctly serialized, raw keys do not escape before commit, and the mounted visibility/rollback proof is otherwise sound, but it does not exercise retained writers. Keep draft/unmerged.

lalalune:
Independent exact-head audit: BLOCK at 52d64d1ddcc220015afc1fbb988b86673f35c024 (tree fdd55ce2ad80d080206e5f32873be732c9c558b5). P1 retained-state TOCTOU: the creation audit advisory lock serializes audit writers only, while retained writers such as proxy_audit_log insert independently. A late old-tenant completion can land after tenantIdHasRetainedState reports empty but before new-tenant commit, aliasing old evidence to reused authority. Require a canonical tenant-ID lifetime/generation fence held by create/delete/every retained writer plus a two-connection late-writer regression. P1 production topology: both creation routes use ordinary getDb/tenant audited transactions without establishing the required tenant RLS context or supported platform authority path; the mounted PG test uses ambient admin DATABASE_URL. The retained inventory also omits audit_chain_heads, archives, and checkpoints. Admin-topology rollback, post-commit cache/key exposure, duplicate creator, and child-FK proofs are otherwise sound. #712 linkage is correct. Keep draft.

PR #827, opened 2026-08-22
fix(api): make policy template mutations audit-atomic

### LINKED ISSUE #717: security: make policy-template and assignment mutations audit-atomic
Exact develop still commits policy-template update/delete and template-to-agent assignment before required completion audit, then calls `restoreTemplateSnapshot` or `restoreAgentPolicies` after audit failure. Once the first commit releases locks, those stale restores can overwrite concurrent template revisions or policy assignments and contradict a successful winner audit. Move each template/assignment mutation and completion audit into one tenant-audited transaction with locked revision/CAS checks; remove snapshot restore; invalidate policy caches only after commit. Add mounted behavior and deterministic real-PG update/delete/assign races with injected audit failure, proving no lost update, stale resurrection, or unaudited authority. Replace source-order/id validation scans with boundary tests where applicable (#709). Local-only until #657 is durable.

lalalune:
Closes #717

## Summary

- make policy-template create, update, delete, and assignment mutations commit in the same tenant-audited transaction as their required completion audit
- use locked revision/CAS checks for template changes and canonical, sorted agent-authority locks plus policy-set CAS for assignments
- remove stale snapshot/restore compensation and invalidate affected policy caches only after a successful audited commit
- normalize raw database result shapes across PostgreSQL and PGLite
- add mounted PGLite rollback coverage and deterministic PostgreSQL races for create, update, delete, and assignment

## Exact lineage

- head: `5745e11a5638bf12ff3881f8a9c91c67f0f14eef`

## Local validation

- focused isolated API files: 7/7 passed
- API dependency build: 24/24 passed
- Biome: 6 changed files clean
- committed diff check: clean

## Required hosted gate

The four deterministic real-PostgreSQL cases were discovered but skipped locally because no real `DATABASE_URL` was available. Hosted PostgreSQL must execute all four modes successfully before this draft can be accepted. The tests use an actual audit trigger/advisory-lock barrier, an exactly named competing connection, and independent visibility checks to prove the failed mutation never becomes visible and the audited concurrent winner survives.

### DIFF
--- a/packages/api/src/__tests__/audit-policy-hardening.test.ts
+++ b/packages/api/src/__tests__/audit-policy-hardening.test.ts
@@ -105,9 +105,9 @@ describe("audit and policy route hardening", () => {
       policiesSource.indexOf('policiesStandaloneRoutes.post("/:id/assign"'),
     );
     expect(assignRoute.indexOf('action: "policy.template.assign.authorized"')).toBeLessThan(
-      assignRoute.indexOf("await db.transaction"),
+      assignRoute.indexOf("withTenantAuditedTransaction("),
     );
-    expect(assignRoute.indexOf("await db.transaction")).toBeLessThan(
+    expect(assignRoute.indexOf("withTenantAuditedTransaction(")).toBeLessThan(
       assignRoute.indexOf('action: "policy.template.assign"'),
     );
   });
--- a/packages/api/src/__tests__/fixtures/policy-template-concurrent-writer.ts
+++ b/packages/api/src/__tests__/fixtures/policy-template-concurrent-writer.ts
@@ -1,0 +1,75 @@
+import {
+  agents,
+  createDb,
+  policies,
+  policyTemplates,
+  withTenantAuditedTransactionOnDb,
+} from "@stwd/db";
+import { and, eq, sql } from "drizzle-orm";
+
+const databaseUrl = process.env.DATABASE_URL;
+const tenantId = process.env.TEST_TENANT_ID;
+const templateId = process.env.TEST_TEMPLATE_ID;
+const agentId = process.env.TEST_AGENT_ID;
+const requestId = process.env.TEST_REQUEST_ID;
+const mode = process.env.TEST_WRITER_MODE as "create" | "update" | "delete" | "assign" | undefined;
+if (!databaseUrl || !tenantId || !templateId || !agentId || !requestId || !mode) {
+  throw new Error("concurrent policy-template writer environment is incomplete");
+}
+
+const connection = createDb(databaseUrl);
+try {
+  await withTenantAuditedTransactionOnDb(
+    connection.db,
+    tenantId,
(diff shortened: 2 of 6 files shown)


lalalune:
Exact source audit at 5745e11: BLOCK; keep draft.

The new assignment path commits tightened policy authority and then treats Redis invalidation as best-effort. `invalidateAgentPolicyCaches` catches the error and returns success. If Redis is available but the delete fails, the next enforcement read can still consume the old permissive cached policy; it does **not** necessarily fall back to the database. A lower limit/delete must not return 200 until enforcement is guaranteed to observe the committed authority. Use a generation/fence in the cache key, a durable invalidation outbox/read-through version check, or fail closed on invalidation for tightening changes, with a hostile Redis-failure mounted proof.

There is also an inconsistent PGLite gate in the same new route: template quota locking skips advisory SQL for either `STEWARD_DB_MODE=pglite` or `STEWARD_PGLITE_MEMORY=true`, but multi-agent assignment skips it only for `STEWARD_PGLITE_MEMORY=true`. A configured `STEWARD_DB_MODE=pglite` runtime will execute `pg_advisory_xact_lock` on PGLite. Centralize the runtime check and cover both supported modes.

Hosted matrix success will not close the stale-authority/cache fence.

lalalune:
Exact-head correction published at `332749b7c1c853c65ffed766f00179b690bbd7e8`.

The assignment authority fence now uses the canonical PGLite predicate (`STEWARD_DB_MODE=pglite` or `STEWARD_PGLITE_MEMORY=true`) and therefore never calls PostgreSQL advisory functions in supported explicit/file-backed PGLite mode. A mounted regression performs and persists a policy assignment with DB mode set and memory mode unset.

Local exact-head evidence: focused 4/4 (13 assertions), API build/typecheck green, Biome and diff-check clean. Hosted checks and independent approval remain required.

lalalune:
Independent exact-current audit: BLOCK at 332749b7c1c853c65ffed766f00179b690bbd7e8 (tree 624a284c43c3fbc500310c9586165df76647d4dd). P1 authority TOCTOU: mutations trust request-start tenantRole, but the audited transaction never locks/rechecks users/user_tenants. Concurrent removal/demotion can still commit template or assignment authority; the PG fixture fabricates admin context without any membership row. Revalidate current membership in canonical lock order inside the audited transaction and add the revocation race. P1 cache safety: assignment invalidation occurs only after commit, catches/logs Redis failure, and has no durable retry/outbox/version fence. A tightening assignment can return success while a permissive cached policy survives to TTL. Add hostile cache-failure/race proof and fail-closed versioned authority. PG tests use unrestricted DATABASE_URL, manual Hono vars, and direct writers instead of restricted app-role/RLS/auth topology. Mutation+audit structure, template CAS, sorted agent locks, and post-success invalidation ordering are otherwise improved. Body lineage is stale at 5745e11a. #717 linkage is correct. Keep draft.

PR #828, opened 2026-08-22
fix(api): make platform tenant config audit atomic

### LINKED ISSUE #716: security: make platform tenant-config mutations audit-atomic
PR #639 fixed the tenant-config router, but exact current develop still has the same unsafe snapshot/mutate/final-audit/whole-row-restore pattern in platform routes for email config update/delete, join mode, OIDC providers, and test-account enable/disable (`restorePlatformTenantConfigRow`). A completion-audit failure can overwrite a concurrent successful config mutation after locks release; cache invalidation can expose the uncommitted/rolled-back authority. Apply the #639 transaction pattern to every platform mutation: authorization audit first, locked re-read + mutation + required completion audit in one tenant/platform-authority transaction, cache invalidation post-commit, no restore helper. Add mounted and two-connection real-PG audit-failure races across same-column and unrelated-column updates. Keep exact platform role/RLS authority boundaries. Local-only until #657 is durable.

lalalune:
Closes #716

## Summary

- commit platform email, join-mode, OIDC, and test-account config mutations with their required completion audit in one tenant-scoped transaction
- lock and re-read the authoritative tenant config before mutation; remove whole-row snapshot compensation
- invalidate email authority caches only after the mutation and completion audit commit
- preserve the application tenant/RLS authority boundary instead of routing tenant config writes through the restricted platform database role
- add deterministic mounted real-Postgres audit-failure races for same-column and unrelated-column concurrent writes

## Exact candidate

- head: `1da4c41a8a4f9d38cc4b274f675247fe09840a26`
- tree: `ec771ce48e0c237d67938a9daa2dc95a46a595ce`

## Validation

- isolated affected PGLite/source fixtures: 41/41
- API dependency build: 24/24
- API TypeScript: pass
- exact Biome 2.5 full configured scope: 1,375 files pass
- public-package metadata and diff check: pass
- formatter-only repair independently audited: PASS; formatted parent blob is byte-identical to the new head

The two mounted real-Postgres race cases compile and are discovered locally, but were skipped because no DATABASE_URL was available. Hosted real-Postgres execution on this exact head remains a required acceptance gate.

This PR is intentionally draft. Do not ready or merge before exact hosted checks and accepted review complete.


lalalune:
Independent exact-head audit of `f8db41e5` is **BLOCK (P1)** on cache/commit ordering in the real mounted RLS path.

Platform tenant middleware owns the outer tenant transaction. `withTenantAuditedTransactionOnDb` detects that context and runs this unit only in a savepoint, returning before the middleware-owned outer transaction commits. Email PATCH and DELETE then call `invalidateEmailAuthForTenant` immediately after that return while claiming the mutation and audit are durable.

Concrete race: request A releases its savepoint and deletes the cache while its config/audit are still uncommitted; request B sees the cache miss, reads the old committed email config, and caches the old EmailAuth; A then commits the outer transaction, with no true post-commit invalidation. Old provider/key authority can remain cached indefinitely.

Repair needs an outer-commit callback or generation fence (or safe pre-invalidation plus true post-commit invalidation), with a mounted PostgreSQL cache race proving no stale authority survives commit. Existing PG races cover rollback against same/unrelated existing-row writers but seed `tenant_configs` first and do not exercise absent-row upserts or cache visibility.

The database mutation side otherwise looks coherent: tenant advisory locking serializes absent-row upserts and tenant deletion, and partial conflict updates preserve unrelated columns. Keep draft/unmerged.

lalalune:
Independent exact-head audit: BLOCK at f8db41e5a3c5ec777cf65ae3d02afeed19c18fbc (tree b30ddaa619635c9fdfc24a026fbdf3cd4450d0d2). P1 cache timing is false under production topology: platform tenant middleware wraps the request in an outer tenant transaction; withTenantAuditedTransaction detects it and commits only a nested savepoint, then email PATCH/DELETE invalidate cache before the outer commit. A concurrent request can reload old DB authority in that window. The cache has no TTL and invalidation is process-local, so old configuration can survive the later commit indefinitely and other replicas are never invalidated. Require true after-outer-commit invalidation or a durable versioned cache fence, with cache/outer-commit failure races. PG tests reuse privileged DATABASE_URL for setup, route, and writer and do not prove restricted app-role/RLS. Row mutation plus completion audit otherwise composes transactionally. Exact hosted Lint+Typecheck+PR Metadata is already failing, with 35 checks nonterminal and no review. #716 linkage is correct. Keep draft.

lalalune:
Closed as a rejected stale carrier. Cache invalidation can still occur before the outer transaction commits and lacks an authoritative bounded lifetime.

PR #831, opened 2026-08-22
fix(auth): preserve JSON fallback for non-navigation OAuth callbacks

### LINKED ISSUE #695: fix(auth): restore HTML callback fallback for navigation without Accept

lalalune:
Closes #695

## Summary
- keep no-Accept OAuth callbacks on JSON unless the request is an actual browser navigation
- preserve explicit content-negotiation quality and q=0 behavior
- add mounted browser and JSON OAuth success proofs without exposing provider tokens
- use canonical JWT and audit-HMAC fixture authority

## Exact lineage
- head: `f79f9e26bc88b06bc4190f08f7f891bd62e048ef`
- tree: `ff6ad21755b7406b866dda8f1c3f0cea583ca0d8`
- stable patch-id: `7739a4bce08ecefb48e7169f171f9e5df9742a08` (identical to prior independently accepted `a5da7ec`)

## Validation
- OAuth callback + token-encryption focused tests: 21/21
- API dependency build: 24/24
- API TypeScript build: green
- Biome (3 files): green
- `git diff --check`: green

Independent source audit: ACCEPT. Keep draft until exact hosted checks and required review are complete.

### DIFF
--- a/packages/api/src/__tests__/auth-oauth-callback-browser.test.ts
+++ b/packages/api/src/__tests__/auth-oauth-callback-browser.test.ts
@@ -1,6 +1,7 @@
 import { afterAll, afterEach, beforeAll, describe, expect, it, mock } from "bun:test";
-import { closeDb } from "@stwd/db";
+import { accounts, closeDb, getDb, users } from "@stwd/db";
 import { createPGLiteDb, setPGLiteOverride } from "@stwd/db/pglite";
+import { eq } from "drizzle-orm";
 
 const REDIRECT_URI = "https://app.example.test/callback?from=oauth";
 const ORIGINAL_FETCH = globalThis.fetch;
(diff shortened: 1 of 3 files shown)


lalalune:
Independent exact-current audit: semantic ACCEPT / delivery HOLD at f79f9e26bc88b06bc4190f08f7f891bd62e048ef (tree ff6ad21755b7406b866dda8f1c3f0cea583ca0d8). Empty/whitespace Accept selects HTML only for a navigation-mode request; otherwise JSON remains the fail-safe response. Explicit quality/specificity and q=0 behavior remain intact. Failure HTML escapes content, uses only the validated recovery URL, and sets no-store/CSP/no-referrer/nosniff. Navigation-shaped and explicit-JSON provider flows return only opaque exchange code/state through the allowlisted fragment; provider/session tokens are absent from the response and provider credentials remain encrypted at rest. The tests are mounted route/navigation-shaped requests rather than literal Chromium, but they proportionately satisfy #695. Keep draft until exact hosted checks and required review are terminal; do not merge from the current queued matrix.

PR #832, opened 2026-08-22
fix(auth): harden migrated store recovery

NubsCarson:
## Summary

- parameterize the PostgreSQL relation inventory used by the RLS deployment safety gate
- let the application role reuse an already-migrated auth store without attempting owner-only DDL
- retain fail-closed behavior when the auth schema is absent or incomplete

This publishes the two recovery commits already proven against the isolated staging recovery. It does not deploy Steward or change any environment/provider configuration.

@lalalune — Auth/Steward recovery draft for review.

## Type

- [ ] `feat` — New feature
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, dependencies

## Scope

`packages/auth`, `packages/db`

## Evidence

Exact head before publication: `e5f84b782972568027460bec042d1b25a1df3265`

- `bun test packages/auth/src/__tests__/store-backends.test.ts packages/db/src/__tests__/rls-deployment-safety.test.ts` — 23 passed
- `bun run --cwd packages/auth build` — passed
- `bun run --cwd packages/db build` — passed
- focused Biome check on all four changed files — passed
- `git diff --check` — passed
- branch is 2 ahead / 0 behind current Steward `origin/develop`

The broader current Steward Apple/OAuth/phone/SMS auth suite was also validated during recovery (82 passing tests); this PR's exact-head publication gate is the focused 23-test set above.

## Rollback

No deployment is included. Revert `e5f84b78` and `482242f1` in reverse order, or close this draft without merging. Existing local annotated checkpoint tags remain available:

- `auth-gateway-staging-recovery-20260821-checkpoint-1`
- `auth-gateway-staging-recovery-20260821-checkpoint-2`

## Breaking Changes

None

## Checklist

- [x] Documentation updated if needed (not applicable: implementation safety fix)
- [x] Local focused tests and package type builds pass
- [x] PR targets `develop` (not `main`)
- [x] Commit messages follow conventional format

### DIFF
--- a/packages/auth/src/__tests__/store-backends.test.ts
+++ b/packages/auth/src/__tests__/store-backends.test.ts
@@ -4,6 +4,7 @@ import {
   buildBackend,
   MemoryBackend,
   NamespacedStoreBackend,
+  PostgresBackend,
   RedisBackend,
   type RedisLike,
   type StoreBackend,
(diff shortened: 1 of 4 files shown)


lalalune:
Independent exact-head source audit of `e5f84b782` is **ACCEPT**.

The migrated-table probe is parameterized and skips all application-role DDL once `public.auth_kv_store` is committed. The missing-table compatibility path retains the transaction-scoped advisory lock plus idempotent table/index creation, so concurrent initializers stay serialized. Missing/wrong columns, conflict target, types, or ACLs still make the startup probe fail, and production readiness rejects the resulting memory fallback; schema drift remains fail closed.

The RLS query now compiles the static manifest through `sql.join` as `IN ($1, $2, ...)`, eliminating the invalid `ANY(($1,...))` PostgreSQL shape without raw interpolation. Exact delta is four files, diff-check clean. This is source acceptance only: keep draft until the exact hosted PostgreSQL/full matrix and required review are terminal green.

lalalune:
Independent exact-head audit: BLOCK at e5f84b782972568027460bec042d1b25a1df3265 (tree 244a4fd34b63e981108ff5d187cb5a4568d7ac6f). Permission-sensitive recovery is tested only through a hand-built fake SQL tag; it never proves that the restricted application role can resolve auth_kv_store and execute the real startup probe, or that missing/incomplete schema fails closed. ensureTable accepts any to_regclass hit without validating relkind, owner, required columns/PK, or expiry index, so an incomplete table can be selected despite the body claim. Startup probes only set/delete; consume, transition, and serializable publish remain unproven under the actual app credential/driver. There is no real-PG restricted-role Redis-outage fallback, restart, or two-process replay proof. The RLS parameterized-IN correction is sound but likewise lacks exact startup role execution here. The PR has no closing issue reference, and its rollback wording is stale relative to the actual parent. Keep draft; add correct linkage and restricted-role migrated/missing/incomplete plus fallback lifecycle evidence before acceptance.

PR #834, opened 2026-08-22
fix(webhooks): make persistent delivery retries crash-durable

### LINKED ISSUE #752: test(webhooks): replace persistent-queue source scans with claim and retry races
## Problem

`packages/webhooks/src/__tests__/queue.test.ts` behavior-tests the in-memory dispatcher, but its persistent delivery authority is still asserted by reading `persistent-queue.ts` strings: `FOR UPDATE SKIP LOCKED`, processing status, dispatcher retry settings, configuration identity, and snapshot ordering.

That does not prove two workers cannot double-send, a crashed claim becomes visible, retries retain original delivery identity/time/config, or configuration changes are applied safely.

## Acceptance

- Use real PostgreSQL and controlled HTTP/adaptor barriers to run two queue workers against one due delivery; prove exactly one claim/send per attempt.
- Cover visibility-timeout recovery after a crashed worker and no early retry before `nextAttemptAt`.
- Prove one durable delivery ID across retries while signature timestamps advance and attempts remain bounded.
- Prove URL/secret/event snapshot behavior under config disable/change/delete without cross-config delivery or secret substitution.
- Exercise final failure/success persistence and dispatcher no-internal-retry behavior.
- Remove the corresponding source scans; run webhook build/tests, real-PG suite, Biome, and diff check.

lalalune:
Closes #752

## Summary
- replace persistent-queue source assertions with mounted real-PostgreSQL claim, retry, visibility, and configuration-snapshot behavior
- persist delivery identity before work becomes visible and atomically repair legacy identity/timestamp metadata during claim
- consume attempts before external I/O, terminalize exhausted crash recoveries without another send, and preserve stable receiver dedup identity with fresh per-attempt signatures

## Evidence
- Parent candidate: full webhooks suite 57/57 passed
- Parent candidate: new real-PostgreSQL persistent-queue suite 5/5 passed
- Final crash-durability corrective: strict independent static audit PASS; `git diff --check` passed
- The final corrective crash tests have not yet been runtime-executed because the prior local environment was disk-constrained

## Pending gates
- Hosted exact-head CI
- Independent exact-head review


lalalune:
Exact `fd3ee4f61eb44fa5f17395cbc8e331afb1af9337` remains BLOCKED. The claim has no owner/fencing token: it only writes `status=processing`, increments `attempts`, and pushes `next_retry_at`, while every post-dispatch transition updates by delivery `id` alone. If a legitimate dispatch exceeds the 5-minute visibility timeout, worker B can reclaim and send while worker A is still live; A and B can then overwrite each other's delivered/failed/dead result. The new tests model a vanished worker, but never a live slow worker crossing lease expiry, so they do not prove stale-claim exclusion. Add a per-claim token/generation (or CAS on the exact claimed attempt/lease) to every result transition and a deterministic two-worker real-PG test where A remains live past expiry, B reclaims, and A's late completion cannot mutate B's authority. Keep draft; no merge on queued checks.

lalalune:
Exact-head audit BLOCK at `fd3ee4f61eb44fa5f17395cbc8e331afb1af9337` (tree `0cbfc4d5b929409267aeb448349177d74b8a8adc`). All post-I/O outcome writes predicate only on job ID, not claim owner/generation/attempt. A first worker can outlive its lease, a second worker can take over and finish, then the stale first claimant can overwrite the newer delivered/dead state. Crash tests use promises that never return, so they miss this stale-claim completion. Webhook configuration is also read without a lock/version and can be disabled/updated/deleted before submit while the old payload still sends. The two worker proof shares one database handle and does not assert distinct backends or restricted-role/RLS topology. Add durable claim-token CAS on every outcome, a returning-stale-claim race, config fencing through submit, and true two-connection restricted-role PG evidence. Keep draft.

lalalune:
Exact-head two-worker blocker corrected in `31287be15c2833aa4d1a6f7eec604162497aa5e9`.

The visibility timeout previously allowed worker B to reclaim a slow worker A’s row, but every post-dispatch write filtered only by delivery ID. A could therefore return late and overwrite B’s newer delivered/retry/dead outcome. All post-claim writes are now fenced by `(id, status=processing, attempts=claimed generation)`, so stale workers cannot mutate a reclaimed row. A real-Postgres adversarial regression forces A to return a late failure after B has delivered and requires the durable row to remain delivered at attempt 2.

This also repaired the exact-head poison-pill fixture to model the atomic claim’s already-incremented attempt/status; the prior final head’s focused runtime test failed its own attempts expectation.

Local exact-head evidence: full webhooks suite 52 passed/8 real-PG-gated skips (279 assertions), package typecheck, Biome, and diff-check clean. Hosted Postgres will execute the new stale-worker race. Keeping draft pending terminal green checks and independent approval.

lalalune:
New-head exact audit remains BLOCK at `31287be15c2833aa4d1a6f7eec604162497aa5e9` (tree `f3eb422b5dfcdc5fbb5b0b5b4fdb1649451be0d1`). Outcome writes now include status/attempt fencing, but ignore whether the CAS matched and unconditionally return the stale worker result; the new regression explicitly expects stale attempt-one failure after attempt two delivered, codifying a false authoritative outcome. Configuration also has no epoch/fence through external submission, so disable/delete/URL/event changes after SELECT still permit stale delivery. Real-PG uses global DB access without restricted tenant roles/GUCs or distinct backend proof. Check affected rows on every terminal CAS and suppress lost-claim results; add config-generation submit fencing and restricted two-connection races. PR is dirty, all checks queued, no review; keep draft.

PR #835, opened 2026-08-22
test(auth): execute SAML trust pins and replay fences

### LINKED ISSUE #759: test(auth): execute SAML trust pins instead of scanning adapter options
## Problem

`packages/auth/src/__tests__/saml.test.ts` has useful signed-response tests, but it proves several trust pins by reading `saml.ts` strings: signed assertion/response requirements, audience, ACS, IdP issuer, request replay cache, SHA-256 algorithms, and the absence of raw XML parsing.

Current behavior covers happy path, request-ID mismatch, tampering, and configured email attributes, but not each trust-pin mismatch or replay. The source checks can pass if an option is ignored/misbound or a validated profile is confused with untrusted XML.

## Acceptance

- Generate signed fixtures that independently vary response/assertion signing, audience, destination/recipient ACS, IdP issuer, request ID, signature/digest algorithms, and assertion ID.
- Prove each invalid variant rejects while the exact valid fixture succeeds.
- Replay the same valid assertion/request and prove the second use fails through the production cache/replay boundary.
- Include hostile unsigned/raw attributes that disagree with the validated signed profile and prove only validated profile data is returned.
- Remove or narrow the option/raw-parser source assertions after behavioral parity.
- Run full auth tests/build, Biome, and diff check.

lalalune:
Closes #759

## Summary
- replace SAML option/source scans with signed-fixture behavior across response/assertion signatures, audience, ACS destination/recipient, issuer, request IDs, algorithms, and assertion IDs
- prove hostile unsigned/raw attributes cannot override the validated signed profile
- exercise production request/assertion replay fencing through the mounted ACS flow

## Evidence
- Focused SAML and mounted ACS tests: 17/17 passed
- Local exact-head validation completed on `d4db4cde870cfd378cebb3e3a7433b9f3bbc1a6d`

## Pending gates
- Hosted exact-head CI
- Independent exact-head review


lalalune:
Independent exact-head audit: BLOCK at d4db4cde870cfd378cebb3e3a7433b9f3bbc1a6d (tree 1d4f691258be6280cc41fa11d6eec54a65698c29). P1 fixture/production mismatch: replay assertionId is derived from profile.ID, but node-saml does not populate that from Assertion @ID; the fixture manufactures an AttributeStatement attribute named ID equal to the XML ID and masks the failure. A normal signed assertion without that custom attribute is rejected, and replay is not bound to the cryptographically validated Assertion @ID. Extract/validate the signed element ID and add absent/mismatched hostile attribute cases. P1 request-ID mismatch: the cache returns the request ID itself, while node-saml parses the SubjectConfirmationData InResponseTo cache value as request creation time; a standards-normal assertion yields Invalid Date. Fixtures omit SubjectConfirmationData InResponseTo and exercise only Response InResponseTo. Return an issuance timestamp tied to the DB request and test matching/wrong subject IDs. Coverage is also direct authRoutes/PGLite under NODE_ENV=test, not composed restricted real-PG/RLS topology. #759 linkage is correct. Keep draft.

PR #836, opened 2026-08-22
fix(trading): make collateral transfers durable and replay-safe

### LINKED ISSUE #750: test(trading): replace collateral-transfer route source scans with mounted behavior
## Problem

`packages/plugin-trading/src/__tests__/hip3-collateral-transfer.test.ts` is entirely a source-string/order test. It never mounts `/:venue/transfer`, authenticates a request, calls the adapter, or observes audits.

The suite can pass while route mounting, auth discrimination, transfer argument binding, failure classification, or audit persistence is broken.

## Acceptance

- Mount the assembled trading API and exercise the collateral-transfer route.
- Prove platform authority succeeds while tenant/admin/agent auth fails before adapter calls.
- Assert exact `signSendAsset`/`submitSendAsset` inputs for builder-dex exit and core withdraw cases.
- Inject signing, definite submission rejection, ambiguous transport loss, and completion-audit failure; assert stable HTTP/replay behavior and exact requested/failed/submitted audit state.
- Add real-Postgres coverage for mutation/audit rollback or durable ambiguous evidence where applicable.
- Remove the source scan; run plugin/API builds, focused suites, Biome, and diff check.

lalalune:
Closes #750

## Summary

- replace the collateral-transfer source scan with mounted production-route behavior
- gate operator fund movements to platform authority and bind exact adapter inputs
- persist tenant-scoped, chained replay evidence for requested, ambiguous, failed, and submitted outcomes
- serialize fresh-worker claims before venue execution so identical retries submit exactly once and conflicting fingerprints fail closed
- reject tampered or deleted replay evidence instead of resubmitting
- keep required audit-chain writes authoritative while denied proxy best-effort audit remains isolated

## Exact candidate

- head: `b8a5974cd80084994fa5fe0aae6bd1cc8c15a8aa`
- tree: `7ea6fa35450548953e3e2af2e7d7b88f4d60259d`

## Local evidence

- independent read-only exact-head semantic audit: PASS
- mounted route coverage exercises platform success, unauthorized denial, exact transfer arguments, signing failure, definite rejection, ambiguous loss, and completion-audit failure
- mounted PostgreSQL coverage includes cold replay, pending evidence after audit outage, restricted cross-tenant denial, identical fresh-worker serialization, and conflicting-fingerprint serialization
- changed diff passes `git diff --check`

Exact hosted PostgreSQL execution, the complete protected matrix, and independent PR review remain required. This PR is intentionally draft; do not ready or merge until those gates are terminal green.


lalalune:
Exact `b8a5974cd80084994fa5fe0aae6bd1cc8c15a8aa` remains BLOCKED.

P1 — the durable replay authority expires under supported retention. Redis/process idempotency expires after 24h, while the normal audit retention sweep eventually archives/deletes old `audit_events` and advances `floor_seq`. Once the keyed requested/terminal rows fall below that floor, `findDurableTransferState` finds no matching live row and a valid post-floor chain is interpreted as authoritative absence; the route can append a new requested marker and submit the same irreversible idempotency key again. A replay-safe capital movement needs a non-expiring dedicated replay/tombstone row retained independently of audit-event retention, or an authoritative archive lookup/retention pin.

P2 — the issue's assembled production-auth proof is still missing. Both mounted transfer suites inject local `operatorAuth`/agent stubs rather than mounting the production middleware/app, and the PG race does not assert distinct backend PIDs. Add a production-context platform-vs-tenant/admin/agent auth test and explicit independent-connection PID proof.

Keep draft; queued checks cannot close the retention hole.

lalalune:
Additional exact blocker: Redis currently short-circuits before the replay guard and HMAC-chain verification. A cached completed response is returned immediately, so deleted/tampered durable audit evidence is not examined during the 24h cache window, contrary to the PR's “reject tampered or deleted replay evidence” claim. The tamper tests hide this by supplying `getRedisClient: () => null` and a fresh local store. Redis may optimize lookup only after authoritative evidence verification (or must itself carry a cryptographically verified binding); it cannot outrank the claimed durable ledger.

lalalune:
Exact-head audit BLOCK at `b8a5974cd80084994fa5fe0aae6bd1cc8c15a8aa` (tree `7ea6fa35450548953e3e2af2e7d7b88f4d60259d`). `operatorAuth` accepts any valid platform key but never resolves scopes, while the collateral transfer route checks only platform auth type; an unscoped/read-only key can move collateral. Replay durability also expires: Redis lasts 24h and lookup searches only live audit rows, so legitimate retention/archive can remove evidence and allow the same key to submit again. Finally, submitted-audit failure is swallowed and returns 200 while cold replay becomes 409 rather than stable success. Fingerprint/advisory serialization and two restricted PG handles are useful, but these P1s block. Keep draft.

PR #837, opened 2026-08-22
fix(trading): make session creation audit-atomic

### LINKED ISSUE #753: fix(trading): make session creation and completion audit atomic
## Problem

`POST /v1/trade/sessions` creates the DB/Redis session, writes `trade.session.created` afterward, then compensates an audit failure with `sessionManager.deleteSession`. The only regression is a source-order scan in `trade-policy-audit.test.ts`.

This is the same stale-compensation pattern being removed elsewhere: a concurrent reader/mutator can observe or change the session before the failure cleanup, Redis/DB deletion can partially fail, and cleanup can erase a concurrent legitimate state. The authorization audit may remain before mutation, but the session row and required completion audit need one locked DB transaction; cache/session publication must be post-commit or durably recoverable.

The same file also source-scans agent-token auth exclusion and idempotent replay envelopes instead of executing them.

## Acceptance

- Create the session row and required completion audit in one tenant transaction with the required agent/session lock ordering and locked policy revalidation.
- Publish Redis/cache state only after commit, with bounded durable recovery or fail-closed behavior if publication fails; remove delete compensation.
- Add mounted and deterministic real-Postgres audit-failure/concurrent create tests proving no transient committed row, no lost winner, and no completion audit on rollback.
- Behavior-test agent-token exclusion and exact replay envelopes/status/headers for success, validation failure, and ambiguous venue failure; remove the source scans.
- Run trade-session/plugin/API builds and focused suites, real-PG proof, Biome, and diff check.

lalalune:
Closes #753

## Summary

- create the durable trade-session row and required `trade.session.created` audit in one tenant-scoped transaction
- lock and revalidate agent, policy, and venue-wallet authority before mutation
- remove post-commit delete compensation and keep the database row authoritative with bounded read-through recovery instead of premature cache publication
- replace session-creation source assertions with mounted authorization and rollback behavior
- add deterministic PostgreSQL audit-failure, policy-race, wallet-race, and concurrent-create proofs

## Exact candidate

- head: `874648460b8a5da7628fe4ce15889585d95be651`
- tree: `313717ee02b05977c7ec1896e1b507cc921c4c14`

## Local evidence

- independent local exact-delta audit: PASS
- mounted trade-policy suite covers tenant-level auth and required-audit rollback
- mounted PostgreSQL suite defines four deterministic cases: completion-audit rollback, concurrent policy change, concurrent venue-wallet rotation, and two concurrent successful creates with matching audits
- changed diff passes `git diff --check`

Exact hosted PostgreSQL execution, package/API builds, the complete protected matrix, and independent PR review remain required. This PR is intentionally draft; do not ready or merge until those gates are terminal green.


lalalune:
Exact `874648460b8a5da7628fe4ce15889585d95be651` remains BLOCKED on the required concurrency invariant. Session creation takes `pg_advisory_xact_lock("agent-policy:...")`, but the production agent-policy mutation in `packages/api/src/routes/agents.ts` does not take that lock or an equivalent agent lock; it performs the `agentPolicies` upsert directly. Therefore the claimed no-row serialization is ineffective: creation can lock the absent-policy key, observe no row/default policy, while a concurrent first policy insert (which does not share the advisory lock and has no row to block on) commits a tightening; the session can still commit from the stale default snapshot. The real-PG test updates an existing row before the session enters its transaction, so it cannot expose this missing-row interleaving. Make the policy writer and session creator share one canonical lock/order, and add a deterministic two-connection absent-row insert race proving tightening wins or creation revalidates. Audit the analogous absent venue-wallet insert path as well. Keep draft.

lalalune:
Exact-head audit BLOCK on required production-topology evidence at `874648460b8a5da7628fe4ce15889585d95be651` (tree `313717ee02b05977c7ec1896e1b507cc921c4c14`). The source transaction appears materially sound: session insertion and required completion audit are composed, authority rows are locked/revalidated, and Redis is not published before outer commit. But the claimed real-PG proof mounts synthetic auth variables and uses broad `createDb(DATABASE_URL)` writers; it never exercises tenant auth, the request-owned outer transaction/savepoint, restricted-app RLS, or canonical route-vs-route revocation. Add a restricted-app real-PG mount through actual tenant middleware with two canonical route connections and outer-commit rollback/visibility plus create-vs-policy/wallet/revoke/delete races. Keep draft; 60 checks queued and zero reviews.

lalalune:
BLOCKER on exact `874648460b8a5da7628fe4ce15889585d95be651`: final wallet authority is not serialized for no-row/deletion races. The audited transaction uses `SELECT ... FOR UPDATE` on `agent_wallets`; absence locks nothing, so a concurrent venue-wallet create/import can commit after the no-row read while session creation commits against caller-supplied fallback authority. If the preflight canonical row is deleted and the supplied address matches, the fallback/provenance branch can likewise create a session with no current venue wallet. Existing PostgreSQL coverage races only UPDATE of an existing row.

Repair must acquire the shared custody-transition authority lock before the final re-read, preserve canonical-row versus fallback provenance, and add two-connection absence→insert and delete races. This PR also predates the merge wave and must be restacked on current `develop`, preserving merged #836. Repair is in progress; keep draft.

PR #838, opened 2026-08-22
test(web): exercise invitation acceptance confirmation in browser

### LINKED ISSUE #755: test(web): mount invitation acceptance confirmation instead of scanning source
## Problem

`web/src/app/accept-invitation/accept-invitation.test.ts` protects the SEC-075 confirmation boundary only by reading `page.tsx` and matching `useEffect`, handler, button, and call-count strings. There is no mounted or browser interaction test.

The scan can pass while navigation/load triggers a POST, Accept submits twice, Decline submits, a stale token/tenant is used after navigation, or failure/success state is misleading.

## Acceptance

- Mount the page or exercise it in Playwright with controlled auth/session, query parameters, and API requests.
- Prove initial navigation and rerender never POST; only explicit Accept sends exactly one request with the current tenant/token/session.
- Prove Decline and browser back/close paths send nothing.
- Cover double click/pending disable, success, sanitized failure, missing/invalid parameters, auth rotation, and route navigation with a delayed prior request.
- Remove the source scan; run web typecheck/build, focused browser or mounted tests, Biome, and diff check.

lalalune:
Closes #755

## Summary
- replace the invitation page source-string scan with mounted Playwright coverage
- require explicit, single acceptance using current validated tenant, claim, and concrete session credentials
- fence duplicate clicks and stale navigation completions; cover decline/back, missing and invalid inputs, auth rotation, sanitized server/network/parse failures, and strict success response validation

## Isolated scope
- web/e2e/accept-invitation.spec.ts
- web/src/app/accept-invitation/page.tsx
- delete web/src/app/accept-invitation/accept-invitation.test.ts

No middleware, provider-trust, or other multi-issue carrier bytes are included. The two surviving file blobs exactly match corrective source c0c71b377614f830790d6187da4d034411873e0d.

## Evidence
- exact head: 7428648565efc4fe715d76b357b2d482557d2bc7
- exact tree: 73ffb5efa30ce3a3e35a9000bdac57c2ef5572df
- prior focused Playwright: 6/6 PASS
- prior web build/typecheck: PASS
- prior Biome: PASS
- exact isolated diff check: PASS

## Pending gates
- hosted protected checks on this exact isolated head
- independent review

This remains draft and must not be readied or merged until those gates are terminal green.


lalalune:
Queue audit: the hardened client validator matches the API route’s top-level success contract, explicit-confirmation/race fences are covered by focused Playwright tests, and the patch is ready for independent review. Hosted required checks remain queued.

lalalune:
Independent exact-head audit: BLOCK at 7428648565efc4fe715d76b357b2d482557d2bc7 (tree 73ffb5efa30ce3a3e35a9000bdac57c2ef5572df). P1 production-auth mismatch: the real invitation-accept route requires a personal-user session with tenantId personal-<userId>, but every successful Playwright request uses a non-personal tenant claim and is accepted only by page.route mocks. Those exact credentials receive 403 in production, while the page offers no personal-session rotation/actionable recovery. P1 stale-auth fence: requestGeneration changes only when the invitation URL changes, not when auth user/token changes. A pending request from user A can complete after logout/login to user B on the same URL and set accepted/success in B’s retained document. The auth-rotation test rotates only after completion and also changes the URL, so it misses the race. Add mounted API/browser personal-session success, hostile non-personal rejection/rotation, and identity/credential-epoch invalidation. Other explicit-click, duplicate-click, decline, malformed, URL-navigation, strict-response, and sanitized-error behavior is sound; #755 linkage is correct. Keep draft.

PR #840, opened 2026-08-22
test(web): execute provider auth-token transitions

### LINKED ISSUE #756: test(web): execute provider auth-token and refresh-custody transitions
## Problem

`web/src/components/providers.test.ts` behavior-tests wallet chunk timeout helpers, but the security-critical `AuthTokenSync` and refresh-custody cases only read `providers.tsx`/`api.ts` strings. They do not mount the provider or observe token installation/removal.

The scans can pass while a signed-out or tenant/user-rotated page retains a prior bearer token, `getToken()` races the concrete session token, a stale passive effect wins, legacy refresh material survives, or the SDK is configured without the same-origin proxy.

## Acceptance

- Mount the production provider/auth sync with controlled auth epochs and observable SDK + legacy API clients.
- Prove sign-in installs the current concrete session token before descendant loads; tenant/user/session rotation replaces it exactly once; sign-out clears it before descendant actions.
- Prove a delayed old token/getToken result cannot overwrite a newer epoch.
- Prove the SDK uses `/api/auth` and removes legacy JS-readable refresh material through behavior, including failure paths.
- Retain wallet-runtime helper tests, remove the token/custody source scans, and add browser coverage where effect ordering matters.
- Run web typecheck/build, focused mounted/browser tests, Biome, and diff check.

lalalune:
Closes #756

## Summary
- replace provider/API source scans with mounted production `AuthTokenSync` behavior
- prove concrete session installation and sign-out clearing occur before descendant passive effects and explicit actions
- prove retained-document tenant, user, and session rotation installs each current token once while a superseded epoch cannot win
- exercise same-origin `/api/auth` configuration and legacy refresh-token removal for authenticated and authentication-failure states
- retain the wallet-runtime timeout behavior tests

## Evidence
- Focused mounted + real Chromium provider suites: 9/9 passed
- Web TypeScript check: passed (`tsc --noEmit --incremental false`)
- Biome: passed
- `git diff --check`: passed
- Independent static audit: passed

## Pending gates
- Hosted exact-head CI and independent review
- Local `next build` was intentionally deferred because the disk had only about 311 MiB free; no dependency install or build expansion was attempted

### DIFF
--- a/web/src/components/providers-auth-sync.test.tsx
+++ b/web/src/components/providers-auth-sync.test.tsx
@@ -1,0 +1,499 @@
+import { afterAll, beforeEach, describe, expect, mock, test } from "bun:test";
+import { resolve } from "node:path";
+import { chromium } from "@playwright/test";
+import { Window } from "happy-dom";
+import * as React from "react";
+import { createRoot, type Root } from "react-dom/client";
+
+const window = new Window({ url: "https://steward.test/accept-invitation" });
+Object.assign(globalThis, {
+  window,
+  document: window.document,
+  navigator: window.navigator,
+  HTMLElement: window.HTMLElement,
+  IS_REACT_ACT_ENVIRONMENT: true,
+});
+
+type AuthState = {
+  activeTenantId: string | null;
+  getToken: () => string | null;
+  isAuthenticated: boolean;
+  session: { token: string } | null;
+  user: { id: string } | null;
+};
+
+let authState: AuthState;
+const installedTokens: string[] = [];
+let clearedTokens = 0;
+let currentToken: string | null = null;
+let observedProviderAuth: Record<string, unknown> | undefined;
+
+mock.module("@stwd/react", () => ({
+  StewardProvider: ({
+    auth,
+    children,
+  }: {
+    auth: Record<string, unknown>;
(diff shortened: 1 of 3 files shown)


lalalune:
Independent exact-head audit: BLOCK at 809d4fbdac5c05d2a54f1edba46b30a0a05c0d95 (tree d32cb8f37304cd6da61b83f5c40fc92adc491f95). The new tests mock the entire @stwd/react boundary; mocked StewardProvider only copies auth props and never constructs production StewardAuth. The browser bundle repeats a no-op provider/mock useAuth and mocks the legacy API. This proves AuthTokenSync effect ordering against controlled state, not the required mounted provider/observable SDK state machine. The refresh-custody case merely observes authProxyUrl and preloads unauthenticated state; it does not execute failed session deposit, real SDK sign-in, BFF/cookie behavior, or storeAndReturn/depositRefreshToken, so it cannot catch refresh-token custody regressions. Mount real StewardProvider/StewardAuth with intercepted same-origin auth endpoints and real legacy-client observation, including successful and failed deposit/refresh with no JS refresh token or stale access identity. A-to-B-to-stale-discard-to-signout ordering is otherwise sound, and #756 linkage is correct. Keep draft.

Issue #770, opened 2026-08-20
security(worker): close the request-local runtime environment inventory


lalalune:
## Problem

Worker request-local environment isolation is incomplete. The Worker wraps each request in an immutable runtime snapshot, but still hydrates one mutable global `process.env`; dozens of request-path readers consume that global directly. Overlapping requests in a reused isolate can therefore read another binding generation, and binding-only deploys can leave cached/module-level configuration stale.

At exact `develop` `3559ac2108d118730ff99e0b8a2a9fdc55895706`, security-sensitive direct readers remain across:

- authorization-signature roots/requirement and request-expiry windows;
- Redis driver/credentials, idempotency durability/TTL, and auth-rate-limit fail-open controls;
- trusted-proxy/Cloudflare client-IP authority;
- OAuth unbound-code exchange, passkey RP/origin allowlists, SIWE domains, email/SMS/Telegram providers and credentials;
- scheduler, receipt polling, checkpointing, gas sponsorship, provider lifecycle, and other route/service policy.

For example, `authorization-signature.ts` reads request-signing secrets and required-mode flags from `process.env`, while `auth.ts` reads trusted-proxy, soft-fail, OAuth, passkey, and provider credentials the same way. A suspended request can resume after another request hydrates a different binding set and authorize/deny or call an external provider under the wrong authority.

Prior focused fixes (#668, #671, #674, #764, #768, #769) cover individual surfaces, but there is no closed inventory preventing the next security-sensitive global reader.

## Required fix

- Inventory every production `process.env` reader reachable from Worker fetch/scheduled handlers and classify it as immutable build metadata, Bun-only startup config, or request-scoped runtime authority.
- Migrate every request-scoped reader to `runtimeEnvironmentValue` or a typed immutable authority object constructed inside `withRuntimeEnvironment`.
- Pass authorities explicitly into cached services/routes or key safe caches by a canonical full authority fingerprint.
- Remove the mutable `hydrateProcessEnv` bridge from security decisions; retain it only for audited legacy/non-authoritative compatibility until the inventory reaches zero.
- Add a checked-in static inventory/guard that rejects new Worker-reachable security configuration reads from global `process.env`.
- Default/fail closed whenever the current request snapshot lacks a required production binding.

Coordinate rather than duplicate the bounded custody, adapter, and webhook fixes in #764/#768/#769.

## Acceptance evidence

[...]

PR #845, opened 2026-08-22
fix(worker): bind webhook transport policy per request

### LINKED ISSUE #769: fix(worker): bind webhook transport and SSRF policy to each request

lalalune:
Closes #769.

## What changed
- resolve webhook insecure/private-network authority from the immutable request runtime snapshot
- capture and thread one frozen authority through registration, update, detached/durable/test/replay delivery paths
- make cached dispatchers re-resolve omitted authority per operation while preserving explicit caller options
- keep SAML, gas-provider, and ERC-8004 generic URL validation fail-closed
- add mounted route, hostile overlap, binding-removal, DNS/literal, and cached-dispatch regressions

This is the isolated 18-file webhook delta only; it contains no #770 custody/runtime-inventory changes.

## Exact local evidence
- webhooks package: 61/61
- focused API/Worker/signed-delivery: 52/52
- authority-focused group: 44/44
- API and webhooks TypeScript builds: pass
- Biome, public-package metadata, and diff check: pass

Hosted exact-head checks and independent GitHub review remain required. Keep draft; do not merge until both are terminal green/accepted.


lalalune:
Independent exact-head source audit of `352ecec7915ee7f514e0b0558bc4a4c993d4cf0a` (tree `241921ec`, base `201c1869`) is **ACCEPT**. Request-local transport authority is captured from the ALS-backed runtime environment and explicitly threaded through route validation, durable/detached dispatch, retries, test/replay paths, and Worker fetch/scheduled execution. Cached dispatchers resolve omitted flags per operation while explicit true/false options remain authoritative, closing stale-isolate and overlapping-request leakage. Socket-time validation rechecks literals and every DNS answer; redirects remain disabled. The layered overlap, binding-removal, cached-policy, redirect, and dispatcher tests exercise the boundary without weakening unrelated SAML/gas/ERC defaults. Keep draft until this exact head has terminal hosted checks and required independent review; do not merge from source acceptance alone.

lalalune:
Exact-head source audit at `352ecec7915ee7f514e0b0558bc4a4c993d4cf0a` finds request authority correctly captured and threaded through registration, delivery, replay, and cached retry dispatchers; focused tests are green. HOLD/BLOCK remains because the claimed mounted Worker proof manually wraps a hand-mounted Hono route with `withRuntimeEnvironment` and never executes production `worker.fetch`. Add sequential A→B→missing plus overlap through the exported Worker composition, including cached retry/delivery, so wrapper ordering cannot regress invisibly. The PR was externally promoted with almost all checks queued and no review; I restored it to draft. Keep draft pending production-composition proof, terminal exact checks, and approval.

PR #846, opened 2026-08-22
fix(security): make wallet claims and accounts audit-atomic

### LINKED ISSUE #721: security: make pregenerated wallet claims durable and recoverable
## Gap

Exact `origin/develop` `3559ac210` implements pregenerated user-wallet claim in `packages/api/src/routes/user.ts` as an in-memory multi-step custody transfer:

1. CAS the source agent's `platformId` to a claimed marker;
2. export private keys from the source Vault authority;
3. import Solana then EVM keys into the personal target;
4. update the target agent and apply defaults;
5. delete the source agent;
6. append required completion audit afterward.

If an import/DB step fails, the catch only CAS-restores the source `platformId`. Partially imported target keys/agent/defaults remain. A retry can conflict with or overwrite that partial target while the claim token becomes reusable. If the final audit fails after source deletion, the custody transfer is complete but unaudited and the route fails with no durable recovery/replay record. Existing tests do not inject failures at every custody boundary or concurrent claim.

## Acceptance

- Introduce a durable claim lifecycle with exact token/source/target/user binding, owner token/lease, and terminal/recoverable states before any key export/import.
- Make retries and concurrent claims idempotent/CAS-fenced; never reset a claim token to reusable after target authority may have been written.
- Preserve source authority until the target key set and agent/default metadata are durably adopted, then retire source and append the required completion audit under one authoritative lifecycle boundary.
- Define recovery for partial Solana/EVM import and post-transfer audit failure without exposing or duplicating key material.
- Add mounted Vault/DB fault injection at every boundary plus deterministic concurrent claimant tests; prove exactly one target authority, no orphan/duplicate key rows, no reusable consumed token, and one ordered completion audit.

### LINKED ISSUE #723: security: make digital-asset account mutations audit-atomic
## Gap

Exact `origin/develop` `3559ac210` still separates durable digital-asset account mutations from their required audits in `packages/api/src/routes/accounts.ts`. Confirmed carriers include aggregation create/delete and account update/delete; account create/update may also provision wallet agents before the account mutation. Examples:

- aggregation create inserts then audits; audit failure returns 400 while the row remains;
- aggregation/account delete commits before audit;
- account update applies membership/metadata changes and only then audits;
- cleanup of newly provisioned wallet agents is best-effort compensation outside the account transaction.

Clients can receive failure for committed state, or a crash/audit failure can leave unaudited authority and orphan/delete races. Concurrent updates can invalidate cleanup assumptions.

## Acceptance

- Inventory every write route in `accounts.ts` and bind account/aggregation row mutation plus required completion audit in one tenant-audited transaction with account-scoped serialization.
- Stage external/Vault wallet provisioning with a durable lifecycle; adopt it atomically into the account or retire it safely without deleting a concurrent winner.
- Remove post-commit cleanup as the correctness mechanism.
- Preserve authorization metadata validation, wallet membership constraints, error redaction, and response contracts.
- Add mounted failures and deterministic real-PostgreSQL/Vault races for create/update/delete/aggregation and partial wallet provisioning; prove no committed failure response, unaudited row, stale overwrite, or orphan authority.

lalalune:
Closes #721.
Closes #723.

## Summary
- make pregenerated wallet claims durable, restart-safe, and fenced against concurrent claimers
- make digital-asset account mutations commit with their required audit events atomically
- add the schema and contiguous `0114_durable_wallet_claim_account_audit` migration needed by both invariants
- replace source-order assumptions with mounted lifecycle, rollback, and concurrency behavior

## Exact local evidence
- exact base: `201c1869d40ffca8a207a32a24eecc7eb6f24cd6`
- exact head: `89cb3beca14a1bea40d6417f17431ca9d01207b1`
- focused mounted/DB suites: 26 passed, 251 assertions
- DB and API builds/typechecks: pass
- 12 changed files Biome-clean
- `git diff --check`: clean
- worktree clean; no dependency installation

This remains draft until exact hosted checks and independent review complete. Migration numbering must be restacked if another migration lands first.


lalalune:
Independent exact-head audit of 89cb3beca14a1bea40d6417f17431ca9d01207b1 is BLOCK (P1). Configured-wallet provisioning creates a normal live Vault agent outside the audited account transaction; a crash before membership/finalization leaves a signable orphan with no autonomous recovery path. The catch path still uses ignored best-effort retirement. Account writers lock account IDs while retirement locks wallet agent IDs, allowing hostile cross-account attach/retire races and cascade deletion of a committed membership; claim final source deletion has the same unfenced membership effect. Claim import also performs vault.importKey before the owner-token CAS, so an expired stale worker can execute the external side effect before discovering takeover. Required real-PG mounted outer-transaction, crash/restart, lease-expiry, and two-connection races are absent (coverage is PGLite/manual-context), and migration 0114 conflicts with the reserved #696 slot. Require a fail-closed staged authority, autonomous recovery, shared wallet/account lock order with evidence-locked effects, hostile PG races, and migration restack. Keep draft; do not merge.

lalalune:
Exact-head blocker audit on `89cb3beca14a1bea40d6417f17431ca9d01207b1`: keep draft. The account advisory-lock predicate misses explicit `STEWARD_DB_MODE=pglite`. More importantly, configured-wallet staging/provisioning has no crash-recovery consumer: only the originating request retains `ownerToken`, and failed `retireStagedAccountWallets` cleanup is dropped, so a crash/failure can leave orphaned authority. The new lifecycle tables also need RLS policy/activation composition and real PostgreSQL/Vault race proof. Core migration `0114` must be renumbered when composed behind #843.

lalalune:
All reported blockers are corrected on exact successor `b84b50c5b05c9fa606d788f5e0620f8589ba12c1`.

- Account lifecycle locks use the canonical PGLite predicate.
- Configured-wallet claims now have a fenced durable reconciler: owner-token takeover, membership adoption, marker-bound Vault authority deletion, required audit in the same transaction, persisted recoverable classification after inline retirement failure, and bounded startup plus request-path recovery.
- Both lifecycle tables now have production RLS policies, inventory coverage, generated manifest coverage, and activation-count coverage.
- The real-Postgres adversarial case uses two processes and an actual Vault-created orphan, proving takeover and encrypted-authority cascade.

Final-byte evidence: DB/API typechecks, Biome, and diff-check green; focused PGLite recovery/accounts/RLS suites green (including 36/36 broader battery and dedicated reruns); real-Postgres two-process Vault recovery 1/1; migration replay through `0114` green.

Composition gate: the branch still names its migration `0114`; it must be renumbered against the final accepted sequence before merge. Keeping draft pending hosted exact-head checks, composition, and independent approval.

lalalune:
Exact-head audit BLOCK at live `b84b50c5b05c9fa606d788f5e0620f8589ba12c1` (tree `744d91ce114bd953b0806e51fcee7370c69f728f`). The claim request is bound to the personal tenant transaction, but it selects/CAS-updates/deletes the pregenerated source agent in `sourceTenantId`; strict agents RLS permits only the current tenant, so the intended cross-tenant source is invisible under the restricted production app role. Vault export is attempted under the same personal-tenant DB context. PGLite tests hide this and the real-PG worker test uses broad direct DB access rather than mounted auth/RLS. Add a narrowly authorized token-fenced cross-tenant primitive or safe split lifecycle, then mounted restricted-role PG claim/replay/takeover/deletion/account tests. Body is stale at `89cb3bec`/12 files; live is `b84b50c5`/22 files. Keep draft.

PR #848, opened 2026-08-22
security: make tenant membership mutations audit-atomic

### LINKED ISSUE #713: security: make tenant membership and invitation mutations audit-atomic
Exact develop `3559ac210` updates `user_tenants.role` in one transaction, appends required `tenant.member.role.update` audit afterward, then restores `previousRole` with an unconditional update if audit fails. A concurrent valid role change after the first commit can be overwritten by that compensation; sole-owner and owner-only checks can be invalidated across the gap. The only API evidence is a source-scan (`tenant-rbac-hardening.test.ts`); the browser spec mocks the backend. Move locked membership re-read, sole-owner check, role update, and required completion audit into one tenant-audited transaction; remove compensation. Add mounted route + deterministic real-PostgreSQL concurrent demotion/promotion and audit-failure tests proving one winner, exact owner count, no stale overwrite, and correct 403/409 boundaries. Keep UI tests separately. Local-only until #657 is durable.

lalalune:
Closes #713.

## What changed
- serialize tenant invitation, membership, role, metadata, deactivation, removal, join, accept, and leave mutations with required completion audits
- bind cross-tenant invitation/open-join transitions to the verified personal-session user inside an exact target-tenant RLS transaction
- consume invitation tokens before existing-member returns and revalidate the current locked verified-email identity at commit time
- derive owner grants and revocation cutoffs from the final locked requester/target state
- preserve canonical user-row -> tenant-lifecycle lock order and reject reserved/personal mutation surfaces
- add mounted matcher/RLS/replay/rollback coverage and deterministic real-Postgres race proofs

## Exact local evidence
- production-mounted PGLite: 15/15
- Biome on all 7 changed files: pass
- diff check: pass
- syntax bundle: pass

The real-Postgres race suite was implemented but DATABASE_URL was unavailable locally. Hosted exact-head Integration execution and independent GitHub review are required. Keep draft; do not merge until both are terminal green/accepted.


lalalune:
Independent exact-head audit: BLOCK at dc738e08ae641bcd3fb3375bb69582d4604fba41 (tree ed2aede5cc0c757bd9dafc35263f8cd5cc0921da). P1 lock inversion: invitation acceptance locks the users row before tenant lifecycle (user.ts 7935-7948 and 8105-8118), while leave locks tenant lifecycle before lockUserSession/users row (8274-8275). Concurrent acceptance for an existing member plus leave can deadlock; the new PG suite omits that race. P1 atomicity gap: Redis revocation is irreversible across PostgreSQL rollback. Platform removal revokes before its audited transaction, and platform/user role, removal, and deactivation paths revoke inside callbacks before completion audit. A forced audit failure can roll back database state while credentials remain revoked; the test at lines 479-533 explicitly expects that partial outcome. The PG topology mocks revocationStore and mounts bare routes on the same database authority rather than restricted production app-role/RLS. Keep draft; do not merge until lock order and durable revocation/audit semantics are unified and proven on production topology.

lalalune:
Exact-head blocker audit on `dc738e08ae641bcd3fb3375bb69582d4604fba41`: keep draft. Tenant-owner lifecycle locks still call PostgreSQL advisory functions in explicit/file-backed PGLite. The open-join path also prechecks membership outside the lock, then `onConflictDoNothing` unconditionally audits/returns `role=member`; a concurrent invitation/admin membership can win while this request emits a false audit/role. Under the lock, reread and return the actual membership and audit only an actual transition; add deterministic race coverage.

lalalune:
Blockers corrected on exact successor `b99efb42fadaf6dbba875b62b4d8f573eed06602`.

- Tenant lifecycle and session advisory locks now use the canonical PGLite predicate, so explicit/file-backed `STEWARD_DB_MODE=pglite` never calls unsupported PostgreSQL lock functions; real PostgreSQL locking remains intact.
- Open-join insertion now uses `RETURNING` and, on a concurrent conflict, re-reads the committed membership. The response and required audit report the actual committed role plus `alreadyMember`, instead of claiming a new member mutation/role for the losing request.
- A pre-existing exact-head TypeScript narrowing failure in invitation acceptance was also corrected.

Local exact-head evidence: mounted explicit-DB_MODE PGLite membership suite 15/15; static security contract 11/11 (combined 26/26, 168 assertions); API typecheck, Biome, and diff-check clean. A new real-Postgres adversarial regression holds the lifecycle lock, races two open joins, and requires one insertion/one existing-membership result with exact audits; hosted Postgres will exercise it. Keeping draft until hosted checks are terminal green and independently approved.

lalalune:
New-head re-audit remains BLOCK at `b99efb42fadaf6dbba875b62b4d8f573eed06602` (tree `d278108b779f5391dabec46adb5413bd1cef99cc`). The delta adds join locking/idempotent membership reporting, but prior atomicity failure remains: role/lifecycle/delete paths call `revocationStore.revokeUserTokens` inside the SQL audited transaction before membership mutation and completion audit commit. If the DB/audit transaction rolls back, the irreversible external Redis revocation survives, violating audit-atomic behavior. The new tests do not supply a transactional/outbox compensation boundary for that store. Exact branch is also stale on base `201c1869` while develop is `b86aee063` after the unsafe merge burst. Restack, move revocation to a durable commit-coupled/outbox lifecycle, and rerun real production-topology races. Keep draft.

lalalune:
Closed as a rejected stale carrier. Canonical user-to-tenant lock ordering and transactionally consistent revocation remain unresolved; its own fault test proves Redis survives SQL rollback.

PR #850, opened 2026-08-22
fix(capabilities): make production rate limits durable

### LINKED ISSUE #761: fix(capabilities): replace process-local production rate-limit fallback

lalalune:
## Summary

- use durable PostgreSQL sliding-window reservations when production Redis is absent
- keep configured-but-failing Redis fail closed and memory limits explicit development/test only
- force tenant RLS, fence agent/tenant deletion, and expose the capability limiter in readiness
- derive all Redis limiter prune/add/reset and advisory cleanup time from Redis TIME so hostile application clock skew cannot bypass limits
- prove exact cross-process boundary, restart, concurrency, Redis outage, restricted app-role, and skewed-clock behavior

## Exact head

`9c8ec208eb083888e6e1329fa2ea068c7a118fdf` on develop base `201c1869d40ffca8a207a32a24eecc7eb6f24cd6`.

## Validation

- PGLite limiter: 6/6, 196 assertions
- independent exact-head Redis Lua execution: requests 1-60 allowed, 61 denied without increment; TTL approximately 61s; reset from oldest live score and Redis time; expired-score cleanup verified
- hostile application clocks one year behind/ahead cannot affect the Lua time boundary
- real Redis boundary/reconnect: 1/1; outage/local wrappers 4/4
- real PostgreSQL two-context 60/61 boundary, restart, and deletion races: 2/2
- restricted app-role mounted/readiness/ACL probes: 1/1
- DB, Redis, plugin-capabilities, and API builds green
- repaired files Biome and diff-check clean
- hosted exact-head matrix and accepted PR review remain required

The plugin migration is namespaced `0004` and does not collide with the core migration journal. Composition with #794 must preserve both global Redis-required readiness and the capability limiter readiness check.

Closes #761


lalalune:
Exact-head audit BLOCK at `6b8676b91fe6dc0df00ecf7833ced09d4b6dcb60` (tree `d12305a19a7333e42168dd0002ed45f933c9e828`). Redis authority remains module-global and process-env based: generation A can retain its client when B rotates/removes bindings, while a fresh B replica counts in B Redis or PostgreSQL, splitting the supposed global 60/30 limit. Tests inject fixed contexts or mutate process state sequentially; they do not mount hostile A→B→missing/overlap Worker requests. Worker init also never runs the new capability PostgreSQL readiness check, which is wired only to Bun `/ready`. The atomic fail-closed PostgreSQL reservation and restricted-role PG evidence are strong, but this exact head cannot independently satisfy the authority gate and its body admits dependency on #794. Keep draft.

lalalune:
Superseding-head re-audit remains BLOCK at `9c8ec208eb083888e6e1329fa2ea068c7a118fdf` (tree `37061082b4b0a6a4f6c0124e8a1ff4fed1c7a055`). Redis `TIME` is a sound clock-skew fix, but the delta does not touch module-global Redis client/availability, direct process authority reads, A→B rotated/missing overlap, or missing Worker capability readiness. Old isolates can still count in Redis A while fresh B replicas count in B/Postgres, splitting the global limit. PR is also stale on base `201c1869` while develop is now `b86aee063`, with overlapping files requiring restack/audit. Zero reviews; only eight queued CodeQL jobs and no 62-check matrix. Keep draft.

lalalune:
Closed as a rejected stale carrier. Redis TIME fixes clock skew but module-global authority, A-vs-B isolation, Worker readiness, and exact hosted production proof remain unresolved.

PR #852, opened 2026-08-22
fix(security): compose durable recovery lifecycles

lalalune:
Closes #653
Closes #701
Closes #724
Closes #765
Closes #766

## Summary

- authoritatively retire signed-only Solana artifacts before agent deletion and reconcile signed evidence with recovery anchors
- recover durable non-Solana accounting effects, including replay-bound Monero/reference transfer identities
- make Monero pre-relay and relay outcomes crash-recoverable with deletion and rollback fences
- journal venue submissions before side effects and repair submitted audit evidence without resubmitting
- deliver required provider audits through a cross-replica outbox with leased claims, scheduling, RLS, and strict recovered identity validation
- validate pre-existing outbox audit rows by action, resource type, resource id, tenant, and outbox UUID before marking delivery complete

## Exact composition

- base: `201c1869d40ffca8a207a32a24eecc7eb6f24cd6`
- head: `f1e7c49390112f6eee923c5cfc792cdc49436836`
- 25 issue/composition commits; no #770/PR #847 bytes
- contiguous migration journal after existing `0113_personal_tenant_account_lifecycle`:
  - `0114_signed_artifact_retirement`
  - `0115_monero_pre_relay_lifecycle`
  - `0116_trade_order_recovery`
  - `0117_provider_action_audit_outbox_delivery`
- final composition also installs the previously missing tenant RLS policy and generated inventory for `trade_order_recoveries`

## Local exact-head evidence

- DB, Vault, API, and plugin-trading builds/typechecks: green
- focused API: 98 passed, 32 real-Postgres-gated skipped, 564 assertions
- focused plugin: 42 passed, 237 assertions
- RLS inventory/deployment: 5 passed, 88 assertions
- PGLite migration replay through `0117`: green
- 61 changed TS/JSON files: Biome clean
- `git diff --check`: clean

## Required gates

- hosted exact-head checks, including real PostgreSQL
- independent review and protected-branch approval
- composition decision with PR #849: #765 already supplies the richer pre-submit recovery journal and terminal replay behavior. Do not blindly layer #849’s separate terminal table/route patch; prefer this carrier semantics and port only unique fault regressions if needed.

This PR remains draft until every gate is terminal green.

Issue #766, opened 2026-08-20
fix(api): make provider required-audit outbox cross-replica idempotent


lalalune:
## Problem

Provider-action required-audit recovery is not cross-replica idempotent.

At exact `develop` `3559ac2108d118730ff99e0b8a2a9fdc55895706`, `ProviderActionService.recoverUnsignedIntents()`:

1. selects every undelivered outbox row without claiming/locking it;
2. checks `audit_events` for the correlation outside the append transaction;
3. appends through the separately advisory-locked audit writer;
4. marks the outbox row delivered afterward.

The code itself acknowledges the cross-replica race. Two replicas can both observe no signed event before either acquires the audit advisory lock. They then serialize their appends, but both append the same logical required event, producing duplicate signed evidence. A crash after append/before `delivered_at` has the same replay shape. The current correlation tuple also has no unique constraint, so the database does not fence it.

This affects the stated C2 invariant and any downstream consumer that treats signed audit events as exactly-once evidence or counts them.

## Required fix

- Claim a bounded batch transactionally with `FOR UPDATE SKIP LOCKED` and a reclaimable lease/token, or lock the exact outbox row while draining.
- Append the required audit and mark that exact outbox row delivered in one tenant-audited transaction on the same database handle.
- Bind the audit event to the immutable outbox identity and add a database-level uniqueness/idempotency fence appropriate for required-outbox events.
- Preserve retry after worker death; stale leases may be reclaimed without producing another signed event.
- Keep policy-reservation reconciliation independently retryable.

## Acceptance evidence

- Deterministic real-PostgreSQL two-process race on one undelivered row produces one signed event and one delivered outbox row.
- Crash/fault after append but before completion rolls back both, then a retry produces exactly one event.
- Stale claim takeover works; a stale worker token cannot append/mark after takeover.
- Multiple distinct outbox events with the same action/resource but different immutable identities retain the intended multiplicity.
- PGLite behavior and existing provider-action recovery suites remain green.

No matching open issue or PR was found during this exact-head audit.

Issue #765, opened 2026-08-20
fix(trading): durably recover post-submit order effects and audits


lalalune:
## Problem

Successful venue submission can permanently lose its required completion evidence.

At exact `develop` `3559ac2108d118730ff99e0b8a2a9fdc55895706`:

- Hyperliquid submits externally, then catches any `trade.order.submitted` audit failure and still returns/caches success (`packages/plugin-trading/src/routes/trade.ts`, successful submit path).
- Polymarket explicitly treats the post-submit audit as best effort and logs/proceeds after failure.
- replay returns the cached idempotency response; it does not drain/retry the missing audit.
- the cache is Redis-backed only when configured and contains a response envelope, not a durable DB recovery record binding venue result, accounting, audit, and webhook completion.
- the unknown-status path similarly caches `502` without a reachable evidence-bound venue reconciliation path.

The choice not to turn an already-landed order into an error is correct, but swallowing the completion failure is not: an externally executed trade can permanently lack required audit/recovery evidence, and retry cannot repair it.

Existing tests prove normal submitted audits and pre-submit denials. They do not inject a post-submit audit failure, replay the same key, and prove the missing effect is durably recovered without a second venue submission.

## Required fix

- Stage a durable per-order recovery/effect record before external submission, bound to tenant, agent, venue, session, canonical intent/idempotency key, and an owner lease/token.
- Checkpoint the exact venue result/order identity at the irreversible boundary.
- Drain required completion audit, spend/accounting, and any durable webhook through idempotent effects with stable occurrence time.
- Matching replay/reconciliation must repair pending effects and return the durable result without resubmitting.
- Represent ambiguous submit status explicitly and provide an evidence-bound venue reconciliation path; never relabel ambiguity as rejection or blindly retry submission.
- Keep authorization and session-revocation race evidence intact.

## Acceptance evidence

- Mounted Hyperliquid and Polymarket tests inject failure after the adapter reports success but before completion audit; response/result is preserved, one venue submit occurs, and replay/recovery writes exactly one required completion audit.
- Crash/failure after each individual effect proves retry completes only missing effects and does not move rolling-window timestamps.
- Concurrent replay/recovery proves one effect owner and no duplicate submit/audit/accounting/webhook.
- Ambiguous transport loss reconciles from exact venue evidence before any resubmission.
- Real Redis/real PostgreSQL coverage validates cross-process ownership and durable recovery.

No matching open issue or PR was found during this exact-head audit.

[...]

Issue #724, opened 2026-08-20
security: make Monero relay recoverable across process death


lalalune:
## Problem

The Monero transfer path has no durable provider checkpoint before `relay_tx`. A process can die after the daemon accepts the relay but before Steward inserts its terminal transaction row. The client then has no authoritative Steward record to replay or reconcile, and retrying may resubmit without an evidence-bound recovery path.

This is distinct from #701, which makes post-terminal Redis spend and aggregation effects durable.

## Required fix

- Stage a request/idempotency-bound transaction recovery anchor before the irreversible relay call.
- Persist the exact signed transaction/relay identity needed to reconcile without exposing secret material.
- On timeout/lost acknowledgement, preserve an explicit ambiguous state rather than reopening a fresh relay.
- Add an authenticated reconciliation/replay path that queries the configured Monero daemon and atomically transitions the anchored row with required audit evidence.
- Prove same-key replay and process-death recovery do not call `relay_tx` twice or double-count durable effects.
- Cover concrete daemon rejection separately from ambiguous transport loss.

## Acceptance evidence

Behavioral mounted tests plus a daemon-compatible integration fixture must cover: checkpoint-before-relay, lost acknowledgement, exact replay, reconciliation, mismatch rejection, and post-terminal effect retry. Source-order scans are not sufficient.

PR #857, opened 2026-08-22
fix(security): restore durable recovery migration chain

### LINKED ISSUE #653: fix(api): safely terminalize signed-only transactions before agent deletion

lalalune:
Successor to closed #852. Closes #653, #701, #724, #765, and #766.

This carrier composes signed-artifact retirement, Monero pre-relay recovery, trade-order recovery, and provider required-audit outbox delivery on exact current `develop`, preserving the merged wallet-account audit migration and generic intent delete fence.

The journal/files are intentionally: 0113 personal lifecycle → 0114 durable wallet audit → 0115 signed retirement → 0116 Monero → 0117 trade recovery → 0118 generic intent fence → 0119 provider audit outbox. The inserted/backfill migrations use timestamps newer than the already-shipped 0118 ledger tip, including an idempotent replay of 0114, so existing deployments do not silently skip the gap.

Exact head `71d3a07591732875e38bfcf3c21e8ac3cf647618` is based on `develop` `0f820da22d47e99cd4daf6abc40c9a29fab8296e`.

Evidence across the exact composition: focused carrier suites 52/52 (203 assertions); full PGLite adapter 19/19 (51 assertions) applying every migration through 0119; prior carrier DB/Vault/plugin-trading/API builds green; real-PG signed retirement/export/generic-boundary/personal-lifecycle and outbox/trade recovery evidence; changed-file Biome and diff-check clean. Final current-base composition added only conflict-free merged bytes plus the required tenant-audit helper import.

Do not merge until hosted exact-head checks pass, migration-gap coverage is green, and an independent reviewer approves.


lalalune:
Current local composition caveat (PR remains draft): DB/Vault/plugin-trading/API builds, full fresh PGLite through 0119, and DB focused 28/28 are green. A broad composed API batch produced 63 passes, 3 Redis skips, and 16 failures involving NODE_ENV leakage between suites, request-local vault mocks falling through to live RPC, and wallet-action duplicate terminal/422 behavior. A dedicated repair lane is reproducing the suites isolated and composed. This PR will not be marked ready until those failures are fixed or exact-head evidence proves them non-product and the harness is corrected.

lalalune:
Closed unmerged for containment during the active branch-protection deletion incident. Exact head 71d3a075 is preserved for independent audit and a current-develop restack if accepted; do not reopen or merge this carrier.

lalalune:
Exact audit at 71d3a075 / tree 3c4011c9: BLOCK and keep closed. 0117 tracks only audit delivery, not durable spend/accounting/webhook effect checkpoints; requiredAuditExists accepts any tenant row with metadata.tradeRecoveryId and does not bind action, actor, resource/session, time, or venue. Recovery tenant/agent/session columns lack FKs and deletion integration. 0115 permits retired rows without coupled evidence/digest and allows an arbitrary first evidence replacement. Replayed 0114 and 0117 do not FORCE RLS, while startup requires enabled+forced. There is no exact current-base restricted-role real-PG upgrade/restart proof. Draft, zero reviews, 61 checks queued; the 73-file resurrection does not satisfy #765.

PR #860, opened 2026-08-22
fix(adapters): enforce server-authoritative spend valuation

lalalune:
Post-merge corrective for #794. The merged durable rate-limit layer still accepted caller-controlled USD estimates on fund-moving adapter routes and reused durable counters across changed request-limit generations.

This correction:
- rejects caller-supplied estimatedUsd on swap, earn, bridge, and Spark money paths
- derives token and BTC notionals from the server price oracle and adapter-produced artifacts
- fails closed when authoritative valuation is unavailable (including unpriced Spark tokens)
- binds auth, proxy, and vault durable counter keys to the effective request limit
- preserves request-local Worker middleware compatibility
- adds focused understatement, unavailable-price, production fail-closed, and key-generation coverage

Exact tested head: 049f416a89b40a95d09babf0fd5129aab821684a
Exact base: f7da4649896a57c7d14f307d2be23eff8d2a6765

Validation:
- adapter policy gate: 22/22 (150 assertions)
- Redis authority wrappers: 15/15 (1050 assertions)
- auth keying: 26/26 (135 assertions)
- global limiter + Worker boot (isolated): 26/26
- Spark adapter: 3/3
- API and adapters TypeScript checks
- Biome and git diff --check


lalalune:
Exact-head review at `049f416a89b40a95d09babf0fd5129aab821684a`: BLOCK pending one remaining valuation-authority fix. Swap prices `metadata.amountIn` against caller-round-tripped `quoteInput.fromToken.address`; internal bridge does the same against `parsed.data.quote.fromToken.address`. `MockSwapAdapter.buildSwap` explicitly treats the quote as untrusted and its returned artifact omits the source-token identity, so the PR body claim that amount + token + chain come from the adapter artifact is not true on this head. Bind source-token identity into the trusted adapter-produced artifact (or a server-held/signed quote receipt), value only the bound artifact fields, fail closed on absence/mismatch, and add hostile cheap-token/tampered-quote tests for swap and bridge.

lalalune:
Keeping this draft: exact head 049f416a still values swap/bridge source-token identity from the caller-controlled quote while the adapter artifact does not bind that identity. Hosted green would not cure the authority gap. Repair must bind token identity into trusted adapter output or a server-held signed quote receipt, with hostile cheap-token/tampered-quote tests, then rerun exact-head gates and independent review.

PR #862, opened 2026-08-22
fix(db): complete four-role production RLS activation

lalalune:
Restacks the semantically accepted four-role RLS work from closed stale carrier #843 onto current `develop`, preserving the Redis/capability-rate-limit integrations merged afterward.

Closes #696.
Closes #669.
Closes #654.
Relates to #343; SEC-169 remains open until hosted and deployment acceptance are complete.

## Outcome

- provisions distinct bootstrap, restricted migrator, app, and platform database authorities
- replaces blanket runtime execution with exact function and ACL manifests
- orders production migration, privileged wrapper, plugin migration, and RLS activation phases
- makes personal user and tenant lifecycle invariants and revocation recovery durable
- adds fail-closed API, Worker, and proxy readiness checks against exact role and policy inventories
- integrates the current Redis readiness and capability rate-limit bucket topology

## Exact candidate

- base: `a521e7ae898fcddb0ce30381a5a9a597ac927d64`
- head: `fd0ed90be0fac8df70ab297ae1d771aedb2a2e4b`
- tree: `ea2ba031090cd1c6508c0cc9394c7a79b095bc9a`

## Exact local evidence

- real PostgreSQL four-role operator lifecycle: 1 pass, 92 assertions
- focused API, DB, Worker, and proxy suite: 129 pass, 904 assertions
- API and CLI TypeScript builds: green
- 12-package dependency build: green
- changed-file Biome and diff checks: green

## Pending gates

- exact hosted checks on this head
- independent GitHub review
- staging/deployment acceptance remains separate

PR #856, opened 2026-08-22
fix(staging): fail closed on database and migration drift

lalalune:
Successor to closed #797; closes #788.

This exact-current repair rejects foreign/shared database targets before mutation, binds every applied plugin-ledger prefix to the required relations/routines/triggers/RLS policies, and composes readiness inventory with the merged 0114 wallet lifecycle and 0118 intent-fence objects.

Exact head `527813e3b01298b49f5d0cf4a2b07424dab02073` is based on `develop` `0f820da22d47e99cd4daf6abc40c9a29fab8296e`.

Evidence: DB build green; ledger/plugin 16/16 (60 assertions); release/startup 6/6 (13); real-PG adversarial inventory 8/8 (30 assertions) before unrelated final rebases and final-SHA fresh-target PG recheck; 21 changed TS/JSON files Biome-clean; diff-check/tree clean.

External staging database drift remains an operational acceptance gate. Do not merge until hosted exact-head CI passes and an independent reviewer approves.

PR #863, opened 2026-08-22
fix(security): restore durable recovery chain through 0119

### LINKED ISSUE #653: fix(api): safely terminalize signed-only transactions before agent deletion

### LINKED ISSUE #696: blocker(security): repair merged personal-tenant account lifecycle invariants

### LINKED ISSUE #701: security: make non-Solana vault spend effects durably recoverable
## Gap

Exact current develop `3559ac210` still treats non-Solana vault accounting as best-effort after the transaction success write (`packages/api/src/routes/vault.ts` around the `solanaNativeExecutionToken` else branch):

- `recordVaultSpend(...)` is fire-and-forget;
- aggregation errors are logged and swallowed;
- the transaction is already terminal before either effect;
- outer idempotency/replay can return the stored/cached success without an effect-completion claim.

The source comment explicitly acknowledges a residual undercount. A Redis outage or process crash after the DB success write can therefore permanently omit spend/cumulative-policy state, allowing later requests to evaluate against an understated window. #699 adds durable recovery effects only to its Solana-owned path; the ordinary EVM/offline branch remains outside that protocol.

## Acceptance

- Persist a stable effect identity and occurrence time before/with the terminal transaction transition.
- Make spend and aggregation writes idempotent and retryable from a durable incomplete-effects state.
- Every same-key replay/reconciliation path finishes incomplete effects before returning success; concurrent workers have a bounded owner/lease/CAS and cannot double count or move the event into a later window.
- Define whether offline signed artifacts consume spend separately from broadcast artifacts and behavior-test that contract.
- Add crash/fault and real-Redis tests covering failure after DB finalization, replay recovery, concurrency, and rolling-window timestamp stability.

Do not broaden the Solana corrective or merge without exact hosted Redis/Postgres evidence and independent approval.

### LINKED ISSUE #724: security: make Monero relay recoverable across process death

### LINKED ISSUE #765: fix(trading): durably recover post-submit order effects and audits

### LINKED ISSUE #766: fix(api): make provider required-audit outbox cross-replica idempotent

lalalune:
## Summary
- restore contiguous migrations 0113 through 0119 around the already-shipped 0118 fence
- safely terminalize signed artifacts before deletion
- add durable non-Solana spend recovery, Monero relay lifecycle, trade recovery, and provider audit outbox fencing
- preserve request-local custody authority and update pre-staged wallet failures without duplicate inserts

Closes #653
Closes #701
Closes #724
Closes #765
Closes #766

Does not close #696; canonical personal-tenant deletion remains a separate corrective.

## Exact-head verification
- carrier/API batch: 105 pass, 13 environment-gated skips, 0 fail
- post-#860 focused adapter/wallet/approval/gateway batch: 71/71
- full carrier batch: 89 pass, 3 Redis-gated skips, 0 fail
- PGLite migration replay through 0119: pass
- migration-gap regression: pass
- DB/Vault/trading/API/Redis builds: pass across final composition
- Biome and diff-check: clean


lalalune:
Exact-head migration audit at `9f5780be0be3da72d05c8a74d349084c5bda337c`: BLOCK. This head retroactively changes the already-shipped 0114 journal timestamp from `1787306400000` to `1787306400001`, inserts 0115/0116/0117 at later values, then retains already-shipped 0118 at the earlier `1787306400000`. Existing databases keep the original 0114 ledger identity, so the exact-ledger verifier rejects their upgrade; the resulting journal is also non-monotonic. Missing historical numbers cannot be inserted before shipped 0118. Preserve 0114 and 0118 byte-for-byte, renumber all new migrations after 0118 (for example 0119 signed-artifact, 0120 Monero, 0121 trade, 0122 outbox), and prove exact current-ledger upgrade/replay before merge. #862 personal lifecycle must then follow at 0123.

lalalune:
Blocking migration-order issue on exact head `9f5780be0be3da72d05c8a74d349084c5bda337c`:

- current `develop` already ships `0114_durable_wallet_claim_account_audit` and `0118_generic_intent_execution_delete_fence` with `when: 1787306400000`
- this head retroactively changes shipped `0114` to `1787306400001`, inserts new `0115`–`0117` at `...002`–`...004`, then leaves the later `0118` journal entry at the earlier `...000`
- the exact-ledger integrity now on `develop` will reject an existing database whose already-applied `0114` row retains its shipped timestamp; the resulting journal chronology is also non-monotonic

Please preserve the shipped `0114` and `0118` journal identities byte-for-byte. Missing historical numbers cannot safely be inserted before already-shipped `0118`; renumber this PR's four new migrations contiguously after it (for example `0119`–`0122`) and prove an upgrade from the exact current ledger, plus PGLite and real-Postgres recovery. PR #862's personal-lifecycle migration will then move after this chain.

lalalune:
Exact-head refresh: `e370f606459d1eca6516e3c82f826e2da80fe434` now includes current develop `dd5619c29101de251c392b0d9e182587c1ea34e3` (#856/#861/#864/#865). Conflict resolution preserves the Postgres-backed provider audit-outbox scheduler even when Redis is unavailable, while retaining the staged startup boundary and outer-commit cache fences.

Post-composition evidence on final bytes:
- authenticated adapter quote receipts: 18/18 adapter unit + 25/25 mounted policy gate
- Worker runtime sentinel: 22/22
- request database membrane: full file passes after explicit 20s bounds on four full-migration PGLite proofs
- contiguous PGLite replay applies 0113→0114→0115→0116→0117→0118→0119
- prior recovery carrier batches remain 105 pass/13 env skips, 71/71, and 89 pass/3 Redis skips
- no dependency overlay remains; worktree clean.

lalalune:
Re-audit at new exact head `e370f606459d1eca6516e3c82f826e2da80fe434`: the migration blocker is NOT fixed. `_journal.json` still rewrites shipped 0114 from `1787306400000` to `1787306400001`, inserts 0115/16/17 at `...0002/3/4`, then places shipped 0118 back at `...0000`. A green fresh PGLite replay cannot prove an upgrade of an existing ledger whose 0114 row retains the shipped identity. Keep blocked: preserve 0114/0118 byte-for-byte and move every new migration after 0118 with monotonic identities, then test upgrade from the exact pre-PR journal rather than only fresh replay.

lalalune:
Exact-head BLOCK at e370f606 (tree deca9e09), base dd5619c2. Do not merge. P0 deterministic upgrade failure: this PR rewrites the already-shipped 0114 journal timestamp, so a normal current-develop database with the original applied 0114 ledger identity is rejected before migration. If 0114 was skipped, the post-migration exact inventory does not allow trade_order_recoveries or the two new steward routines, so the migration rolls back. The static gap test never exercises exact dd5619c2 ledger to head on real PostgreSQL. The rejected #857 recovery blobs are otherwise byte-identical: audit-only recovery omits accounting/spend/webhook completion, requiredAuditExists is spoofable, FKs/deletion integration are missing, signed evidence remains weakly constrained, and RLS FORCE/restricted-role upgrade proof is absent. Issues #653/#701/#724/#765/#766 remain open.

lalalune:
Closed unmerged: deterministic current-base upgrade failure and unchanged security/recovery blockers. Preserve the exact head for forensic comparison; use a new protected successor only after real current-base PostgreSQL upgrade and complete recovery/integrity gates pass.

PR #869, opened 2026-08-22
fix(api): bind tenant policy lookup to agent ownership

### LINKED ISSUE #711: fix(api): remove process-local tenant default-policy mutation

lalalune:
Closes #711

## Summary
- bind policy lookup to both agent id and the authenticated tenant through the agents table
- keep empty policy sets explicit and engine-denied instead of treating them as approved
- add separate-process/restart real-Postgres proof under a NOSUPERUSER/NOBYPASSRLS app role across two tenants

## Verification
- composed cleanly on exact develop 260ed8112
- real PostgreSQL/RLS proof: 1/1, 4 assertions across owner and two restricted processes
- tenant policy retirement suite: 3/3, 15 assertions
- API plus dependency builds passed before conflict-free current-tip merge
- exact current diff Biome and diff-check pass

### DIFF
--- a/packages/api/src/__tests__/fixtures/tenant-policy-lookup-process.ts
+++ b/packages/api/src/__tests__/fixtures/tenant-policy-lookup-process.ts
@@ -1,0 +1,48 @@
+import { closeDb } from "@stwd/db";
+import {
+  getPolicySet,
+  policyEngine,
+  withAuthenticatedTenantDatabase,
+} from "../../services/context";
+
+const tenantId = process.env.TEST_TENANT_ID;
+const ownAgentId = process.env.TEST_OWN_AGENT_ID;
+const foreignAgentId = process.env.TEST_FOREIGN_AGENT_ID;
+
+if (!tenantId || !ownAgentId || !foreignAgentId) {
+  throw new Error("tenant policy lookup fixture environment is incomplete");
+}
+
+try {
+  const result = await withAuthenticatedTenantDatabase(
+    tenantId,
+    "tenant-policy-restart-proof",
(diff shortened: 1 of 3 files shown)


lalalune:
Containment: restored draft after it was externally marked ready again. Exact head 86f173f6 has zero reviews, a failing Lint/Metadata check, and most required checks remain nonterminal. Security-sensitive tenant policy authority is under independent audit; do not merge or mark ready.

PR #881, opened 2026-08-22
fix(db): complete four-role production RLS activation

lalalune:
## Summary
- restack the closed #862 four-role/personal lifecycle work on the shipped 0119-0122 ledger as migration 0123
- preserve #873 narrow shipped-journal recovery and current tenant-before-user lock ordering
- enforce restricted app lifecycle writes through a scoped SECURITY DEFINER wrapper bound to both authenticated tenant and user settings
- derive the durable token-revocation epoch inside the definer; callers cannot suppress or future-poison it
- complete production bootstrap ownership/ACL/function inventories and mounted startup/deletion evidence

Supersedes closed #843 and #862. Depends on merged #872 and composes #873/#877/#882.

## Exact-head evidence
- base: `b489cbbf9ed8f815037b34efc9442239ed7297b3`
- head: `df63f5af5304d6c164202e2dbe21531bd602d70d`
- real-PG shipped-ledger topology: 1 pass, 111 assertions (122.64s), including hostile actor substitution and missing-session-user probes
- DB TypeScript build, focused Biome, and diff check pass

The PostgreSQL topology starts from the exact shipped through-0118 ledger, applies 0119-0123 under separated migration/bootstrap/app/platform roles, validates function/ACL inventories, and exercises mounted API lifecycle, evidence preservation, tenant deletion, hostile direct app-role calls, and lock/race behavior.


lalalune:
Exact immutable-head audit of 66ce2740: BLOCK; closing unmerged. P1: this patch rewrites shipped packages/plugin-capabilities/drizzle/0003_tenant_rls_policies.sql. Existing deployments retain the old ledger hash, while current plugin-migrate.ts hashes the rewritten bytes and migrate.ts rejects the historical row; fresh and upgraded installations diverge. Put the evolution in a new 0005 migration/journal/fingerprint entry with deterministic conditional-effect modeling. P1: exact tenant-rls-policy-migration suite is 3 pass/1 fail because rls-bootstrap.sql still grants EXECUTE ON ALL FUNCTIONS in steward_bootstrap/steward_rls to app after the named ACL reset, creating a live/persisted global SECURITY DEFINER overgrant if activation fails. P1: membership PATCH locks tenant_owner then its UPDATE trigger takes platform_user, while lifecycle deactivation/delete locks platform_user then tenant_owner; the test races helper-vs-helper, not the opposing production paths. Functional blocker: platform global metadata performs UPDATE users, but the exact platform ACL/manifest grants SELECT only. The separate local PR871 legacy-ledger corrective is not superseded. No merge or issue closure is justified from this head.

lalalune:
Closure raced a final head update to df63f5af. Exact 66ce2740..df63f5af read-only delta audit confirms the disposition remains correct: shipped plugin 0003 is still rewritten with no 0005/fingerprint repair; broad app EXECUTE ON ALL remains; tenant/user lock inversion remains; platform/app UPDATE ACL posture is unchanged. The delta usefully binds deactivation actor authority and derives revocation time internally, but that is orthogonal to the four release blockers. Final df63 remains BLOCK/CLOSED UNMERGED.

PR #884, opened 2026-08-22
fix(sdk): use WebAuthn v13 browser call contract

0xSolace:
## Summary

- pass server-issued WebAuthn JSON through the SimpleWebAuthn v13 `optionsJSON` wrapper
- fix all three SDK call sites: passkey login, registration, and MFA
- make the SDK mocks assert the production browser-library argument shape

Dependency audit: `@simplewebauthn/browser` resolves to v13.3.0, whose registration and authentication declarations both require `{ optionsJSON, ... }`. This does not change the API server or its privacy-preserving `allowCredentials` policy.

## Exact-head verification

- base: `6275484e2716fa7df5ba5a5d753330f0a4568566`
- head: `cf9e3bc494cc2d2be98860d4781faee0a660b7fa`
- full SDK suite: 293 tests, 1,300 assertions
- SDK TypeScript build
- Biome on both changed files
- bidirectional mutation: all three call-shape assertions fail with the old direct-JSON calls and pass with this change

### DIFF
--- a/packages/sdk/src/__tests__/auth-add-passkey.test.ts
+++ b/packages/sdk/src/__tests__/auth-add-passkey.test.ts
@@ -68,27 +68,31 @@ function restoreBrowserShim(): void {
   globalThis.window = originalWindow;
 }
 
+const startRegistration = mock(async () => ({
+  id: "credential-1",
+  rawId: "credential-1",
+  response: {
+    clientDataJSON: "client-data",
+    attestationObject: "attestation",
+  },
+  type: "public-key",
+}));
+
+const startAuthentication = mock(async () => ({
+  id: "credential-login",
+  rawId: "credential-login",
+  response: {
+    clientDataJSON: "client-data",
+    authenticatorData: "authenticator-data",
+    signature: "signature",
+    userHandle: "u-1",
+  },
+  type: "public-key",
+}));
+
 mock.module("@simplewebauthn/browser", () => ({
-  startRegistration: async () => ({
-    id: "credential-1",
-    rawId: "credential-1",
-    response: {
-      clientDataJSON: "client-data",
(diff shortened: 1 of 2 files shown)


lalalune:
Interim exact-head hold for f8bd404f: the v13 optionsJSON shape is directionally correct and the changed auth-add-passkey suite reproduces 5/5, but  fails because packages/sdk changed without an SDK version or CHANGELOG update. Add a precise Unreleased/Fixed entry (no version bump if policy permits) and rerun the metadata gate. Also add/reproduce a production-browser test using the real @simplewebauthn/browser v13 call boundary; current tests mock the library and only assert the wrapper object. A local SDK build attempt reached unrelated existing WebCrypto BufferSource typing errors in auth.ts/client.ts, so clean base-vs-head build classification is still required. Keep draft; no acceptance from 61 queued checks/zero reviews.

lalalune:
Final exact audit of f8bd404f: runtime semantics PASS, delivery BLOCK. Registration, login, and MFA correctly use SimpleWebAuthn v13 optionsJSON; direct real @simplewebauthn/browser 13.3.0 conversion passed, focused 36/36, full SDK 293/293, SDK tsc/Biome/diff green. Required repairs: exact public-package metadata gate is red because the SDK changed with no CHANGELOG/version metadata; and the published SDK dynamically imports @simplewebauthn/browser but declares it only as a devDependency, with no optional peerDependencies range or docs telling consumers this path now requires v13. Add an optional peer range ^13.0.0, peerDependenciesMeta/docs, and precise Unreleased/Fixed entry (version per policy), then rerun metadata/full SDK/direct browser boundary. Keep draft; zero reviews/nonterminal exact checks.

PR #890, opened 2026-08-22
fix(auth): backport migrated auth-store reuse to main

0xSolace:
## Summary

Narrow main backport of e5f84b782972568027460bec042d1b25a1df3265 from develop PR #832. A PostgreSQL-backed auth store now checks for the already-migrated auth_kv_store relation before attempting DDL, so a restricted application role can serve email/passkey challenges against a provisioned database.

This intentionally excludes sibling commit 482242f1. That commit modifies the later RLS deployment-safety subsystem, whose source files do not exist on current main, and cannot be cherry-picked without importing its broader migration/release dependency chain. The runtime auth-store change is independently applicable and tested on main.

This PR is a draft and must not bypass main branch protection.

## Validation on this exact main-based head

- @stwd/auth dependency build: 7/7 packages passed
- store-backends suite: 19 passed / 0 failed
- git diff --check: passed

## Operational scope

No database migration or deployment is included. This avoids application-role DDL only when the canonical auth table is already present; absent/incomplete schema still fails closed through the existing path.

### DIFF
--- a/packages/auth/src/__tests__/store-backends.test.ts
+++ b/packages/auth/src/__tests__/store-backends.test.ts
@@ -4,6 +4,7 @@ import {
   buildBackend,
   MemoryBackend,
   NamespacedStoreBackend,
+  PostgresBackend,
   RedisBackend,
   type RedisLike,
   type StoreBackend,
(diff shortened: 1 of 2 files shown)


lalalune:
Closed unmerged as a containment boundary: this PR targets main directly, while current work is authorized only for develop-side audit/correction and repository protection is repeatedly being deleted/recreated. The branch is preserved for read-only semantic audit or a future explicitly authorized release process. This disposition is not a claim that the narrow patch is semantically bad; it prevents an unreviewed/queued main promotion.

PR #895, opened 2026-08-22
fix(auth): scope passkey login to typed email

wakesync:
## Summary

- resolve passkey login options from the normalized typed email and return only that user's credential IDs and stored transports
- return the same generic 404 for unknown accounts and existing accounts without passkeys
- let SDK callers disable automatic registration so an owning UI can enter its email/OTP fallback before any WebAuthn ceremony

## Privacy / UX decision

This is the product-approved tradeoff: typed-email credential scoping intentionally supersedes the prior discoverable-login enumeration test. The rate-limited pre-auth route now reveals passkey availability through 200 versus 404 so the browser cannot offer an unrelated discoverable credential for the RP. Unknown and no-passkey emails still execute the same lookup and receive the identical generic 404 body. Public error copy does not distinguish account existence from passkey absence.

The companion Eliza UI change will call signInWithPasskey with fallbackToRegistration disabled, route this generic 404 directly to email-verified enrollment, and reserve cancellation recovery for a browser-owned WebAuthn cancellation. Merge/deploy order will be coordinated; this PR does not deploy anything.

## Verification

- bun test packages/api/src/__tests__/auth-passkey-enumeration.test.ts packages/sdk/src/__tests__/auth-add-passkey.test.ts (9 passed, 42 expectations)
- bunx turbo run build --filter=@stwd/api --filter=@stwd/sdk (24/24 tasks)
- bunx @biomejs/biome check on all four changed files

### DIFF
--- a/packages/api/src/__tests__/auth-passkey-enumeration.test.ts
+++ b/packages/api/src/__tests__/auth-passkey-enumeration.test.ts
@@ -8,8 +8,12 @@ setDefaultTimeout(120_000);
 const ENABLED_TENANT_ID = "passkey-enumeration-enabled";
 const DISABLED_TENANT_ID = "passkey-enumeration-disabled";
 const KNOWN_EMAIL = "passkey-known@example.test";
+const NO_PASSKEY_EMAIL = "passkey-none@example.test";
+const OTHER_EMAIL = "passkey-other@example.test";
 const UNKNOWN_EMAIL = "passkey-unknown@example.test";
-const SECRET_CREDENTIAL_ID = "credential-id-must-never-leave-options-route";
+const KNOWN_CREDENTIAL_ID = "credential-id-for-known-email";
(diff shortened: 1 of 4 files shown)


lalalune (approved):
Independent exact-head audit at `cd094293ddffb654231f6942416909df532b4a92`: approved. Verified the typed-email credential query returns only that user's credential IDs/transports; unknown and no-passkey accounts share the same generic 404; the SDK opt-out does not invoke WebAuthn registration. Exact validation: Turbo dependency/API/SDK build 24/24, API privacy suite 3/3 (22 assertions), SDK suite 6/6 (20 assertions), Biome and diff checks green. Delivery order remains important: #900 should land on `main` first, then this branch must be restacked and revalidated because both touch auth surfaces.

wakesync:
Superseded by merged #900, which preserves the non-enumerating constant-200 passkey contract and includes the RP-scoped recovery stack from #914. This branch is stale against main and its hosted matrix is red; closing to prevent accidental merge.

Issue #883, opened 2026-08-22
fix(db): reconcile merged four-role RLS migration and lock/ACL regressions


lalalune:
## Context

PR #881 final head `df63f5af5304d6c164202e2dbe21531bd602d70d` was externally merged as `f0ce51c86eb0d92e727b5108ec3133879a422887` with zero reviews and 62 exact checks queued after an exact audit had closed it unmerged. The audited blockers are therefore now on `develop`.

## Required corrective

- Restore `packages/plugin-capabilities/drizzle/0003_tenant_rls_policies.sql` byte-for-byte to its shipped form. Move the new RLS evolution into a new immutable `0005` plugin migration with synchronized journal/catalog/fingerprint state and deterministic handling of conditional effects. Prove an existing old-0003 ledger upgrades without hash rejection and matches a fresh install.
- Remove `GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA steward_bootstrap, steward_rls TO app` from the bootstrap transaction. Grant only the exact app allowlist and reject unknown/future SECURITY DEFINER functions before commit; activation failure must never leave a persisted overgrant.
- Establish one lock order across membership PATCH/UPDATE triggers and user deactivation/deletion lifecycle helpers. Add a two-connection real-Postgres route-vs-lifecycle race proving no deadlock and correct sole-owner/tenant isolation.
- Reconcile platform route behavior with exact ACLs: production global metadata UPDATE must either receive a narrowly scoped, manifest-checked privilege/SECURITY DEFINER path or stop issuing a direct UPDATE under the SELECT-only platform role.
- Preserve the useful final-head improvement that binds deactivation actor authority and derives revocation time internally.
- Compose with the separate exact historical capability-grants adoption corrective for #788; PR #881 did not implement it.

## Acceptance evidence

- Real PostgreSQL old shipped plugin ledger to corrected tip, idempotent rerun, fresh-install equivalence, and hostile hash/function/trigger/policy drift rejection.
- Restricted app/platform/bootstrap/migrator role startup and failure-window tests.
- Two independent PostgreSQL connections for the opposing lock paths.
- Focused DB/API/proxy/Compose tests, builds, static ACL/function inventories, full exact hosted matrix, and independent approving review on the final SHA.

Refs #881, #669, #654, #788, #657.

lalalune:
[...]

PR #903, opened 2026-08-22
fix(trading): complete durable post-submit recovery

lalalune:
## Summary
- recover and finish the post-merge durable trade journal corrective on current `develop`
- persist immutable effect metadata in appended migration 0124 so cold replicas can finish required audits without request-local authority
- process durable submitted/ambiguous recovery before mutable rate-limit, session, and Redis idempotency gates
- safely reclaim only stale `prepared` claims before any venue I/O; old claim tokens cannot checkpoint submission
- preserve the fenced Hyperliquid wallet in the recovery row and require exact venue evidence before ambiguous reconciliation
- retain one stable occurrence time and idempotently finish a committed audit after process death
- harden Solana signed-artifact response inspection and duplicate vault action anchors recovered with the same post-merge corrective

## Exact-head evidence (`92dad9317a5517ab1ea446c34d9d833e2834307d`)
- API dependency graph build: 24/24 packages
- durable recovery unit matrix: 4/4, 26 assertions
- mounted Hyperliquid fence/replay: 8/8, 64 assertions
- mounted Polymarket order/recovery: 37/37, 186 assertions
- vault signed-artifact inspection: 7/7, 13 assertions
- fresh migrated PostgreSQL + disposable real Redis, two cold worker processes: 1/1, 5 assertions
- changed-file Biome and `git diff --check`: green

## Acceptance boundary
No external venue was contacted. Merge and issue closure remain gated on independent review and the terminal protected hosted matrix. The current base also has a known lifecycle fixture regression tracked by PR #899; this branch must be restacked if #899 lands first.

Refs #765
Refs #883
Refs #657

### DIFF
--- a/packages/api/src/routes/vault.ts
+++ b/packages/api/src/routes/vault.ts
@@ -882,19 +882,24 @@ async function findActionByReferenceId(
   return existing ?? null;
 }
 
-async function findActionByIdempotency(agentId: string, actionType: string, keyDigest: string) {
-  const [existing] = await db
+async function findActionByIdempotency(
+  agentId: string,
+  actionType: string,
+  idempotencyKeyDigest: string,
+) {
+  const rows = await db
     .select()
     .from(transactions)
     .where(
       and(
         eq(transactions.agentId, agentId),
         eq(transactions.actionType, actionType),
-        sql`${transactions.actionPayload}->>'idempotencyKeyDigest' = ${keyDigest}`,
+        sql`${transactions.actionPayload}->>'idempotencyKeyDigest' = ${idempotencyKeyDigest}`,
       ),
     )
-    .limit(1);
-  return existing ?? null;
+    .limit(2);
+  if (rows.length > 1) throw new Error("Duplicate action idempotency anchors");
+  return rows[0] ?? null;
 }
 
 function requireBroadcastActionIdempotency(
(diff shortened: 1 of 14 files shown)


lalalune:
Blocking correctness findings at exact head `92dad9317a5517ab1ea446c34d9d833e2834307d`:

1. `0124_trade_recovery_effect_metadata.sql` adds/backfills and immediately sets `effect_metadata` NOT NULL. Our migration-first staging release keeps the old image serving during migration; that image still inserts recovery rows without this column, so an order during the migration-to-rollout window fails. Use expand/backfill/dual-write/contract, or a backward-compatible DB default/trigger.

2. Stale prepared takeover bypasses the prior idempotency key and re-enters the normal spend-reservation path. The original attempt reserves/increments spend before journaling; recovery reserves it again, potentially double-counting daily spend or permanently hitting the cap. Make spend reservation durable/idempotent by recovery identity and assert the daily spend remains exactly once.

3. Recovery treats any object from Polymarket `getOrder` (and similarly weak Hyperliquid status data) as proof of submission without binding returned order identity and effect fields to the journaled immutable tuple. The new fixture even returns an unrelated order id and passes. Parse strict venue schemas and require order/hash, wallet/funder, market/token, side, and amount to match before marking submitted.

Claim-token fencing and the durable effect metadata direction are good, but these P1 gaps make the current head unsafe to merge.

PR #909, opened 2026-08-22
test(web): bind invitation races to signed sessions

lalalune:
## Summary

- include the invitation browser contract in the protected trust-UX Playwright suite
- replace alg-none session fixtures with HMAC-signed fixtures validated before browser use
- combine a pending old acceptance, session credential rotation, and retained-document query navigation in one race proof
- assert the old request uses only the old bearer and the new invitation uses only the rotated bearer

## Stack

This is intentionally stacked on #904 because the expanded trust suite also executes its provider-error selector correction. The only delta in this PR is invitation evidence plus trust-suite inclusion.

## Exact-head verification

Head: b94dd6d1ffeff4c558c58fbd08a6360bbf65637f

- repository typecheck: 36/36 package builds plus web TypeScript passed
- production Next build: passed
- protected trust-UX Chromium suite: 21/21 passed, including 6 invitation cases

Refs #755

### DIFF
--- a/web/e2e/accept-invitation.spec.ts
+++ b/web/e2e/accept-invitation.spec.ts
@@ -1,3 +1,4 @@
+import { createHmac } from "node:crypto";
 import { expect, type Page, type Route, test } from "@playwright/test";
 
 function base64UrlJson(value: unknown): string {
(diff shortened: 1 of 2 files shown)


lalalune:
Review finding: the race assertions and trust-suite inclusion are useful, but the session proof remains synthetic. The test signs an HS256-shaped token with a hardcoded key, verifies it with the same helper, omits production issuer/audience/jti, and intercepts acceptance requests before production verifySessionToken sees it. Please mint through the production signing authority with the test STEWARD_JWT_SECRET, independently verify through the production verifier, and exercise the closest live API auth boundary. Since #904 is already integrated separately, only the b94dd6d delta should ultimately land.

Issue #746, opened 2026-08-20
test(react): exercise linked-account mutations through mounted interactions


lalalune:
## Problem

`packages/react/src/__tests__/StewardLinkedAccounts.test.tsx` renders only the initial SSR shell. Its final test reads `StewardLinkedAccounts.tsx` and asserts SDK method-name substrings, so none of the reachable unlink, phone OTP, Ethereum/Solana wallet, OAuth, Telegram, or Farcaster mutation flows actually execute.

The source scan can pass while argument binding, cancellation, busy/error state, refresh ordering, callback delivery, or last-login-method protection is broken.

## Acceptance

- Replace the SDK-call source scan with mounted interaction coverage using the real component and controlled client promises.
- Exercise unlink plus each link family (phone, wallet, OAuth, Telegram, Farcaster), asserting exact inputs and callback/refresh behavior.
- Prove rejection surfaces a stable user-visible error and does not report success or lose the current account list.
- Add at least one delayed/stale completion case so an older refresh cannot overwrite a newer mutation/result or auth epoch.
- Keep the existing signed-out/render gating assertions where useful.
- Run the full React suite, React/SDK builds, Biome, and diff check on the exact commit.

lalalune:
Reopened after the unsafe unreviewed merge burst into develop f7125cdab867bd96e8f8428adab6a2bcd06209b0. Exact post-merge audit confirms the residual blocker: PR #805 merged while logout does not invalidate an older linked-account refresh generation. Keep open for a current-develop corrective with independent exact-head review and terminal hosted gates.

lalalune:
Resolved in the mounted auth behavior aggregate merged by #790; the exact aggregate commit is an ancestor of current develop. Closing after ancestry and behavior-evidence reconciliation.

lalalune:
Reopened from independent exact merged-byte audit: logout does not invalidate an older linked-account refresh generation.

PR #910, opened 2026-08-22
fix(react): fence linked-account refresh on sign-out

lalalune:
## Summary

- invalidate the linked-account refresh generation before the signed-out early return
- prevent a request started in a prior authenticated epoch from publishing stale accounts or firing onLoaded after sign-out
- preserve all mounted unlink, phone, wallet, OAuth, Telegram, Farcaster, error, and refresh-order coverage

## Exact-head verification

Head: cf3dd95f97c11862869f28535e2a0124c4ccb174

- focused StewardLinkedAccounts suite: 9/9 passed
- full sequential React suite: passed
- SDK build and React build: passed

Refs #746

### DIFF
--- a/packages/react/CHANGELOG.md
+++ b/packages/react/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to `@stwd/react` are documented here.
 
 ### Fixed
 - `StewardLogin` no longer emits the `webauthn` autocomplete token on the email input, so browser passkey conditional-mediation autofill cannot hijack a brand-new-email signup. Explicit passkey button flow is unchanged.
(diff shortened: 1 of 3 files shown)


lalalune:
Security review finding: incrementing a generation before the signed-out return only fences true-to-false sign-out. Reads remain unbound to token/user/tenant changes while isAuthenticated stays true, state is not cleared on auth-epoch changes, and old-user account data can render after reauthentication as another user until the new read resolves. Please key reads and mutation callbacks to a stable auth epoch, clear primaryMethods/linkedAccounts immediately on epoch change, and add a delayed old-user -> sign-out -> new-user regression plus token-rotation coverage.

PR #911, opened 2026-08-23
fix(trading): scope operator authority and durable replay

lalalune:
## Summary
- require the deny-by-default `platform:trade:operator` scope for platform-key operator recovery and propagate the key hash/scopes into request context
- centralize platform-key authentication, scope resolution, hashing, and redacted configuration failures
- make HMAC-chained durable transfer evidence authoritative over cached Redis success
- add a cross-instance Redis regression proving requested-only audit evidence prevents a duplicate capital movement

## Verification (exact head `9fdaaaa63e980bee7ebba6e90c90f7bf6a2e60b3`)
- `bun test packages/auth/src/__tests__/platform.test.ts` (13 pass)
- `bun test packages/plugin-trading/src/__tests__/operator-recovery.test.ts -t "mounted HIP-3 collateral transfer"` (10 pass)
- `bunx turbo run build --filter=@stwd/plugin-trading --filter=@stwd/api --filter=@stwd/auth` (24 dependency/package tasks pass)
- Biome on all changed files; `git diff --check`

The full plugin-trading baseline run completed with 163 pass, 14 skip, and 26 unrelated failures already present outside these changed behaviors (principally stale workspace build exports, the older incomplete PolicyEngine mock, Polymarket expectations, and one rate-limit timeout). The focused mounted transfer suite and all affected builds are green.

## Deployment note
Operator platform keys must have `platform:trade:operator`, `platform:*`, or `*` in `STEWARD_PLATFORM_KEY_SCOPES`; unscoped valid keys now fail closed.

Refs #750

### DIFF
--- a/packages/api/src/middleware/operator-auth.ts
+++ b/packages/api/src/middleware/operator-auth.ts
@@ -6,9 +6,10 @@
  * plane-down scenario the recovery feature exists to solve).
  *
  * This middleware accepts EITHER credential:
- *   1. A platform key — header `X-Steward-Platform-Key`, validated by
- *      `isValidPlatformKey`. This is the cross-tenant operator credential
- *      (issued out-of-band to trusted operators such as Eliza Cloud).
+ *   1. A platform key — header `X-Steward-Platform-Key`, authenticated and
+ *      authorized for `platform:trade:operator`. This is the cross-tenant
+ *      operator credential (issued out-of-band to trusted operators such as
+ *      Eliza Cloud).
  *   2. A tenant-admin credential — falls through to `tenantAuth`, which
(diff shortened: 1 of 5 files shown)


lalalune:
Review finding: the dedicated operator scope and durable requested evidence winning over cached success are good directions, but this is not safe to land yet. Current CLI/init and Railway examples do not provision platform:trade:operator, so documented/generated recovery credentials would immediately 403. Cached-only Redis state remains authoritative when verified durable state is absent, and platformKeyHash is computed but audit actor identity remains the constant platform-operator. Please update provisioning/runbooks and preflight, test real middleware, treat verified durable absence as non-authoritative except under an explicit bounded legacy contract, and use the non-secret key hash for audit attribution.

PR #912, opened 2026-08-23
fix(ci): restore main release gates

0xSolace:
## Summary

Restore the inherited main-branch CI baselines before the auth release PR is rebased:

- backport the proven develop PR #680 PostgreSQL test-runner and repeatable-read snapshot repair
- restore the exact follow-up fixture isolation repair already proven on develop
- replace ineffective proxy audit fault monkey-patches with database-trigger fault injection against the transaction actually used
- make the required Docker PR build eligible for every input copied by the Dockerfile, including source-only package changes
- add a regression test for the Docker workflow path boundary

## Conflict resolution

The #680 backport conflicted in two request-database files. Resolution deliberately preserves both sides:

- retains main request-capability revocation coverage while adding the snapshot/deadline contracts
- retains main request-scoped database guards while adding isolation/read-only identity checks and deadline phases on the exact active tenant transaction

No login/auth behavior is changed by this PR. No deployment or production state is touched.

## Hosted CI follow-up

The initial head exposed three deterministic fixture defects:

- plugin-trading inserted an empty tenant API-key hash before the default-tenant bootstrap, causing the unique index to turn token-status auth checks into 500s
- approvals proxy atomicity monkey-patched tx.execute, which no longer intercepts the real audit insert
- provider-case cleanup deleted every tenant and tripped the unresolved-transaction deletion guard on unrelated state

The current head stores a unique non-matching fixture hash, injects approval audit failure with a database trigger, and ports the exact develop scoped-cleanup plus foreign-state sentinel proof.

## Why this is separate

PR #900 currently inherits main baseline failures, while its required build-and-push check was not scheduled for source-only changes. Landing this prerequisite first gives the auth release a meaningful changed-path signal after rebase.

## Validation

Head: 7694d0d25953f3110143c4977182891e1a1c37f4

Current-head delta proof with Bun 1.3.14:

- plugin-trading failed file: 5 passed, 0 failed
- approval/audit atomicity: 5 passed, 0 failed
- mounted real-Postgres snapshot: 1 passed, 0 failed
- isolated approval/provider-case corpus: 10/10 files passed
- API and plugin-trading typechecks passed
- Biome on all four changed files passed
- git diff --check clean

Earlier branch proof retained by this head:

- relevant workspace build: 24/24 tasks
- full proxy package: 319 passed, 0 failed
- request/tenant DB contracts: 29 passed, 0 failed
- affected API suite: 14/14 isolated files passed
- Docker path regression: 2 passed, 0 failed
[...]

PR #900, opened 2026-08-22
fix(auth): promote staging auth recovery to main

0xSolace:
## Summary

Promotes the narrowly audited staging authentication recovery set to `main` in one branch-protected release gate. This replaces the separate draft backports in #888, #890, #891, #892, #893, and #894 without merging them independently.

### Authentication recovery

- Reuse a pre-migrated auth store without application-role DDL.
- Resolve the default auth tenant from each Worker request environment.
- Bind login audit writes to the verified tenant RLS context.
- Restore SimpleWebAuthn v13 browser calls and keep new-email signup free of conditional passkey autofill.
- Preserve the non-enumerating passkey pre-auth contract: known, no-passkey, and unknown emails all receive the same `200` response with empty `allowCredentials`, no credential IDs/transports/user IDs, and both supported challenge lookup shapes are stored.
- Keep registration an explicit email/OTP path; React does not infer account existence from a `404`, and the SDK retains the caller-controlled fallback behavior.
- Mark session tokens as JWTs and bind their standard subject claim.
- Recover OAuth/OIDC browser callback failures, content negotiation, and GitHub email fallback behavior.

### Email identity and callback routing

- Add validated tenant `brandName` support to built-in magic-link and OTP templates.
- Add deployment fallbacks for `EMAIL_BRAND_NAME`, `EMAIL_MAGIC_LINK_BASE_URL`, and `EMAIL_MAGIC_LINK_CALLBACK_PATH` so hosted email identity and callback routing survive an absent or temporarily unavailable tenant config.
- Keep the email landing origin independent from API-oriented `APP_URL`.

### Migration/readiness floor

- Add only the exact already-applied `0112_rls_activation_release_gates.sql` and `0113_personal_tenant_account_lifecycle.sql` blobs plus journal indexes 112/113.
- The SQL SHA-256 values and complete resulting journal byte-match source commits `100073d8` and `05944dda`.
- No later migration, fake journal entry, database mutation, or readiness relaxation is included. Staging runs with `SKIP_MIGRATIONS=1`; this aligns the image manifest with its already-applied database ledger.

### Promotion guardrails

- Accept only a full commit already on `main` with an exact successful `docker.yml` main push.
- Resolve and validate OCI revision/version/source provenance, then pass Railway only `repository@sha256:digest`.
- Require a configured health origin before production mutation and ship `railway.json` with `/health`.
- Keep `dry_run=true` from creating or marking an explicit GitHub Production deployment.
[...]

PR #914, opened 2026-08-24
fix(auth): scope passkey recovery to relying party

wakesync:
## Summary

Stacked on #900 for review; do not merge independently. This makes the ownership-bound orphaned-passkey recovery safe across relying parties and preserves the structured recovery code through the SDK.

## Why

The staging hotfix treated any authenticator for an OTP-verified email as proof that a passkey already existed for the active RP. The current `authenticators` schema has no RP provenance, so a credential created for another RP can incorrectly block enrollment and redirect the user into a login ceremony that cannot succeed.

## Behavior

- adds nullable `authenticators.rp_id` in migration `0114`;
- persists exact `PasskeyAuth.rpID` for new registrations;
- returns ownership-bound `passkey_already_registered` only for an explicit same-RP credential;
- allows registration when only explicit cross-RP credentials exist;
- never guesses for legacy `NULL` rows: they remain in `excludeCredentials` for browser adjudication and are backfilled only after a verified assertion;
- filters/rejects explicit cross-RP credentials in MFA/login while preserving generic errors;
- leaves the privacy-preserving, non-enumerating login/options response unchanged;
- preserves the structured 409 payload in `StewardApiError.data` and exports a type guard for clients.

## Verification

All run with Bun 1.3.9 on exact head:

- focused route/DB/SDK/enumeration: 14/14
- broader passkey/email/migration/PGLite: 39/39
- auth audit/SSO: 22/22
- linked accounts: 20/20
- DB, auth, SDK, and API builds/typechecks

## Release sequencing

- Draft stacked on #900 head `f8083c2f` because no canonical post-#912/#900 lineage exists yet.
- #912 is an independent prerequisite and should merge first after review.
- Rebase #900 after #912, then rebase this patch onto the resulting release branch and rerun the full PostgreSQL/CI matrix.
- Production promotion is blocked separately: production's migration ledger is newer than #900's embedded expected tip, so #900 would currently fail `/ready` on production.
- No staging or production mutation is part of this PR.

## Related live proof

The staging overlay proved the user flow before this hardening: OTP verification -> ownership-bound recovery -> exact saved passkey -> session -> `/chat`, with no auth 500 and no duplicate register/verify request. The canonical release must be rebuilt and restaged from the reviewed main lineage before promotion.

### DIFF
--- a/packages/api/src/__tests__/auth-passkey-mfa-hardening.test.ts
+++ b/packages/api/src/__tests__/auth-passkey-mfa-hardening.test.ts
@@ -32,7 +32,7 @@ describe("passkey MFA hardening", () => {
       "getChallengeStore().consume(challengeKey)",
       handlerStart,
     );
-    const counterUpdate = authSource.indexOf(".set({ counter:", handlerStart);
+    const counterUpdate = authSource.indexOf(".update(authenticators)", handlerStart);
     const mfaMethod = authSource.indexOf('mfaMethod: "passkey"', handlerStart);
 
     expect(readChallenge).toBeGreaterThan(handlerStart);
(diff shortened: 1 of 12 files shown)


lalalune (approved):
Independent exact-head review of 86d4a381538ed9c6ceee18c32049b1046094671a: verified same-RP-only recovery, explicit cross-RP enrollment, legacy NULL provenance handling/backfill, generic cross-RP verification failures, nullable append-only migration 0114, and structured SDK error preservation. Focused exact-head route/MFA/SDK/migration suite passed 16/16; auth and DB typechecks, Biome, diff check, and conflict-free stack passed. Merge only after #900 and the protected exact-head matrix.

PR #916, opened 2026-08-24
fix(auth): serialize refresh rotation expiry

wakesync:
## Summary

Fixes staging `POST /auth/refresh` failures during valid token rotation. The raw SQL path passed a JavaScript `Date` directly to the Neon WebSocket driver, which rejects it before PostgreSQL executes `steward_bootstrap.auth_rotate_refresh_token`.

The expiry is now serialized to ISO text and explicitly cast to `timestamptz` at the SQL boundary. Unknown and reused refresh tokens retain their intended 401 behavior.

## Type

- [x] `fix` — Bug fix

## Scope

- `packages/api` auth refresh rotation
- Valid, unknown, and reused refresh-token lifecycle coverage

## Breaking Changes

None

## Staging proof

- Durable Railway source deploy from the equivalent fix commit is healthy at `/health`.
- Synthetic valid refresh returned 200, removed the predecessor row, inserted the successor, and returned the matching successor token.
- Random invalid refresh returned 401 with `Invalid or expired refresh token`.
- No production changes were made.

## Checklist

- [x] Focused tests pass (10/10, 44 assertions)
- [x] `@stwd/api` dependency build passes (24/24 tasks)
- [x] Biome and `git diff --check` pass
- [x] Commit messages follow conventional format
- [x] PR targets `main` as the authorized isolated staging recovery follow-up

### DIFF
--- a/packages/api/src/__tests__/auth-refresh-reuse.test.ts
+++ b/packages/api/src/__tests__/auth-refresh-reuse.test.ts
@@ -28,4 +28,16 @@ describe("refresh token reuse detection", () => {
     expect(userLock).toBeLessThan(atomicRotate);
     expect(rotationBody.indexOf("revocationStore.getUserRevokedBefore")).toBeLessThan(atomicRotate);
   });
+
+  it("serializes the raw rotation expiry parameter before it reaches the database driver", () => {
+    const rotationStart = authSource.indexOf("async function rotateRefreshTokenInsideTenant");
+    expect(rotationStart).toBeGreaterThanOrEqual(0);
+    const rotationBody = authSource.slice(
+      rotationStart,
+      authSource.indexOf("/** Build the standard dual-token auth response.", rotationStart),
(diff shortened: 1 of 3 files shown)


lalalune (approved):
Reviewed exact head `f3b9827da519c6c6149b97872eb13a8acd533f04`. The change keeps the raw Neon SQL boundary primitive-only by serializing the successor expiry to ISO text and explicitly casting to `timestamptz`; it does not alter token lifetime or reuse semantics. Exact-head refresh lifecycle tests pass 10/10 (44 assertions), covering valid rotation, predecessor deletion/successor creation, unknown token 401, reused token 401, tenant switching, revocation, and deactivation. Biome and diff check pass. The API typecheck reached an unchanged missing built `@stwd/attestation` workspace artifact in this isolated worktree; the protected clean-install matrix is authoritative and must be green before merge.

PR #918, opened 2026-08-24
fix(auth): log redacted email acceptance receipts

wakesync:
## Summary

- log the validated provider name and provider message ID after email acceptance
- keep recipient, subject, rendered body, OTP, token, and callback URL out of logs
- prove rejected, missing, and malformed receipts never emit a false acceptance line

## Incident context

Staging Magic Link requests returned 200 because Resend accepted them, but neither message reached the QA inbox. The runtime key is intentionally send-only, and the validated Resend message ID was discarded, preventing exact dashboard correlation. This restores that safe correlation receipt without widening credential access or changing challenge activation.

## Verification

- `bunx turbo run build --filter=@stwd/auth...`
- `bun test packages/auth/src/__tests__/email-delivery-failclosed.test.ts` (42 pass, 163 assertions)
- `bunx biome check packages/auth/src/email.ts packages/auth/src/__tests__/email-delivery-failclosed.test.ts`
- `git diff --check`

No production or Railway state changes.

### DIFF
--- a/packages/auth/src/__tests__/email-delivery-failclosed.test.ts
+++ b/packages/auth/src/__tests__/email-delivery-failclosed.test.ts
@@ -443,6 +443,90 @@ describe("fail-closed magic-link delivery", () => {
     auth.destroy();
   });
 
+  it("logs only the provider acceptance receipt after a successful send", async () => {
+    const backend = new CapturingBackend();
+    const recipient = "acceptance-log-recipient@example.com";
+    const subjectCanary = "acceptance-log-subject-canary";
+    const bodyCanary = "acceptance-log-body-token-canary";
+    const infoCalls: unknown[][] = [];
+    const originalInfo = console.info;
+    console.info = (...args: unknown[]) => infoCalls.push(args);
+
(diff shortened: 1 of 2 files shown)


lalalune:
Review fix pushed at exact head `09f112e382f9a82065b93659a3c7cd369ab9faf9`: acceptance receipt fields now reject non-ASCII characters, including Unicode line separators and bidi controls, before structured logging. Added adversarial invalid-receipt cases. Targeted Biome and `git diff --check` pass locally; the focused test could not start in the isolated worktree because the shared dependency cache lacks `resend`, so hosted exact-head auth tests remain the execution gate.

lalalune:
Follow-up local execution at exact head `09f112e38cd18cb517706991b93d9b9314b7c072`: after restoring the filtered offline dependency graph, `email-delivery-failclosed.test.ts` passes 42/42 with 165 assertions (including the new Unicode line-separator/bidi cases); `@stwd/auth` TypeScript, targeted Biome, and `git diff --check` also pass.

PR #925, opened 2026-08-26
ci: back-merge main into develop and repair workflows

lalalune:
## Summary
- back-merge current `main` into `develop` with all conflicts resolved
- restore database/tenant context helpers required by the merged API and DB surfaces
- preserve database-scoped vault test isolation and lazy tenant bootstrap
- raise vault isolated-suite timeout to 30s in push and PR CI

## Validation
- targeted DB, plugin-trading, proxy, recovery, KMS tests
- API dependency build, lint, and typecheck pass locally
- full vault suite is running with the corrected timeout

The required hosted checks should provide the final exact-head CI and deployed-environment evidence.


lalalune:
Reopening immediately to refresh exact-head checks after the final RLS inventory fix.

lalalune:
Refreshing exact-head checks after synchronizing the activation ACL contract for core and plugin migration ledgers.

lalalune:
Refreshing exact-head checks after correcting the final RLS activation connection to use the restricted migration role.

